PR #9, opened 2019-02-12
Create dask-cuda-worker CLI utility

### LINKED ISSUE #3: Create dask-cuda-worker CLI utility
The logic in `LocalCUDACluster` might be also encoded into a command line utility like `dask-worker`.  This would presumably start many workers on the same machine, assigning environment variables as necessary.  

This would be useful because it might compose nicely with the other vanilla deployment solutions like dask-yarn, dask-jobqueue, and dask-kubernetes because we could then simply have them point to `dask-cuda-worker` rather than `dask-worker` and hopefully reuse most of the existing code with minimal modification.

beberg:
Closes #3

WIP: fixing tests (env use overlap)

### DIFF
--- a/dask_cuda/dask_cuda_worker.py
+++ b/dask_cuda/dask_cuda_worker.py
@@ -24,6 +24,7 @@ from distributed.proctitle import (
 )
 
 from .local_cuda_cluster import cuda_visible_devices
+from .utils import get_n_gpus
 
 from toolz import valmap
 from tornado.ioloop import IOLoop, TimeoutError
@@ -56,15 +57,6 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     help="private key file for TLS (in PEM format)",
 )
 @click.option(
-    "--worker-port",
-    type=int,
-    default=0,
-    help="Serving computation port, defaults to random",
-)
-@click.option(
-    "--nanny-port", type=int, default=0, help="Serving nanny port, defaults to random"
-)
-@click.option(
     "--bokeh-port", type=int, default=0, help="Bokeh port, defaults to random port"
 )
 @click.option(
@@ -74,20 +66,6 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     show_default=True,
     required=False,
     help="Launch Bokeh Web UI",
-)
-@click.option(
-    "--listen-address",
-    type=str,
-    default=None,
-    help="The address to which the worker binds. " "Example: tcp://0.0.0.0:9000",
-)
-@click.option(
-    "--contact-address",
-    type=str,
-    default=None,
-    help="The address the worker advertises to the scheduler for "
-    "communication with it and other workers. "
-    "Example: tcp://127.0.0.1:9000",
 )
 @click.option(
     "--host",
@@ -103,12 +81,6 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     "--interface", type=str, default=None, help="Network interface like 'eth0' or 'ib0'"
 )
 @click.option("--nthreads", type=int, default=0, help="Number of threads per process.")
-@click.option(
-    "--nprocs",
-    type=int,
-    default=1,
-    help="Number of worker processes to launch.  Defaults to one.",
-)
 @click.option(
     "--name",
     type=str,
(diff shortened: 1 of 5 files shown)


mrocklin:
To get automatic styling as a precommit hook try the following from the root directory of this repository.

Then any commit should style things automatically for you

```
pip install pre-commit
pre-commit install
```

Alternatively you can do this manually with `black`

```
pip install black
black dask_cuda
```

dask_cuda/dask_cuda_worker.py
```diff
@@ -0,0 +1,268 @@
...
+                   " visible to the scheduler and other workers. "
+                   "See --listen-address and --contact-address if you "
+                   "need different listen and contact addresses. "
+                   "See --interface.")
+@click.option('--interface', type=str, default=None,
+              help="Network interface like 'eth0' or 'ib0'")
+@click.option('--nthreads', type=int, default=0,
+              help="Number of threads per process.")
+@click.option('--nprocs', type=int, default=1,
+              help="Number of worker processes to launch.  Defaults to one.")
```
mrocklin: I suggest that for now we remove this option and define `nprocs` ourselves as the number of available GPUs.  I think that the ideal  current behavior of `dask-cuda-worker` on a DGX would be to create eight processes, unless restricted by `CUDA_VISIBLE_DEVICES`.  

The function `utils.py::get_n_gpus` may be useful.  I think that this logic is also already in LocalCUDACluster, which may serve as a reference.

dask_cuda/tests/test_dask_cuda_worker.py
```diff
@@ -0,0 +1,41 @@
...
+from distributed import Client
+from distributed.metrics import time
+from distributed.utils import sync, tmpfile
+from distributed.utils_test import (popen, slow, terminate_process,
+                                    wait_for_port)
+from distributed.utils_test import loop  # noqa: F401
+
+
+def test_nanny_worker_ports(loop):
+    os.environ["CUDA_VISIBLE_DEVICES"] = "2,3,7,8"
```
mrocklin: The way to make this clean is to use either a try-finally block or perhaps the following context manager

```python
from contextlib import contextmanager

@contextmanager
def env(**kwargs):
    old = {k: os.environ.get(k, '--missing--') for k in kwargs}
    for k, v in kwargs.items():
        os.environ[k] = v

    yield # give control to the rest of the block

    for k, v in old.items():
        if v == '--missing--':
            del os.environ[k]
        else:
            os.environ[k] = v
```

Then you would use this as follows:

```python
with env(CUDA_VISIBLE_DEVICES='2,3,7,8'):
    with popen(...)
        ...
```

dask_cuda/dask_cuda_worker.py
beberg: I'd like t keep this open for the non-1:1 setups i described in https://github.com/rapidsai/dask-cuda/issues/3#issuecomment-461190475 where we have multiple GPUs per process.

beberg: Many flags interact with nprocs flag, mostly ones that can't be used with it. If we remove the flag and internally set nprocs = #GPUs, then those flags are never usable and things unravel a bit. How about leaving the flag but defaulting it to #GPUs instead of 1?

mrocklin: Most of the other flags we'll also want to remove.  The nanny option will go away (we need to have a nanny regardless).  And the threads option will also go away (we need to have one thread per gpu until we figure out a nice concurrency story for threads on GPUs)

mrocklin: @beberg mentioned that other parts of Dask may expect keywords like `--nthreads` to exist.  Indeed after some looking into things it looks like dask-jobqueue does expect an `--nthreads` keyword.

mrocklin: I still think that we should drop the `--nprocs` keyword (this is infrequently used in deployment solutions), and probably also the `--nanny` keyword.  If we keep either of those then they should probably err if the user tries to use them.  I'm inclined to remove though.

beberg: Dropped
-    worker_port,
-    listen_address,
-    contact_address,
-    nanny_port,
-    nprocs,
-    nanny,

nprocs internally set to #GPUs

PR #13, opened 2019-03-13
[REVIEW] Add ci scripts & conda recipe

raydouglass:


mrocklin:
cc'd @jakirkham , who knows much more than I do about packaging :)

conda/recipes/dask-cuda/meta.yaml
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) 2019, NVIDIA CORPORATION.
+
+# Usage:
+#   conda build -c conda-forge -c defaults .
+{% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev').lstrip('v') %}
+{% set git_revision_count=environ.get('GIT_DESCRIBE_NUMBER', 0) %}
+{% set cuda_version='.'.join(environ.get('CUDA', 'unknown').split('.')[:2]) %}
```
kkraus14: Given this is a pure python project I don't think we need to CUDA version it.

conda/recipes/dask-cuda/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  path: ../../..
+
+build:
+  number: {{ git_revision_count }}
+  string: cuda{{ cuda_version }}_{{ git_revision_count }}
+  script_env:
+    - CC
+    - CXX
+    - CUDAHOSTCXX
+    - PARALLEL_LEVEL
```
kkraus14: No actual compilation so this isn't needed.
```suggestion
```

conda/recipes/dask-cuda/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - BUILD_ABI
```
kkraus14: No compilation since this is pure Python so this isn't needed:
```suggestion
```

conda/recipes/dask-cuda/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+requirements:
+  build:
+    - python x.x
+    - dask
```
kkraus14: @mrocklin I believe this depends on a relatively new version of dask, we should specify that here.

ci/gpu/build.sh
```diff
@@ -0,0 +1,49 @@
...
+
+################################################################################
+# TEST - Run tests
+################################################################################
+
+pip install -e .
+pip install pytest
+py.test --cache-clear --junitxml=${WORKSPACE}/junit-libgdf.xml -v
+
+exit 0
```
dillon-cullinan: Drop the exit 0.

conda/recipes/dask-cuda/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - distributed>=1.25.2
+  run:
+    - python x.x
+    - dask
```
kkraus14: Here too.

ci/gpu/build.sh
```diff
@@ -0,0 +1,49 @@
...
+# FIX Added to deal with Anancoda SSL verification issues during conda builds
+conda config --set ssl_verify False
+
+################################################################################
+# TEST - Run tests
+################################################################################
+
+pip install -e .
+pip install pytest
+py.test --cache-clear --junitxml=${WORKSPACE}/junit-libgdf.xml -v
```
mt-jones: We should consider migrating to `pytest`, as `py.test` is no longer the well supported/recommended option, and may well be deprecated soon.

ci/cpu/upload-anaconda.sh
```diff
@@ -0,0 +1,43 @@
...
+fi
+
+SOURCE_BRANCH=master
+
+if [ "$LABEL_MAIN" == "1" -a "$BUILD_ABI" == "1" ]; then
+  LABEL_OPTION="--label main --label cuda${CUDA_REL}"
+elif [ "$LABEL_MAIN" == "0" -a "$BUILD_ABI" == "1" ]; then
+  LABEL_OPTION="--label dev --label cuda${CUDA_REL}"
+elif [ "$LABEL_MAIN" == "1" -a "$BUILD_ABI" == "0" ]; then
+  LABEL_OPTION="--label cf201901 --label cf201901-cuda${CUDA_REL}"
```
mt-jones: does `cf201901` have a particular significance?

setup.py
```diff
@@ -20,8 +23,8 @@
         "Topic :: Database",
         "Topic :: Scientific/Engineering",
         "License :: OSI Approved :: Apache License",
-        "Programming Language :: Python :: 3.5",
```
mt-jones: ❤️

ci/checks/style.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# Activate common conda env
+source activate gdf
+
+# Run flake8 and get results/return code
+FLAKE=`flake8 python`
+RETVAL=$?
+
+# Output results if failure otherwise show pass
+if [ "$FLAKE" != "" ]; then
```
rlratzel: Also add (( $RETVAL != 0 )) to the if stmnt?

ci/cpu/upload-anaconda.sh
raydouglass: That's the label for ABI=OFF, but we don't support that anymore, so I'll remove this actually.

conda/recipes/dask-cuda/meta.yaml
mrocklin: Lets go for `dask-core=1.1.4` and drop `dask`

mrocklin:
```suggestion
    - dask-core>=1.1.4
```

mrocklin: Oh, actually I'm sorry.  I was thinking dask-cudf.  You probably still want dask-core and distributed, but probably don't need to be as aggressive.  Shouldn't matter much either way though.

    distributed>=1.26

jakirkham:
Thanks for working on this all. Sorry to get to reviewing it late (was flying back across the country). Added a small tweak in PR ( https://github.com/rapidsai/dask-cuda/pull/15 ).

There are some other things that would be good to discuss about conda packaging (mostly due to improvements in conda and conda-build over the years). Though we can save this discussion until after things have cooled down a bit.

Issue #16, opened 2019-03-15
[BUG] Conda vs Pypi dask dependency issue


raydouglass:
Since the conda dependency is `dask-core`, but the pypi one is `dask`, this prevents the conda build from working.

The conda build delegates to `setup.py` and this process will attempt to download `dask`. conda-build errors on any setuptools downloads.

jakirkham:
So what we typically did in conda-forge when building a Python package is run the following, which stops `setuptools` from pulling in dependencies. These days conda-forge builds with `pip`, but I don't think we should get into that yet.

```
python setup.py install --single-version-externally-managed --record=record.txt
```

Where is `setup.py` being called currently? Would it be easy to add these flags there? If not, there are other options. Just trying to get a sense of what it would take to correct this near term.

raydouglass:
> Where is `setup.py` being called currently? Would it be easy to add these flags there? If not, there are other options. Just trying to get a sense of what it would take to correct this near term.

https://github.com/rapidsai/dask-cuda/tree/branch-0.6/conda/recipes/dask-cuda

jakirkham:
Alright have submitted PR ( https://github.com/rapidsai/dask-cuda/pull/17 ) to fix this. Please take a look when you have a chance :)

jakirkham:
Please let me know if that fixes it or if there are still issues that need attention.

PR #17, opened 2019-03-15
Have setuptools ignore installed

jakirkham:
Fixes https://github.com/rapidsai/dask-cuda/issues/16
Closes https://github.com/rapidsai/dask-cuda/pull/15

This disables Setuptools default behavior of pulling in dependencies at install time in the Conda package. Updates the Conda packages requirements at build time accordingly.

Also fixes the `requirements.txt` file to pull in `dask` (not `dask-core`) as that is how things are named on PyPI.

Finally pulls in the fix from PR ( https://github.com/rapidsai/dask-cuda/pull/15 ).

cc @raydouglass


jakirkham:
The first failure looks like a GitHub rate limiting thing. So could be restarted.

Wasn't able to see what was going on with the second one, but I guess it is already restarted :)

jakirkham:
Oops, sorry, pushed that commit to the wrong branch. Fixed now. 😅

PR #22, opened 2019-03-18
FIX Package name

raydouglass:
Fixes #22


mrocklin:
Dask packages generally use the `dask-foo` convention.  Is there a reason to use underscores here instead?

Here is an example from the dask-core recipe

https://github.com/conda-forge/dask-core-feedstock/blob/master/recipe/meta.yaml

PR #23, opened 2019-03-18
[REVIEW] Replace underscore with dash in conda packaging

dillon-cullinan:


dillon-cullinan:
No changes needed. Should use underscore for import test

PR #24, opened 2019-03-18
Remove comma preventing importing dask_cuda

pentschev:
This has been fixed in branch-0.7, but we should fix it in branch-0.6 as well.


pentschev:
@mike-wendt can we add this in the branch-0.6 too?

Issue #28, opened 2019-03-25
Implement dask_cuda.__version__


paulhendricks:
It'd be great for reproducibility if `dask_cuda.__version__` was implemented.

```python
import dask_cuda  # ; print('Dask CUDA Version:', dask_cuda.__version__)
```

mrocklin:
Heartily agreed.

pentschev:
This has been addressed by #88 , closing.

PR #29, opened 2019-03-25
added versioning capability

paulhendricks:


GPUtester:

paulhendricks:
@mrocklin @randerzander

paulhendricks:
Talking with @randerzander, this would be a "Nice to Have" feature: https://github.com/rapidsai/dask-cuda/issues/28

Should versioning start at 0.1? At 0.6/0.7?

mrocklin:
You will also want to mirror this change in the `setup.py` file so that it affects PyPI downloads.  

I don't have a preference around which version number we use.

mrocklin:
An alternative would be to use a project like versioneer, which would generate version numbers given git tags.  A number of other projects do this.  This is fine too though.

raydouglass:
rerun tests

pentschev:
#88 was merged yesterday, and should improve versioning. Closing now, but feel free to open a new PR if something is still missing.

PR #34, opened 2019-04-16
Apply all label options in CPU CI script

mrocklin:
cc @rlratzel


mrocklin:
I'm happy to merge this any time.  @rlratzel can you confirm that this is what you need?

PR #39, opened 2019-04-30
Support protocol= keyword in LocalCUDACluster

mrocklin:
The test is currently failing for unclear reasons.  

Hopefully that resolves itself as ucx-py becomes more stable

### DIFF
--- a/dask_cuda/local_cuda_cluster.py
+++ b/dask_cuda/local_cuda_cluster.py
@@ -84,6 +84,7 @@ class LocalCUDACluster(LocalCluster):
             self._start_worker(
                 **self.worker_kwargs,
                 env={"CUDA_VISIBLE_DEVICES": cuda_visible_devices(i)},
+                name="gpu-" + str(i),
             )
             for i in range(n_workers)
         ]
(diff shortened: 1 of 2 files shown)


dask_cuda/local_cuda_cluster.py
```diff
@@ -35,7 +36,7 @@ def __init__(
         threads_per_worker=1,
         processes=True,
         memory_limit=None,
-        **kwargs
+        **kwargs,
```
jakirkham:
```suggestion
        **kwargs
```

mrocklin: This is an issue with black.  It's fairly insistent about this, which is a bit of a pain for python 3.5 users.

mrocklin: See https://github.com/python/black/issues/419

In the meantime, I don't particularly care to fight against black.  I think I'm just going to merge as is for now.  Hopefully once black releases again this will go back.

jakirkham: Ok, to be clear wasn't meaning this comment to hold things up. Just that stuck out to me was all.

PR #40, opened 2019-05-03
[REVIEW] Use latest tag in update-version

raydouglass:
Change update-version script to ignore dev/alpha/rc versions and use only the latest version

Also updated where changes are made


mrocklin:
@raydouglass do I need to do anything here or is this something that you're handling?

raydouglass:
@mrocklin Just wanted @dillon-cullinan to double check me before I merge it. No action on your part!

PR #46, opened 2019-05-09
Add a hard-coded DGX configuration

mrocklin:
This currently depends on github.com/mrocklin/distributed@dev

### DIFF
--- a/dask_cuda/dgx.py
+++ b/dask_cuda/dgx.py
@@ -25,17 +25,61 @@ affinity = {  # nvidia-smi topo -m
 }
 
 
-def DGX(interface="ib", dashboard_address=":8787", threads_per_worker=1, **kwargs):
-    gpus = list(
-        map(int, os.environ.get("CUDA_VISIBLE_DEVICES", "0,1,2,3,4,5,6,7").split(","))
-    )
+def DGX(
+    interface="ib",
+    dashboard_address=":8787",
+    threads_per_worker=1,
+    silence_logs=True,
+    CUDA_VISIBLE_DEVICES=None,
+    **kwargs
+):
+    """ A Local Cluster for a DGX 1 machine
+
+    NVIDIA's DGX-1 machine has a complex architecture mapping CPUs, GPUs, and
+    network hardware.  This function creates a local cluster that tries to
+    respect this hardware as much as possible.
+
+    It creates one Dask worker process per GPU, and assigns each worker process
+    the correct CPU cores and Network interface cards to maximize performance.
+
+    That being said, things aren't perfect.  Today a DGX has very high
+    performance between certain sets of GPUs and not others.  A Dask DGX
+    cluster that uses only certain tightly coupled parts of the computer will
+    have significantly higher bandwidth than a deployment on the entire thing.
+
+    Parameters
+    ----------
+    interface: str
+        The interface prefix for the infiniband networking cards.  This is
+        often "ib"` or "bond".  We will add the numeric suffix 0,1,2,3 as
+        appropriate.  Defaults to "ib".
+    dashboard_address: str
+        The address for the scheduler dashboard.  Defaults to ":8787".
+    CUDA_VISIBLE_DEVICES: str
+        String like ``"0,1,2,3"`` or ``[0, 1, 2, 3]`` to restrict activity to
+        different GPUs
+
+    Examples
+    --------
+    >>> from dask_cuda import DGX
+    >>> from dask.distributed import Client
+    >>> cluster = DGX(interface='ib')
+    >>> client = Client(cluster)
+    """
+    if CUDA_VISIBLE_DEVICES is None:
+        CUDA_VISIBLE_DEVICES = os.environ.get("CUDA_VISIBLE_DEVICES", "0,1,2,3,4,5,6,7")
+    if isinstance(CUDA_VISIBLE_DEVICES, str):
+        CUDA_VISIBLE_DEVICES = CUDA_VISIBLE_DEVICES.split(",")
+    CUDA_VISIBLE_DEVICES = list(map(int, CUDA_VISIBLE_DEVICES))
 
     spec = {
         i: {
             "cls": Nanny,
             "options": {
                 "env": {
-                    "CUDA_VISIBLE_DEVICES": cuda_visible_devices(ii, gpus),
+                    "CUDA_VISIBLE_DEVICES": cuda_visible_devices(
+                        ii, CUDA_VISIBLE_DEVICES
+                    ),
                     # 'UCX_NET_DEVICES': 'mlx5_%d:1' % (i // 2)
                     "UCX_TLS": "rc,cuda_copy,cuda_ipc",
                 },
@@ -46,18 +90,21 @@ def DGX(interface="ib", dashboard_address=":8787", threads_per_worker=1, **kwarg
                 "preload": ["dask_cuda.initialize_context"],
                 "dashboard_address": ":0",
                 "plugins": [CPUAffinity(affinity[i])],
+                "silence_logs": silence_logs,
             },
         }
-        for ii, i in enumerate(gpus)
+        for ii, i in enumerate(CUDA_VISIBLE_DEVICES)
     }
 
     scheduler = {
         "cls": Scheduler,
         "options": {
-            "interface": interface + str(gpus[0] // 2),
+            "interface": interface + str(CUDA_VISIBLE_DEVICES[0] // 2),
             "protocol": "ucx",
             "dashboard_address": dashboard_address,
         },
     }
 
-    return SpecCluster(workers=spec, scheduler=scheduler, **kwargs)
+    return SpecCluster(
+        workers=spec, scheduler=scheduler, silence_logs=silence_logs, **kwargs
+    )
--- a/dask_cuda/dgx.py
+++ b/dask_cuda/dgx.py
@@ -25,7 +25,7 @@ affinity = {  # nvidia-smi topo -m
 }
 
 
-def DGX(interface="ib", **kwargs):
+def DGX(interface="ib", dashboard_address=":8787", threads_per_worker=1, **kwargs):
     gpus = list(
         map(int, os.environ.get("CUDA_VISIBLE_DEVICES", "0,1,2,3,4,5,6,7").split(","))
     )
@@ -41,7 +41,7 @@ def DGX(interface="ib", **kwargs):
                 },
                 "interface": interface + str(i // 2),
                 "protocol": "ucx",
-                "ncores": 1,
+                "ncores": threads_per_worker,
                 "data": dict,
                 "preload": ["dask_cuda.initialize_context"],
                 "dashboard_address": ":0",
@@ -56,7 +56,7 @@ def DGX(interface="ib", **kwargs):
         "options": {
             "interface": interface + str(gpus[0] // 2),
             "protocol": "ucx",
-            "dashboard_address": ":8787",
+            "dashboard_address": dashboard_address,
         },
     }
 
--- a/dask_cuda/dgx.py
+++ b/dask_cuda/dgx.py
@@ -1,7 +1,8 @@
 import os
 
 from dask.distributed import Nanny, SpecCluster, Scheduler
-from distributed.bokeh.scheduler import BokehScheduler
+
+# from distributed.bokeh.scheduler import BokehScheduler
 
 from .local_cuda_cluster import cuda_visible_devices
 
@@ -21,7 +22,6 @@ def DGX(**kwargs):
                 },
                 "interface": "ib%d" % (i // 2),
                 "protocol": "ucx",
-                "local_dir": "/raid/mrocklin",
                 "ncores": 1,
                 "data": dict,
             },
@@ -34,7 +34,7 @@ def DGX(**kwargs):
         "options": {
             "interface": "ib0",
             "protocol": "ucx",
-            "services": {("bokeh", "10.33.227.165:8900"): BokehScheduler},
+            # "services": {("bokeh", "10.33.227.165:8900"): BokehScheduler},
         },
     }
 
--- a/dask_cuda/dgx.py
+++ b/dask_cuda/dgx.py
@@ -1,15 +1,31 @@
 import os
 
 from dask.distributed import Nanny, SpecCluster, Scheduler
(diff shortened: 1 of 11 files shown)


mrocklin:
Currently benchmarking with 

```python
from dask_cuda import DGX
from dask.distributed import Client, wait
cluster = DGX()
client = Client(cluster)

import cupy, dask.array as da, dask
rs = da.random.RandomState(RandomState=cupy.random.RandomState)
x = rs.normal(10, 1, size=(50000, 50000)).persist()
y = (x + x.T).sum().compute()
```

## Script

```python
from dask_cuda import DGX
from dask.distributed import Client, wait
import cupy, dask.array as da, dask

with DGX() as cluster:
    with Client(cluster) as client
        rs = da.random.RandomState(RandomState=cupy.random.RandomState)
        x = rs.normal(10, 1, size=(50000, 50000)).persist()
        wait(x)
        y = (x + x.T).sum().compute()
```

dask_cuda/dgx.py
```diff
@@ -0,0 +1,41 @@
...
+        int, os.environ.get("CUDA_VISIBLE_DEVICES", "0,1,2,3,4,5,6,7").split(",")
+    )
+
+    spec = {
+        i: {
+            "cls": Nanny,
+            "options": {
+                "env": {
+                    "CUDA_VISIBLE_DEVICES": cuda_visible_devices(i, range(8)),
+                    # 'UCX_NET_DEVICES': 'mlx5_%d:1' % (i // 2)
```
mrocklin: When I uncomment this line, things break.

mrocklin:
@Akshay-Venkatesh @rjzamora if you all have a chance, here is a system that, I think, should set up a DGX properly.  I'm still not able to use the `UCX_NET_DEVICES` environment variable effectively.

dask_cuda/dgx.py
rjzamora: Does each `i ` correspond to a distinct worker?  If so, it makes sense to me that the `UCX_NET_DEVICES` line is causing you problems.  I think you would need to list the HCA:port combo for the worker, as well as the scheduler (and maybe even other workers if there is worker-worker communication)

dask_cuda/dgx.py
```diff
@@ -0,0 +1,41 @@
...
+                "data": dict,
+            },
+        }
+        for i in gpus
+    }
+
+    scheduler = {
+        "cls": Scheduler,
+        "options": {
+            "interface": "ib0",
```
rjzamora: It might make sense to use a different interface for the scheduler in the case that none of your workers are running on gpu-0 or gpu-1

dask_cuda/dgx.py
mrocklin: There is definitely worker-worker communication.  All workers will talk to each other.

What should this line be?

mrocklin: Also, if you're interested and have time, I would encourage you to play with this branch and see if you can get the benchmark below working well.  I'd be happy to help in any way with startup difficulties.

rjzamora: I will experiment a bit and keep you updated :)

rjzamora:
@mrocklin - I am not able to run the benchmark you shared. My env:
```
mrocklin/distributed@dev
mrocklin/dask-cuda@dgx-1
Akshay-Venkatesh/ucx@1ad5b17/ucx-cuda
Akshay-Venkatesh/ucx-py@devel
```
Anything I am getting wrong here?

mrocklin:
That environment looks fine to me.  Let me push a couple changes to make it easy to get a dashboard up, and then lets jump on a screenshare to see what's happening.

dask_cuda/dgx.py
Akshay-Venkatesh: @mrocklin When you uncomment, you're asking different ucx processes to use different IB devices. Now that I think about this, I think during endpoint creation and connection establishment, if the states aren't consistent across processes there could be potential issues. I know this is contrary to the approach we discussed recently and apologize if I mislead you into doing this. Ideally, with the patches I have in `1ad5b17/ucx-cuda` you don't have to manually pick the IB devices but based on the GPU context selected, the closest HCA gets picked for transfers. 

Regardless, can you paste the error here?

I'll also give this is a shot today soon.

mrocklin: I believe that @rjzamora is getting an error similar to what we found in https://github.com/Akshay-Venkatesh/ucx-py/issues/102

mrocklin: > I'll also give this is a shot today soon.

That would be very welcome :)

mrocklin:
In offline conversation @Akshay-Venkatesh mentioned 

> Is the `cuda_visible_devices` env param consistent across dask workers for this test? That would be one explanation of why you would see the `invalid device context` error

Then I said 

```python
In [1]: from dask_cuda.local_cuda_cluster import cuda_visible_devices

In [2]: for i in range(8):
   ...:     print(i, cuda_visible_devices(i, range(8)))
   ...:
0 0,1,2,3,4,5,6,7
1 1,2,3,4,5,6,7,0
2 2,3,4,5,6,7,0,1
3 3,4,5,6,7,0,1,2
4 4,5,6,7,0,1,2,3
5 5,6,7,0,1,2,3,4
6 6,7,0,1,2,3,4,5
7 7,0,1,2,3,4,5,6
```

Then @Akshay-Venkatesh 

> is it not possible to select appropriate devices from within the python program having given all workers the same env param?

Then me again

> How can we select the appropriate device from within Python?  And actually, can I ask that we move this to GitHub?

And here we are

mrocklin:
> is it not possible to select appropriate devices from within the python program having given all workers the same env param?

Ah, so if you mean, can we ask that CuPy use a particular GPU, then yes, cupy has API for that.  cuDF may have API for that (actually I don't think that it does) but it would be different.  Same with TensorFlow and PyTorch.

Unfortunately, the only consistent way to have user code prefer a particular GPU today that is cross-library is to use `CUDA_VISIBLE_DEVICES`.

mrocklin:
I can restrict the visible devices to just one GPU though, maybe that will help?

Akshay-Venkatesh:
> How can we select the appropriate device from within Python?  And actually, can I ask that we move this to GitHub?

@mrocklin I'm pasting what I added [here](https://github.com/Akshay-Venkatesh/ucx-py/issues/99#issuecomment-488873665):

```diff
akvenkatesh@prm-dgx-33:~/ucx-py$ git diff benchmarks/recv-into-client.py
diff --git a/benchmarks/recv-into-client.py b/benchmarks/recv-into-client.py
index bba9eb8..4ef1f15 100644
--- a/benchmarks/recv-into-client.py
+++ b/benchmarks/recv-into-client.py
@@ -134,9 +134,15 @@ async def main(args=None):
         import cupy as xp
 
     if args.server:
+        if args.object_type == 'cupy':
+            xp.cuda.runtime.setDevice(0)
+            print(xp.cuda.runtime.getDevice())
         await connect(args.server, args.port, args.n_bytes, args.n_iter,
                       args.recv, xp, args.verbose, args.inc)
     else:
+        if args.object_type == 'cupy':
+            xp.cuda.runtime.setDevice(1)
+            print(xp.cuda.runtime.getDevice())
         await serve(args.port, args.n_bytes, args.n_iter,
                     args.recv, xp, args.verbose, args.inc)
```

Akshay-Venkatesh:
> I can restrict the visible devices to just one GPU though, maybe that will help?

That may work for this test but I assume you'd like to be able to use all GPUs for most workloads. Am I misinterpreting this?

mrocklin:
I tried the following spec, which allows only one visible device per process, and didn't see any improvement.

```python
    spec = {
        i: {
            "cls": Nanny,
            "options": {
                "env": {
                    "CUDA_VISIBLE_DEVICES": str(i),  # <<<----- this is the major change
                    # 'UCX_NET_DEVICES': 'mlx5_%d:1' % (i // 2)
                },
                "interface": "ib%d" % (i // 2),
                "protocol": "ucx",
                "ncores": 1,
            },
        }
        for i in range(8)
    }
```

mrocklin:
Right, so the challenge with the `xp.cuda.runtime.setDevice` approach is that we probably can't do that for all possible libraries that the user might want to use.  Unfortunately there isn't any standard API for CUDA itself in Python.

mrocklin:
> That may work for this test but I assume you'd like to be able to use all GPUs for most workloads. Am I misinterpreting this?

Yes, but I'm comfortable with each dask worker having access to only one of them.  I have one dask worker per GPU.

Regardless, it didn't seem to solve my immediate problem.

Akshay-Venkatesh:
are you seeing these errors or something else?

```
cuda_ipc_md.c:62   UCX  ERROR cuCtxGetDevice(&cu_device) is failed. ret:invalid device context
[1557433464.508925] [dgx15:55612:0]       ucp_rkey.c:250  UCX  ERROR Failed to unpack remote key from remote md[5]: Input/output error
[dgx15:55612:0:55931] Caught signal 11 (Segmentation fault: address not mapped to object at address 0x20)
```

If yes, do you mind trying `UCX_TLS=rc,sm,tcp,cuda_copy` ? Since the error is showing up in `cuda_ipc` transport I'm trying to avoid using that.

rjzamora:
Setting `UCX_TLS=rc,sm,tcp,cuda_copy` helped for me.  I also needed to set a non-default port for the scheduler on dgx15 (e.g. `"port": 8788`).

mrocklin:
> Right, so the challenge with the xp.cuda.runtime.setDevice approach is that we probably can't do that for all possible libraries that the user might want to use. Unfortunately there isn't any standard API for CUDA itself in Python.

@Akshay-Venkatesh would it make sense for UCX to respect the preference implied by `CUDA_VISIBLE_DEVICES` where the first value is used by default?  This seems to be the strongest convention on this topic that I've seen so far (though I'm fairly new).

Akshay-Venkatesh:
I think it's possible. Needs some changes in UCX but I see good reason to
do it now. I'll push a PR to address this.

mrocklin:
Thanks @Akshay-Venkatesh .  It sounds like many of the issues are consistent with that problem.  Hopefully that helps to resolve things.  If there is a development branch that I can build from (that also includes your other changes) please let me know.  I'd be keen to try it out.

Akshay-Venkatesh:
I haven't completed the PR yet but I'll definitely post here when the PR is complete.

mrocklin:
This repository will start depending on the dask/distributed master branch. 

Is it ok to place this in the ci/cpu/build.sh and ci/gpu/build.sh files?

cc @rlratzel , who I think has been thinking about cross-project testing.

mrocklin:
> This repository will start depending on the dask/distributed master branch.

> Is it ok to place this in the ci/cpu/build.sh and ci/gpu/build.sh files?

> cc @rlratzel , who I think has been thinking about cross-project testing.



Or maybe @raydouglass knows the answer to this?

rlratzel:
> This repository will start depending on the dask/distributed master branch.
> 
> Is it ok to place this in the ci/cpu/build.sh and ci/gpu/build.sh files?

Is the dependency needed just for running tests, or does the library itself require it?

If it's only needed for tests, then I believe it's a fairly common pattern to add a `pip install` line to the gpu/cpu scripts.  As a nicety to future maintainers, I've found it helpful to add checks for special dependencies my tests need in the corresponding setup, with a clear error message if they're missing. 

If the library itself needs it, then we'd obviously want to ensure it's installed with the library's conda package (meta.yml/requirements.txt). I'm not familiar with dask/distributed, but is that an option if the library depends on it?

mrocklin:
The library itself now depends on it.  We'll have to release `distributed` before the next release of `dask-cuda`

mrocklin:
@rlratzel given the information above, what is the best approach here?

mrocklin:
> If the library itself needs it, then we'd obviously want to ensure it's installed with the library's conda package (meta.yml/requirements.txt). I'm not familiar with dask/distributed, but is that an option if the library depends on it?

So, yes, it will depend on the most recent version, which is not currently released.  It seems odd to put a dev version in a requirements.txt/meta.yml file.  I can though if you prefer.  

If I don't get a follow up here soonish I'm going to just add a pip install line in the CI script.

raydouglass:
> So, yes, it will depend on the most recent version, which is not currently released. It seems odd to put a dev version in a requirements.txt/meta.yml file. I can though if you prefer.
> 

I think this would be necessary though right? Otherwise the conda packages won't install the right dependencies right?

> If I don't get a follow up here soonish I'm going to just add a pip install line in the CI script.

When is `distributed` being released? Can this PR wait for that?

mrocklin:
> I think this would be necessary though right? Otherwise the conda packages won't install the right dependencies.

How do we handle other packages on which we depend on master, like cudf?

> When is distributed being released? Can this PR wait for that?

A few weeks probably (there is a lot of churn in the upcoming release), but before RAPIDS 0.8.  

Currently this repository won't work with master branch of distributed.  This PR will fix those issues.

mrocklin:
Any further suggestions @raydouglass ?  I think it's ok to merge something so that master works with master, but I'm open to other suggestions.

raydouglass:
> Any further suggestions @raydouglass ? I think it's ok to merge something so that master works with master, but I'm open to other suggestions.

The only suggestion I have is to wait which isn't ideal. So maybe make the pip install changes. Please open an issue to remove those changes after distributed is released. That way we'll have a record and won't forget to do it before v0.8 release.

mrocklin:
rerun tests

mrocklin:
> If it's only needed for tests, then I believe it's a fairly common pattern to add a pip install line to the gpu/cpu scripts. 

This didn't suffice.  The conda build tests fail.  How would you all like this handled?

raydouglass:
The pip install has to be in `conda/recipes/dask-cuda/build.sh` script. During a conda build, a new environment is created, so environment changes in CI scripts aren't transferred over.

mrocklin:
Ah crap.  It looks like this was targetting branch 0.7 when I merged.  My apologies.  What do I need to do to correct this on the ops end?  I can easily move the commit over to 0.8, but I'm concerned that I might have messed with one of your internal systems.

mrocklin:
If it was up to me I would probably just remove the commit and force-push, or add a revert commit.

kkraus14:
@raydouglass @mike-wendt need you to jump in here to resolve this.

PR #50, opened 2019-05-13
Added installation info and example to the README

madsbk:


GPUtester:

pentschev:
CuPy 6.0.0 is expected to be released this Thursday, May 16th. I suggest we wait, then update the README to reflect that and merge.

Otherwise, LGTM.

mrocklin:
Thanks for doing this @madsbk 

One concern I have is the focus on CuPy here.  CuPy isn't a dependency of dask-cuda, or even a primary focus. People will use this library with other packages like cudf, torch, tensorflow and so on. I don't think that we should provide special support to CuPy in this project's readme.

Most people using dask-cuda today are probably people associated with RAPIDS development, particularly around cudf.

pentschev:
That's a good point @mrocklin. If there are no objections, I think we could restructure the document a bit, keeping the CuPy instructions, and over time include examples for the various libraries that are supported (along with requirements to run those).

mrocklin:
I agree that instructions like that should exist somewhere.  I'm not sure if here is the correct place.  I can't immediately think of a better place though.

pentschev:
What if we move the instructions out of `README.md` into a different file in this repository?

mrocklin:
That could make sense.  I think that we still need to figure out what is in-scope for this repository.  Originally it was just around improving the ability to set up Dask with CUDA GPUs.  It sounds like now we're considering it as a home for documentation on how to use Dask with GPU enabled libraries, which is new.

Some other options:

-  Somewhere within the RAPIDS documentation
-  Get things to a more stable state (like how you suggested to wait for cupy 6 and maybe maybe finish up the meta PR) and then add it to the standard Dask documentation
-  ...

pentschev:
I think it's helpful to have this information rather sooner than later, independent of where. On the original discussion we had with @jakirkham we decided to create a gist where we could point people to. That would be fine to me as well until we are on a more stable state.

I don't know where would be (or if there is) an appropriate place for this documentation within RAPIDS outside of this repo.

jakirkham:
FWIW moving the CuPy stuff to a gist still seems reasonable to me 🙂

kkraus14:
add to whitelist

PR #51, opened 2019-05-14
Device memory spill support (LRU-based only)

pentschev:
Based on https://github.com/rapidsai/dask-cuda/pull/35, but provides only the simple LRU mechanism (i.e., no device memory monitoring).

### DIFF
--- a/dask_cuda/dask_cuda_worker.py
+++ b/dask_cuda/dask_cuda_worker.py
@@ -23,6 +23,7 @@ from distributed.proctitle import (
     enable_proctitle_on_current,
 )
 
+from .device_host_file import DeviceHostFile
 from .local_cuda_cluster import cuda_visible_devices
 from .utils import get_n_gpus
 
@@ -252,6 +253,7 @@ def main(
             contact_address=None,
             env={"CUDA_VISIBLE_DEVICES": cuda_visible_devices(i)},
             name=name if nprocs == 1 or not name else name + "-" + str(i),
+            data=DeviceHostFile
             **kwargs
         )
         for i in range(nprocs)
--- a/dask_cuda/dask_cuda_worker.py
+++ b/dask_cuda/dask_cuda_worker.py
@@ -8,7 +8,7 @@ from sys import exit
 import click
 from distributed import Nanny, Worker
 from distributed.config import config
-from distributed.utils import get_ip_interface, parse_timedelta
+from distributed.utils import get_ip_interface, parse_timedelta, parse_bytes
 from distributed.worker import _ncores, parse_memory_limit
 from distributed.security import Security
 from distributed.cli.utils import (
@@ -100,6 +100,16 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     "'auto', or zero for no memory management",
 )
 @click.option(
+    "--device-memory-limit",
+    default="auto",
+    help="Bytes of memory per CUDA device that the worker can use. "
+    "This can be an integer (bytes), "
+    "float (fraction of total device memory), "
+    "string (like 5GB or 5000M), "
+    "'auto', or zero for no memory management "
+    "(i.e., allow full device memory usage).",
+)
+@click.option(
     "--reconnect/--no-reconnect",
     default=True,
     help="Reconnect to scheduler if disconnected",
@@ -147,6 +157,7 @@ def main(
     nthreads,
     name,
     memory_limit,
+    device_memory_limit,
     pid_file,
     reconnect,
     resources,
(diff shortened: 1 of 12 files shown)


pentschev:
@mrocklin let's move the conversation here.

> Right, mostly I want to say lets not add a new device_memory_foo= keyword yet if we can avoid it.

We need to handle it somehow. We have to pass that information to `DeviceHostFile` somehow, and we can create the object before passing to `Worker`/`Nanny` (which is fine, IMO). But the user can already pass `--nthreads` to `dask-cuda-worker`, or `n_workers` to `LocalCUDACluster`, so we should at least handle that appropriately. I'll see how this can be handled in a minimalistic way, probably by just dividing memory evenly.

mrocklin:
Under the assumption that we'll have one worker per GPU I think that we probably don't need to divide at all.  We just take the maximum available device memory of the current GPU and use that.

pentschev:
I know you @mrocklin said not to add `--device-memory-limit`, but this allows us to create the `DeviceHostFile` object without querying GPUs for memory. This allows the `CUDA_VISIBLE_DEVICES` test to work properly.

dask_cuda/tests/test_spill.py
```diff
@@ -0,0 +1,92 @@
...
+        cupy.cuda.set_pinned_memory_allocator(None)
+
+        rs = da.random.RandomState(RandomState=cupy.random.RandomState)
+        x = rs.random(int(250e6), chunks=10e6)
+        yield wait(x)
+
+        xx = x.persist()
+        yield wait(xx)
+
+        print(worker.data.device_buffer)
```
mrocklin: We might want to remove the print statements for now.

dask_cuda/tests/test_spill.py
```diff
@@ -0,0 +1,92 @@
...
+        {
+            "device_memory_limit": 1e9,
+            "memory_limit": 1e9,
+            "host_target": 0.3,
+            "host_spill": 0.3,
+            "spills_to_disk": True,
+        },
+    ],
+)
+def test_device_spill(params):
```
mrocklin: It is also sometimes common to include many keywords in parametrized tests.  Here is an example:

https://github.com/dask/dask/blob/8ca553dcf701a3f5580939e9da8eba487a24edb3/dask/dataframe/tests/test_dataframe.py#L278-L279

I think that what is here is good too.  I just wanted to make sure that you were aware of the other option.

dask_cuda/dask_cuda_worker.py
```diff
@@ -252,6 +264,14 @@ def del_pid_file():
...
             env={"CUDA_VISIBLE_DEVICES": cuda_visible_devices(i)},
             name=name if nprocs == 1 or not name else name + "-" + str(i),
+            data=DeviceHostFile(
+                device_memory_limit=get_device_total_memory(index=i)
+                if (device_memory_limit == "auto" or device_memory_limit == int(0))
+                else parse_bytes(device_memory_limit),
+                memory_limit=parse_memory_limit(
+                    memory_limit, nthreads, total_cores=nprocs
+                ),
+            ),
```
mrocklin: I recommend the following instead:

```python
data=(DeviceHostFile, {...})
```

Otherwise we create the data object immediately here, and then need to pass it down to the worker through a process boundary.

This is implemented here, but it looks like it's not documented anywhere (my mistake)

https://github.com/dask/distributed/blob/8e449d392e91eff0a3454ee98ef362de8f78cc4f/distributed/worker.py#L500-L501

Also, if the user specifies `device_memory_limit=0` then we might want something simpler.  I can imagine wanting to turn off this behavior if things get complex.

We probably also want the same treatment in `LocalCUDACluster`, though as a first pass we can also include this ourselves.

dask_cuda/tests/test_spill.py
pentschev: Yes, that was accidental.

dask_cuda/dask_cuda_worker.py
pentschev: I was in the process of doing that, only saw your comment now. Indeed, this is a better solution, but either way, the downside is that I had to create the work directory in `dask-cuda-worker`, which I'm not particularly happy with, so if you have a suggestion on how to avoid that, it would be great!

I'm already working on the `LocalCUDACluster`, I'm still not done, will soon push those.

mrocklin: > the downside is that I had to create the work directory in dask-cuda-worker, which I'm not particularly happy with, so if you have a suggestion on how to avoid that, it would be great!

Hrm, yes I can see how that would be a concern.  I don't have a good solution currently unfortunately.  I'll think about it though.

dask_cuda/tests/test_spill.py
pentschev: Yes, I'm aware of that, but I wanted to be clear what each means while I was writing the code. If you prefer, I wouldn't mind changing it now.

pentschev:
@mrocklin I think this is good now, tests in place too. IMO, the only thing to be changed is how to create the worker directory if we can think of a way, but besides that, it's probably good for a review.

mrocklin:
Looks good to me.  Thank you for putting in the effort here @pentschev .  I'm looking forward to using this :)

pentschev:
Thanks for the review @mrocklin!

PR #67, opened 2019-06-04
Fix LocalCUDACluster data spilling and its test

pentschev:
There are two issues fixed here:

1. Creating the local workspace doesn't work, which gets destroyed and recreated by `Worker`;
2. It turns out that the test for `LocalCUDACluster` wasn't running at all, since the internal function didn't get called, and it had some other issues, such as the missing usage of `loop` and handling the proper closure of objects.

### DIFF
--- a/dask_cuda/tests/test_spill.py
+++ b/dask_cuda/tests/test_spill.py
@@ -10,7 +10,6 @@ from dask_cuda.device_host_file import DeviceHostFile
 import dask.array as da
 import cupy
 from zict.file import _safe_key as safe_key
-from tornado.ioloop import IOLoop
 
 
 def assert_device_host_file_size(dhf, total_bytes, chunk_overhead=1024):
@@ -111,11 +110,8 @@ def test_device_spill(params):
 def test_cluster_device_spill(params):
     @gen_test(timeout=30)
     def test():
-        loop = IOLoop.current()
-
         cluster = yield LocalCUDACluster(
             1,
-            loop=loop,
             scheduler_port=0,
             processes=True,
             silence_logs=False,
--- a/dask_cuda/tests/test_spill.py
+++ b/dask_cuda/tests/test_spill.py
@@ -1,3 +1,7 @@
+import pytest
+
+import os
+
 from distributed.utils_test import gen_cluster, loop, gen_test
 from distributed.worker import Worker
 from distributed import Client, get_worker, wait
@@ -5,9 +9,8 @@ from dask_cuda import LocalCUDACluster
 from dask_cuda.device_host_file import DeviceHostFile
 import dask.array as da
 import cupy
-import pytest
-import os
 from zict.file import _safe_key as safe_key
+from tornado.ioloop import IOLoop
 
 
 def assert_device_host_file_size(dhf, total_bytes, chunk_overhead=1024):
@@ -105,16 +108,20 @@ def test_device_spill(params):
         },
     ],
 )
-def test_cluster_device_spill(loop, params):
-    @gen_test()
-    def test_cluster_device_spill():
+def test_cluster_device_spill(params):
+    @gen_test(timeout=30)
+    def test():
+        loop = IOLoop.current()
+
         cluster = yield LocalCUDACluster(
             1,
+            loop=loop,
             scheduler_port=0,
             processes=True,
             silence_logs=False,
             dashboard_address=None,
             asynchronous=True,
+            death_timeout=10,
             device_memory_limit=params["device_memory_limit"],
             memory_limit=params["memory_limit"],
             memory_target_fraction=params["host_target"],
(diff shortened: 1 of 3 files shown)


dask_cuda/tests/test_spill.py
```diff
@@ -147,3 +154,13 @@ def get_data(worker, total_size):
                 assert dc > 0
             else:
                 assert dc == 0
+
+        yield client.close()
+
+        # Ensure timeout is respected for workers
+        for w in cluster.workers:
+            yield w.close(timeout=10)
```
mrocklin: Why was this necessary?  Is `yield cluster.close()` not enough?

dask_cuda/tests/test_spill.py
```diff
@@ -105,16 +108,20 @@ def test_device_spill(client, scheduler, worker):
...
-    @gen_test()
-    def test_cluster_device_spill():
+def test_cluster_device_spill(params):
+    @gen_test(timeout=30)
+    def test():
+        loop = IOLoop.current()
+
         cluster = yield LocalCUDACluster(
             1,
+            loop=loop,
```
mrocklin: I would expect this to use `IOLoop.current()` by default if asynchronous=True.  Is this not the case?

dask_cuda/tests/test_spill.py
pentschev: No, for some reason the timeout isn't respected, and I had identified where this wasn't working but now why. Now I realized that `LocalCluster` was changed significantly in distributed master, so maybe that would work now. Since we're still on 1.28.1 here, I would keep this and raise an issue to verify this with the new distributed release (1.29?), do you agree with that?

dask_cuda/tests/test_spill.py
pentschev: You're right, we don't need it, at some point I was probably doing something wrong and it worked with this, I'm removing it now.

dask_cuda/tests/test_spill.py
mrocklin: I plan to change our testing here to point to distributed master pretty soon, so we'll have to handle that then.  Lets wait a day for that to happen and then check in here again.

pentschev: Ok, I will try doing that tomorrow then.

pentschev:
@mrocklin when you have some time to take a look at this, please feel free to comment or merge it if there's no other concerns here.

pentschev:
@mrocklin when you have some time to take a look at this, please feel free to comment or merge it if there's no other concerns here.

dask_cuda/tests/test_spill.py
```diff
@@ -106,44 +108,49 @@ def test_device_spill(client, scheduler, worker):
...
+                await client.run(get_data, cluster.workers[0].Worker, x.nbytes)
+                disk_chunks = await client.run(lambda: len(get_worker().data.disk))
+                for dc in disk_chunks.values():
+                    if params["spills_to_disk"]:
+                        assert dc > 0
+                    else:
+                        assert dc == 0
+
+                await client.close()
+                await cluster.close()
```
mrocklin:
```suggestion
```

dask_cuda/tests/test_spill.py
```diff
@@ -106,44 +108,49 @@ def test_device_spill(client, scheduler, worker):
...
+
+                await client.run(get_data, cluster.workers[0].Worker, x.nbytes)
+                disk_chunks = await client.run(lambda: len(get_worker().data.disk))
+                for dc in disk_chunks.values():
+                    if params["spills_to_disk"]:
+                        assert dc > 0
+                    else:
+                        assert dc == 0
+
+                await client.close()
```
mrocklin:
```suggestion
```

These are closed as part of the `async with` context manager

dask_cuda/tests/test_spill.py
```diff
@@ -106,44 +108,49 @@ def test_device_spill(client, scheduler, worker):
...
+                await cluster.close()
 
-        yield client.run(get_data, cluster.workers[0].Worker, x.nbytes)
-        disk_chunks = yield client.run(lambda: len(get_worker().data.disk))
-        for dc in disk_chunks.values():
-            if params["spills_to_disk"]:
-                assert dc > 0
-            else:
-                assert dc == 0
+    loop.run_sync(test)
```
mrocklin: The other option is to mark the test function as follows:

```python
@pytest.mark.asyncio
async def test_foo(...):
    normal async code here
```

This manages setting up the IOLoop and running it for you.   What we have here is good too though, and also fairly common.

pentschev: This requires pytest-asyncio or some similar package, is this generally fine to be added?

mrocklin:
Yes

pentschev:
Thanks for the review and suggestions @mrocklin, merging.

PR #71, opened 2019-06-11
[REVIEW] Add test skipping functionality to build.sh

dillon-cullinan:


pentschev:
LGTM. Just for my own understanding, when/how is `--skip-tests` is defined in the CI context?

dillon-cullinan:
> LGTM. Just for my own understanding, when/how is `--skip-tests` is defined in the CI context?

In our Jenkins builds we call `build.sh` to run via bash, the argument can be passed in then. This won't apply to PR's, though. This is for builds that do not need to run tests and just need a source build to work (such as building docs).

dillon-cullinan:
Good to merge!

PR #78, opened 2019-06-26
LocalCUDACluster calls _correct_state() to ensure workers started

pentschev:
This fix ensures `LocalCUDACluster` properly informs on workers immediately after creation. A proper fix should be in dask/distributed, but as a quick workaround I'm adding this here.

@cjnolet this should fix your issue and allow pinning dask-cuda=0.8 back in.

cc @mrocklin @dantegd @raydouglass

### DIFF
--- a/dask_cuda/local_cuda_cluster.py
+++ b/dask_cuda/local_cuda_cluster.py
@@ -104,6 +104,7 @@ class LocalCUDACluster(LocalCluster):
 
         self.cuda_visible_devices = CUDA_VISIBLE_DEVICES
         self.scale(n_workers)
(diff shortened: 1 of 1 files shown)


pentschev:
Thanks @cjnolet for checking. Merging this now.

pentschev:
Following on our discussion in https://github.com/dask/distributed/pull/2808, @mrocklin would you mind checking whether the change in this PR is ok or if there is a more appropriate one? Since this was a release blocker, I merged it in, but I don't know whether this is the best solution.

mrocklin:
I think it's fine.  It's probably good in the sync case and harmless in the async case.

PR #97, opened 2019-07-25
Update cudf, dask, dask-cudf, distributed version requirements

pentschev:
It seems that CI is failing because it's using the wrong version of distributed, even though the looks seem to show that it's up-to-date. Hopefully this will fix that.

cc @mike-wendt

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -23,9 +23,9 @@ export HOME=$WORKSPACE
 # Switch to project root; also root of repo checkout
 cd $WORKSPACE
 
-# Get latest tag and number of commits since tag
-export GIT_DESCRIBE_TAG=`git describe --abbrev=0 --tags`
-export GIT_DESCRIBE_NUMBER=`git rev-list ${GIT_DESCRIBE_TAG}..HEAD --count`
+# Parse git describe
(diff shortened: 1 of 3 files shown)


ci/gpu/build.sh
```diff
@@ -50,10 +50,8 @@ g++ --version
 conda config --set ssl_verify False
 
 conda install \
-    'cudf=0.8' \
-    'dask-cudf=0.8'
-
-pip install git+https://github.com/dask/distributed.git@master
+    'cudf=0.9' \
+    'dask-cudf=0.9'
```
mike-wendt: Hard coding these will only end up in causing issues in the future. I believe the better approach is to use what cuDF does from lines [28-31](https://github.com/rapidsai/cudf/blob/branch-0.9/ci/gpu/build.sh#L28). With the `MINOR_VERSION` env var set they you can install the correct versions of cudf/dask-cudf based on your current version.

pentschev: Thanks for pointing that out, I'm not sure why they were hardcoded. This should be resolved in the newest commit.

pentschev:
Thanks for the review @mike-wendt, merging now.

PR #98, opened 2019-07-25
Improve device memory spilling performance

### LINKED ISSUE #92: Spill quickly from device to host
I'm curious how fast we are at moving data back and forth between device and host memory.  I suspect that currently we end up serializing with standard Dask serialization, which may not be aware of device memory.  We should take a look at using the `"cuda"` serialization family within Dask, that was added as part of the UCX work.   

```python
from distributed.protocol import serialize, deserialize
x = cupy.arange(100)
header, frames = serialize(x, serializers=("cuda", "dask", "pickle"))
y = deserialize(header, frames, deserializers=("cuda", "dask", "pickle",    "error"))
```

(more here https://github.com/dask/distributed/tree/master/distributed/protocol/tests)

And then we should probably do a bit of profiling.  I would focus this on just the `DeviceHostDisk` class, rather than engaging the full dask.distributed cluster machinery.

This is relevant for a couple of applications that people care about today

1.  Larger than memory image processing (a bit)
2.  Out-of-core sorts and joins (cc @randerzander @kkraus14) (more relevant)

@pentschev , you seem like the obvious person to work on this if you have some time.  I think it'd be useful to know how fast we are today, then look at serialization with the new "cuda" Dask serializer, and then see how fast we can get things relative to theoretical hardware speeds.

pentschev:
Fixes #92 

I compiled some performance numbers before and after this PR, using the code from https://github.com/rapidsai/dask-cuda/pull/65#issuecomment-501921953 on a 16GB V100:

Before:
Run 1: 428 seconds
Run 2: 429 seconds

After:
Run 1: 233 seconds
Run 2: 235 seconds

cc @mrocklin @randerzander


dask_cuda/device_host_file.py
```diff
@@ -1,28 +1,39 @@
...
 
-def _deserialize_if_device(obj):
-    """ Deserialize an object if it's an instance of bytes """
-    if isinstance(obj, bytes):
-        return deserialize_bytes(obj)
-    else:
+def _deserialize(obj):
+    """ Deserialize an object if it's a serialized object, thus assumes it's a tuple of
+    length 2, and moves frames to device if the serialized object is a CUDA object """
+    if not isinstance(obj, tuple):
```
mrocklin: Is there a situation where this isn't true?

dask_cuda/device_host_file.py
```diff
@@ -1,28 +1,39 @@
...
-def _serialize_if_device(obj):
-    """ Serialize an object if it's a device object """
-    if is_device_object(obj):
-        return serialize_bytes(obj, on_error="raise")
-    else:
-        return obj
+def _serialize(obj):
+    """ Serialize an object and moves frames to host if it's a device object """
+    header, frames = serialize(
+        obj, serializers=["cuda", "dask", "pickle"], on_error="raise"
```
mrocklin: Are there situations where we would use the dask or pickle serializers here?

dask_cuda/utils.py
```diff
@@ -21,3 +22,25 @@ def get_device_total_memory(index=0):
...
         return cuda.current_context().get_memory_info()[1]
+
+
+def convert_frames_to_numba(frames):
+    try:
+        import cupy
+    except ImportError:
+        return frames
+
+    return [cuda.as_cuda_array(f) if isinstance(f, cupy.ndarray) else f for f in frames]
```
mrocklin: Why the cupy check here?  I would expect numba to rely on the `__cuda_array_interface__` here rather than explicit type checking.  Does this not work today?  If not, can we raise an issue upstream?

dask_cuda/utils.py
```diff
@@ -21,3 +22,25 @@ def get_device_total_memory(index=0):
...
+    # Conversion of frames to Numba can/should be eventually done during the
+    # serialization process in dask/distributed
+    frames = convert_frames_to_numba(frames)
+
+    return [f.copy_to_host() if isinstance(f, cuda.cudadrv.devicearray.DeviceNDArray)
+            else f for f in frames]
+
+
+def move_frames_to_device(frames):
+    return [cuda.to_device(f) if isinstance(f, np.ndarray) else f for f in frames]
```
mrocklin: What are the other types of frames that might be here?

dask_cuda/device_host_file.py
pentschev: There was a situation of `obj` being an `int`. I should probably check why is that.

dask_cuda/device_host_file.py
pentschev: Yes, we also use this for serializing host objects into the `host_buffer`. But maybe "pickle" could be removed, is there any host-side objects that can't be serialized with "dask" serializer?

mrocklin: Ah, I figured that this was just for `device<->host`.  Never mind then

mrocklin: "dask" and "pickle" go together.  The "dask" serializer just special-cases a few host-side objects to get better performance.  There are many host-side objects that the "dask" serializer doesn't handle.

dask_cuda/utils.py
pentschev: cuDF dataframes have various frames, some of them are not Numba device arrays. Example:

```python
In [1]: import pandas as pd, cudf

In [2]: from distributed.protocol import serialize

In [3]: pdf = pd.DataFrame({'a': [0, 1, 2, 3],'b': [0.1, 0.2, None, 0.3]})

In [4]: df = cudf.from_pandas(pdf)

In [5]: serialize(df, serializers=["cuda"])[1]
Out[5]:
[<numba.cuda.cudadrv.devicearray.DeviceNDArray at 0x7f58673875c0>,
 b'\x80\x04N.',
 <numba.cuda.cudadrv.devicearray.DeviceNDArray at 0x7f580ab45860>,
 b'\x80\x04N.',
 <numba.cuda.cudadrv.devicearray.DeviceNDArray at 0x7f580ab60e80>,
 <numba.cuda.cudadrv.devicearray.DeviceNDArray at 0x7f580ab607f0>]
```

This is why I'm avoiding traps of frames being non-device objects.

dask_cuda/utils.py
pentschev: Same as above, since not all frames are necessarily device-objects, we have to make sure not to deserialize them.

dask_cuda/device_host_file.py
pentschev: Thanks for the clarification!

dask_cuda/utils.py
mrocklin: In that case rather than type check, maybe we can check for the presense of `__cuda_array_interface__`.  In general I think that we should avoid explicit type checks when possible.  For example, if someone wants to use Dask with PyTorch or TensorFlow or some other CUDA library in the future, we shouldn't have to modify this code for them.

Perhaps we also record which of the frames we converted as a list of booleans, and include that information in the header.

pentschev: Yes, that's a good idea, thanks for the suggestion. I'll fix that.

pentschev:
@mrocklin alright, I refactored the `DeviceHostFile` code a bit. I believe it's now a bit clearer to understand all the serialization steps.

dask_cuda/device_host_file.py
```diff
@@ -76,18 +115,22 @@ def __init__(
 
     def __setitem__(self, key, value):
         if is_device_object(value):
-            self.device_buffer[key] = value
+            self.device_buffer[key] = _device_serialize(value)
```
mrocklin: It looks like we're serializing objects as they arrive.  This seems odd to me.  I would only expect to serialize objects as we needed to move between one of the different memory spaces (host, device, disk).

dask_cuda/device_host_file.py
```diff
@@ -1,28 +1,67 @@
...
+    """ Deserialize a tuple of header and frames containing a device object, returning
+    the device object. """
+    return deserialize(header, frames)
+
+
+def _host_serialize(obj):
+    """ Serialize host objects or move an already serialized device object to host
+    memory. Always returns a tuple of header and frames.
+    """
+    if isinstance(obj, tuple):
```
mrocklin: What if the user puts a tuple into host storage themselves?

mrocklin:
I'm having difficulty following all of the possible conditions of what can be where, so I thought I'd summarize the proposed model a bit (this is mostly to help myself think).  There are three different storage pools:

1.  Device: Always contains normal Python objects (cudf dataframes, cupy arrays, and so on)
2.  Host: Now contains both normal Python objects  (pandas dataframes, numpy arrays, ...) and now also these SerializedDeviceData pieces, which effectively hold just header and frames.
3.  Disk: Always holds serialized data

It seems like a lot of the odd conditions in the code here come from our double-use of Host, both as storage of normal objects and for serialized device data.  We could imagine separating this out to two different host-storage pools, but then there would be some complexity around tracking the total number of bytes that we were using on host memory.

I wonder if we might reduce the complexity in the code here by making an explicit `SerializedDeviceData` type for device data that is being stored on the host.  Hopefully if we then implement all of the standard things like serialization and nbytes functions for that type then we can avoid all of this `isinstance(obj, tuple)` and such, which I suspect will fail if users store tuples (which they will).

We use one function each to go between device and host.  We box header and frames into a `Serialized` object in order to avoid ambiguity with user data.  My hope is that this integrates well with all of the other functionality, like host<->disk without need for special casing.


```python
def device_to_host(obj: object) -> Serialized:
    header, frames = serialize(obj, serializers=["cuda"])
    is_cuda = [hasattr(frame, "__cuda_device_array__") for frame in frames]
    frames = [numba.cuda.as_cuda_array(frame) if ic else frame for ic, frame in zip(is_cuda, frames)]
    header = {"sub-header": header, "is-cuda": is_cuda}
    return SerializedDeviceData(header, frames)

def host_to_device(s: Serialized) -> object:
    is_cuda = s.header["is-cuda"]
    header = s.header["sub-header"]
    frames = [numba.cuda.to_device(frame) if ic else frame for ic, frame in zip(is_cuda, s.frames)]
    return deserialize(header, frames, serializers=["cuda"])

@dask.sizeof.register(Serialized)
def sizeof_serialized(obj):
    return sum(map(dask.utils.nbytes, obj.frames))
```

This uses

https://github.com/dask/distributed/blob/909a943b67b6b472a2d77afa13a8caa61f25f972/distributed/protocol/serialize.py#L226-L253

Thoughts?

pentschev:
Would `SerializedDeviceData` have its own serializer/deserializer?

To clarify, I ask that because the issue with the approach your proposed is on `deserialize_bytes`, which will try to deserialize CUDA frames directly when deserializing from disk.

mrocklin:
Oops, I was intermixing `Serialized` and `SerializedDeviceData` terms.  At first I thought that we might need a new type for boxing up serialized data from the device, but then I realized that we already have one, `Serialized` (see the link at the bottom of the last comment).  I've changed the code above to use `Serialized` uniformly rather than `SerializedDeviceData`.

> Would SerializedDeviceData have its own serializer/deserializer?

The `Serialized` type may have this already I think?  It  just passes things through.  Looking at the code it looks like this is special-cased in the serialize function

https://github.com/dask/distributed/blob/208e0bc2d313809b41e0998b75b0996cb39390b2/distributed/protocol/serialize.py#L137-L138

But may not be special cased the other way around (The Serialized object is used today in some odd ways, and so is a bit of a special case).

> Would SerializedDeviceData have its own serializer/deserializer?

Anyway, the answer to this is "yes, it would".  Either because it has one already, or because we would need to add one upstream.

> To clarify, I ask that because the issue with the approach your proposed is on deserialize_bytes, which will try to deserialize CUDA frames directly when deserializing from disk.

Ah, but at this point the `Serialized` object is only holding normal host-based frames.  Note that we converted the cuda frames to host frames here, and recorded which ones we converted.

```python
def device_to_host(obj: object) -> Serialized:
    header, frames = serialize(obj, serializers=["cuda"])
    is_cuda = [hasattr(frame, "__cuda_device_array__") for frame in frames]
    frames = [numba.cuda.as_cuda_array(frame) if ic else frame for ic, frame in zip(is_cuda, frames)]
    header = {"sub-header": header, "is-cuda": is_cuda}
    return Serialized(header, frames)
```

So the Serialized objects are very normal.  They just hold normal host-based frames and they only live in the `host` dict.  They can move to disk normally just like anything else (no need for specialization) and when we need to move them back to device we know exactly which frames need to be converted back to the device side.

pentschev:
But this causes issues when converting the object into `bytes`. Example:

```python
from distributed.protocol.serialize import Serialized, serialize, serialize_bytes
from numba import cuda
import cupy

def device_to_host(obj: object) -> Serialized:
    header, frames = serialize(obj, serializers=["cuda"])
    is_cuda = [hasattr(frame, "__cuda_array_interface__") for frame in frames]
    frames = [cuda.as_cuda_array(frame) if ic else frame for ic, frame in zip(is_cuda, frames)]
    header = {"sub-header": header, "is-cuda": is_cuda}
    return Serialized(header, frames)

a = cupy.arange(4)
sa = device_to_host(a)
ssa = serialize_bytes(sa)
```

This is the output I get:

```python-traceback
  File "cupy_serialize.py", line 14, in <module>
    ssa = serialize_bytes(sa)
  File "/home/nfs/pentschev/miniconda3/envs/rapids-nightly-0.9/lib/python3.7/site-packages/distributed/protocol/serialize.py", line 381, in serialize_bytes
    return b"".join(L)
TypeError: sequence item 2: expected a bytes-like object, DeviceNDArray found
```

Is this a bug or is it intended to be this way? To be honest, there are various conditions and indirection in the serialization code that are unclear and make extending (e.g., having a multi-level cache) that difficult.

mrocklin:
```python
ssa = serialize_bytes(sa)
```

I think that we should probably never convert something like this to bytes explicitly. If the object stays on the host then it should stay in its current `Serialized(header, frames)` form.  If we're moving this object to a file then we should write each element individually, rather than incurring an additional copy on the host side.

You might want to look at how we do this in serialize_bytelist currently for the normal Worker's dict-like-object. 

https://github.com/dask/distributed/blob/208e0bc2d313809b41e0998b75b0996cb39390b2/distributed/protocol/serialize.py#L362-L374

We call `header, frames = serialize(obj)` which in this case we already have.  Then we compress frames individually, write down lengths and compressions, call `msgpack.dumps` on the header, and then hand that list off to be written down to disk.

In this way the different frames are handled individually (which is good for compression) and we never make a new copy of everything.

pentschev:
Sure, I agree with that. My question of that being a bug is, since we already have `header, frames` in the form of a `Serialized` object, wouldn't it make sense that `serialize_bytes`/`serialize_bytelist` handle that in the same way, only ignoring the `serialize` call? Or in other words, shouldn't there be a specialization for `Serialized` objects (or more simply, just a simple conditional inside the functions) to do that?

mrocklin:
> Or in other words, shouldn't there be a specialization for Serialized objects (or more simply, just a simple conditional inside the functions) to do that?

Sure.  It appears to be there already.  

```python
In [1]: from distributed.protocol import Serialized, serialize

In [2]: s = Serialized({}, [b'123'])

In [3]: serialize(s)
Out[3]: ({}, [b'123'])
```

pentschev:
@mrocklin I'm not talking about `serialize`, but `serialize_bytes`. Currently, the latter doesn't take a `Serialized` object, just an unserialized one and serializes it. In our case for this PR, we have a CUDA object that is transformed into a `Serialized` object, frames then are moved to host, and we want to be able to spill that to disk, thus `serialize_bytes` (which implies `serialize_bytelist`) support for `Serialized` would be very useful.

This is how it looks like today:

```python
def serialize_bytelist(x, **kwargs):
    header, frames = serialize(x, **kwargs)
    ...
```

This is my suggestion:

```python
def serialize_bytelist(x, **kwargs):
    if isinstance(x, Serialized):
        header = x.header
        frames = x.frames
    else:
        header, frames = serialize(x, **kwargs)
    ...
```

mrocklin:
That condition is already special-cased, but closer to the core where it can be more generally applied.  Calling `serialize` on a `Serialized` object is trivially fast because of exactly a condition like what you describe.

I would encourage you to think about using `serialize_bytelist` instead of `serialize_bytes` when moving data to disk.  I think that ideally `serialized_bytes` as a term shouldn't exist within the dask-cuda codebase.

pentschev:
> I would encourage you to think about using serialize_bytelist instead of serialize_bytes when moving data to disk.

My suggestion was modifying only `serialize_bytelist`. So I either need to change that in distributed or copy the entire function into dask-cuda to change it to work with `Serialized`. Which one are you suggesting? FWIW, I think this is a modification that could be in distributed, since we're only adding support for a different type, a type that is defined in distributed itself.

mrocklin:
I think that the modification that you're suggesting is already done, and there is no need to do it further.  Calling `serialize(...)` on a `Serialized` object is already trivially fast, it does exactly what you propose today, without the need for any further modification.

pentschev:
Ok, I understand. But that then doesn't work with `zict`, because it expects all values to be strings, and with `serialize` the result will be a `dict` that can't be pickled. Is this something that we should fix in `zict` (e.g., pickling non-pickled objects)?

mrocklin:
Actually zict.File does support lists of bytes-like objects, for exactly this purpose.

See the following:

https://github.com/dask/zict/blob/f95e2314a05759c2319c842e962cd6277f67774f/zict/file.py#L28-L57

https://github.com/dask/zict/blob/f95e2314a05759c2319c842e962cd6277f67774f/zict/file.py#L79-L86

https://github.com/dask/distributed/blob/594589e7dd7112c557661a4cb890440cceaacb9c/distributed/worker.py#L519-L523

pentschev:
Not if the value is a `dict`:

```python-traceback
    def __setitem__(self, key, value):
        with open(os.path.join(self.directory, _safe_key(key)), 'wb') as f:
            if isinstance(value, (tuple, list)):
                for v in value:
>                   f.write(v)
E                   TypeError: a bytes-like object is required, not 'dict'

...
(Pdb) p v
{'serializer': 'pickle'}
```

In fact, it seems that `dict`s must be pickled, or else: https://stackoverflow.com/a/36965528/3070401

Is there a different way to do it that I don't see?

Btw, this is how we end up with one of the values being a `dict`:

```python
In [7]: serialize(serialize(np.arange(4)))
Out[7]:
({'serializer': 'pickle'},
 [b'\x80\x04\x95\xc8\x00\x00\x00\x00\x00\x00\x00}\x94(\x8c\x05dtype\x94K\x00\x8c\x03<i8\x94\x86\x94\x8c\x05shape\x94K\x04\x85\x94\x8c\x07strides\x94K\x08\x85\x94\x8c\x07lengths\x94]\x94K a\x8c\x04type\x94\x8c\rnumpy.ndarray\x94\x8c\x0ftype-serialized\x94C \x80\x04\x95\x15\x00\x00\x00\x00\x00\x00\x00\x8c\x05numpy\x94\x8c\x07ndarray\x94\x93\x94.\x94\x8c\nserializer\x94\x8c\x04dask\x94u]\x94C \x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00\x03\x00\x00\x00\x00\x00\x00\x00\x94a\x86\x94.'])
```

pentschev:
Nevermind, you're suggesting that after calling `serialize`, we call `serialize_bytelist` on the result.

That probably works, but I think that is a very cluttered and confusing code. IMO, all these changes don't really make anything clearer than the changes I proposed when I first opened this PR.

mrocklin:
If you take a look at the definition for serialize_bytelist you'll see that we use msgpack to serialize the header just before dumping.  You shouldn't need to touch anything here.  You should be able to use the current serialize_bytelist function as-is.

mrocklin:
Ah, we raced to the answer there. 

> IMO, all these changes don't really make anything clearer than the changes I proposed when I first opened this PR.

I don't think that there are that many changes here.  I think that the code I proposed above should work ok.  I also think that this approach is more robust.

pentschev:
I think everything's in now. Please take another look @mrocklin.

dask_cuda/device_host_file.py
```diff
@@ -1,28 +1,59 @@
...
+        def sizeof_numba_devicearray(x):
+            return int(x.nbytes)
+
+
+# Register sizeof for Serialized while Dask doesn't add it
+if not hasattr(sizeof, "register_serialized"):
+
+    @sizeof.register(Serialized)
+    def register_serialized(obj):
+        return sum(map(nbytes, obj.frames))
```
mrocklin: Can I ask you to also include a PR for this upstream in dask/distributed

dask_cuda/device_host_file.py
```diff
@@ -53,18 +84,15 @@ def __init__(
...
+            partial(serialize_bytelist, on_error="raise"), deserialize_bytes, File(path)
         )
         self.host_buffer = Buffer(
             self.host_func, self.disk_func, memory_limit, weight=weight
         )
 
         self.device_func = dict()
-        self.device_host_func = Func(
-            _serialize_if_device, _deserialize_if_device, self.host_buffer
-        )
```
mrocklin: It looks like this disappeared.  I would expect `host_to_device` and `device_to_host` to be used here I think.

dask_cuda/device_host_file.py
```diff
@@ -83,8 +111,11 @@ def __setitem__(self, key, value):
     def __getitem__(self, key):
         if key in self.host_buffer:
             obj = self.host_buffer[key]
-            del self.host_buffer[key]
-            self.device_buffer[key] = _deserialize_if_device(obj)
+            if isinstance(obj, Serialized):
+                del self.host_buffer[key]
+                self.device_buffer[key] = host_to_device(obj)
+            else:
+                return obj
```
mrocklin: I'm curious if there is a way to offload these checks onto the `device_buffer` object entirely.  The buffer object is supposed to be able to get keys regardless of if they are in the fast or the slow storage.  So in theory I would hope that the `__getitem__` call here would look something like the following:

```python
def __getitem__(self, key):
    if key in self.device_buffer:
        return self.device_buffer[key]  # automatically calls host_to_device if necessary
    else:
        return self.host_buffer[key]  # automatically calls disk_to_host if necessary
```

I'm likely missing things here though.   For the plan Worker's storage, we do this by specifying functions that map both ways in the `Func` mapping:

https://github.com/dask/distributed/blob/65001f2d1c796f652bf77bbf222cf900bf1062b8/distributed/worker.py#L519-L523

dask_cuda/device_host_file.py
pentschev: Yes, will do that for this and for the `DeviceNDArray` (into dask core) later today.

dask_cuda/device_host_file.py
pentschev: Yes, my mistake when doing it while tired. I fixed that in the latest commits.

dask_cuda/device_host_file.py
pentschev: Indeed, you're right. Only the condition has to be inverted (`host_buffer` must come first). I added this in the latest commits too. Thanks for noticing.

pentschev:
There's one strange issue that I don't know if it's a problem with `DeviceHostFile` or somewhere in dask/distributed, perhaps you can shed some light here @mrocklin . In the end of the cuDF spilling tests, we delete all dataframes and check whether `worker.data` is empty https://github.com/rapidsai/dask-cuda/blob/branch-0.9/dask_cuda/tests/test_spill.py#L327-L329. For the case where we don't spill to disk, it works, but when we do, all the chunks that are _not_ located on the `zict.File` object get deleted, and those on disk get ignored.

Printing all calls to `DeviceHostFile.__delitem__` shows that there's a call for all chunks that are not on disk, but those that are never get called. For the case above, `from_pandas` chunks `4` onwards get deleted, but chunks `0-3` remain on disk and never even try to be deleted.

Any ideas on what could be the issue here?

dask_cuda/device_host_file.py
```diff
@@ -82,14 +127,9 @@ def __setitem__(self, key, value):
 
     def __getitem__(self, key):
         if key in self.host_buffer:
-            obj = self.host_buffer[key]
-            del self.host_buffer[key]
-            self.device_buffer[key] = _deserialize_if_device(obj)
-
-        if key in self.device_buffer:
-            return self.device_buffer[key]
+            return self.host_buffer[key]
```
mrocklin: Just to make sure, if I put a device-backed object into a DeviceHostDisk and it gets moved to the host for space reasons, will that key show up in `self.host_buffer`?

dask_cuda/device_host_file.py
```diff
@@ -1,26 +1,71 @@
...
     else:
         return obj
 
 
 def _deserialize_if_device(obj):
-    """ Deserialize an object if it's an instance of bytes """
-    if isinstance(obj, bytes):
-        return deserialize_bytes(obj)
+    if isinstance(obj, Serialized):
+        return host_to_device(obj)
```
mrocklin: Will the else condition of this check ever occur?  Are there cases when place device objects into host without serializing them?  If this shouldn't occur then perhaps we could turn this check into an assertion?

mrocklin:
> In the end of the cuDF spilling tests, we delete all dataframes and check whether worker.data is empty 

My first attempt here would be to remove Dask from the situation, and place cudf objects into a DeviceHostDisk object, delete all of those keys, and then verify that the directory it used was empty.  Does this function as expected?  

If this does work then my next suggestion would be to put a `breakpoint()` where we  would expect this action to take place and inspect state a bit.

Or, in short, nothing obvious comes to mind, but hopefully additional testing of the components reveals something useful.

dask_cuda/device_host_file.py
pentschev: Yes, `self.host_buffer` is the de-facto `self.device_buffer.slow`.

dask_cuda/device_host_file.py
pentschev: > Will the else condition of this check ever occur?

I expect so.

> Are there cases when place device objects into host without serializing them?

This case shouldn't exist.

> If this shouldn't occur then perhaps we could turn this check into an assertion?

 It should always be the case that a device object will be an instance of `Serialized` and we would want to move it back to device. However, as I was trying to confirm this, none of the tests seem to call `_deserialize_if_device` at any point, which is weird, I'm now investigating it.

dask_cuda/device_host_file.py
mrocklin: In that case shouldn't we check the device Buffer first so that it pulls the Serialized object currently residing in `self.host_buffer` back down to the device and then return that deserialized object?  I'm naively proposing switching the order of conditions here:

```python
if key in self.device_buffer:
    return self.device_buffer[key]
if key in self.host_buffer:
    return self.host_buffer[key]
raise KeyError(key)
```

pentschev:
> My first attempt here would be to remove Dask from the situation, and place cudf objects into a DeviceHostDisk object, delete all of those keys, and then verify that the directory it used was empty. Does this function as expected?

At the moment, some of the `test_device_host_file` tests are failing, I'm trying to fix them now. Maybe that will reveal something.

> If this does work then my next suggestion would be to put a `breakpoint()` where we would expect this action to take place and inspect state a bit.
> 
> Or, in short, nothing obvious comes to mind, but hopefully additional testing of the components reveals something useful.

Thanks, I will continue to investigate things.

pentschev:
Alright, everything seems to work now, including the issues I mentioned earlier today.

Here are some clarifications on the latest commit:

1. Deserialization of the `Serialized` object we created (that with a `sub-header`) doesn't work directly, by which I mean `deserialize_bytes(serialize_bytelist(device_to_host(device_object)))`. However, if we serialize the result of `device_to_host`, that works, I haven't checked in detail, but I think the reason for that is we then have a common type (such as a `pickle`d object). Due to that, I had to create a deserializer (in `deserializer_bytes`). Should we perhaps create a pair of `dumps`/`loads` for `Serialized` objects? Note that this would probably require us to create a new type `DeviceSerialized` (such as you've suggested before) to allow us to differentiate between other possible non-device `Serialized` objects.
1. I was wrong, and we can't do what is suggested in https://github.com/rapidsai/dask-cuda/pull/98#discussion_r307910899. The reason we need the more complex code is that we must ensure we never accidentally move host objects (such as `np.ndarray`) into `self.device_buffer`, which would happen if we call `self.device_buffer["np.ndarray-key"]`. If we just invert the order as I proposed, we lose the ability of moving a device object that is currently on host/disk back to the device.

mrocklin:
Yeah, sorry about that. I suspect that `Serialized` is doing something odd due to other special cases in which it's used.  Making our own `DeviceSerialized` type that was not a subclass (there are `isinstance` checks which respect inheritence in the mainline codebase) but was essentially the same, along with trivial implementations for `dask_serialize` and `dask_deserialize` for the type might work well.

```python
class DeviceSerialized:
    def __init__(self, header, frames):
        self.header = header
        self.frames = frames

from distributed.protocol import dask_serialize, dask_deserialize

@dask_serialize.register(DeviceSerialized):
 def _(ds):
    return ds.header, ds.frames

@dask_deserialize.register(DeviceSerialized):
def _(header, frames):
    return DeviceSerialized(header, frames)
```

The only purpose of these functions is to move these host-side frames down to disk, so they should almost be a no-op, just a small boxing/unboxing that doesn't make any copies.

mrocklin:
> ensure we never accidentally move host objects (such as np.ndarray) into self.device_buffer, which would happen if we call `self.device_buffer["np.ndarray-key"]`

But we should never do this though, right?

pentschev:
I was trying to implement your suggestion, but there's always a difficult-to-debug case. Right now the issue seems to be that we can't deserialize:

```python
(Pdb) p dfssa
<dask_cuda.device_host_file.DeviceSerialized object at 0x7f7b89c3a5f8>
(Pdb) p deserialize(dfssa.header, dfssa.frames)
<dask_cuda.device_host_file.DeviceSerialized object at 0x7f7b89c3add8>
```

I guess what happens is that the deserialization function always returns the data wrapped in another `DeviceSerialized`, so we will end up in a loop (even though we don't have an infinite recursion in this case).

In the latest commit here, we had to do `Serialized(*deserialize_bytes(b))`, so the `deserialize` function (contained in `deserialize_bytes` was another, probably `dask_loads`), which was then wrapped in `Serialized`. I don't see yet another way of solving this, let me know if you have an idea @mrocklin.

> > ensure we never accidentally move host objects (such as np.ndarray) into self.device_buffer, which would happen if we call `self.device_buffer["np.ndarray-key"]`
> 
> But we should never do this though, right?

I'm not sure I understood your question, but I'm assuming that you're asking: "We should never have a host object moved into `self.device_buffer`, correct?". If that's your question, the answer is: yes.

mrocklin:
```
(Pdb) p dfssa
<dask_cuda.device_host_file.DeviceSerialized object at 0x7f7b89c3a5f8>
(Pdb) p deserialize(dfssa.header, dfssa.frames)
<dask_cuda.device_host_file.DeviceSerialized object at 0x7f7b89c3add8>
```

This seems like what I would expect.  We're only using the serialize/deserialization routines to move from host to disk and back, not to rehydrate into a device object.  For that we're using the `host_to_device` function you have here:

```python
def device_to_host(obj: object) -> Serialized:
    header, frames = serialize(obj, serializers=["cuda"])
    is_cuda = [hasattr(f, "__cuda_array_interface__") for f in frames]
    frames = [
        cuda.as_cuda_array(f).copy_to_host() if ic else f
        for ic, f in zip(is_cuda, frames)
    ]
    header = {"sub-header": header, "is-cuda": is_cuda}
    return Serialized(header, frames)


 def host_to_device(s: Serialized) -> object:
    is_cuda = s.header["is-cuda"]
    header = s.header["sub-header"]
    frames = [cuda.to_device(f) if ic else f for ic, f in zip(is_cuda, s.frames)]
    return deserialize(header, frames)
```

(though presumably we would replace `Serialized` with `DeviceSerialized`)

It seems like I have something specific in mind, and we're losing time in commication.  Perhaps I should just give this a shot?

pentschev:
> It seems like I have something specific in mind, and we're losing time in commication. Perhaps I should just give this a shot?

Go ahead. FWIW, I changed all `Serialized` entries to `DeviceSerialized`. Also, make sure that both `test_device_host_file.py` and `test_spill.py` pass, it can be very annoying that issues may arise in only a few tests at a time.

pentschev:
@mrocklin thanks for the changes. Everything is in, including my fix to the last remaining issue and extension of the test to check for that. Please take a look when you have the chance.

mrocklin:
Everything here looks good to me.  Thanks for keeping up with this @pentschev

mrocklin:
I'm curious what you think next steps would be here to improve bandwidth. Given that we should be able to allocate memory on the host side at 10-20 GB/s, what are some things that we should investigate to see what the next bottleneck is?

pentschev:
Honestly, I don't have many ideas. I think maintaining a memory pool has many setbacks, so it's probably not a good idea. Apart from that, my only other idea would be to make allocation much faster (e.g., by writing a single byte to all pages beforehand to avoid page faults), I have a gist on how to do that with Cython:

https://gist.github.com/pentschev/cf616610646c9de5ab08e4dc130bd8b3

With that solution, we would still need a way to return a NumPy array with a memory buffer created like that, and I don't know if that's currently possible with NumPy, but perhaps we could write our own NumPy array by ensuring the ABI (or whatever it's called in Python) of an object we create is compatible with NumPy's.

mrocklin:
@madsbk any thoughts?

pentschev:
@mrocklin I also saw workers pausing when running `ci/gpu/build.sh` locally, but for some reason it didn't happen in my development environment. So now I disabled all pausing since we don't want that to happen anyway.

madsbk:
I don't know if I have the full picture, but the performance of `device_host_file.device_to_host()` is definitely bounded by the host memory allocation. I have rewritten the code a but to seperate the allocation from the dev2host copy:

```python
    header, frames = serialize(obj, serializers=["cuda"])
    is_cuda = [hasattr(f, "__cuda_array_interface__") for f in frames]

    frames2 = []
    for ic, f in zip(is_cuda, frames):
        if ic:
            t1 = time.time()
            a = numpy.empty(shape=f.shape, dtype=f.dtype)
            a[::pagesize//a.itemsize] = 0  # writing to every memory page
            t2 = time.time()
            f = cuda.as_cuda_array(f).copy_to_host(a)
            t3 = time.time()
            print("alloc: ", t2 - t1, ", dev2host: ", t3 - t2, ", total: ", t3 - t1)
        frames2.append(f)
    return DeviceSerialized(header, frames2, is_cuda)
```

When running this code on my GV100 workstation:
```python
import cupy
from dask_cuda import device_host_file
x = cupy.arange(2**30//8)
device_host_file.device_to_host(x)
```
I get: `alloc:  0.3266 , dev2host:  0.0997 , total:  0.4264`, which translate into approx. 10 GB/s (dev2host).
And as @pentschev suggested, removing the `a[::pagesize//a.itemsize] = 0` line we see a performence drop to: `alloc:  0 , dev2host:  0.5479 , total:  0.5480`.

Thus, it makes sense to do the pre-allocation and write to every memory page but to further improve the performance, I think we need some kind of a memory allocation pool :/

pentschev:
Thanks @madsbk. Indeed, this seems like a good potential solution. After some tests, I could verify that this is faster on my workstation (Intel i7 @3.50GHz), but on a DGX-1 (Intel Xeon @2.2 GHZ) there's no difference whatsoever. I can't say what's the reason for that, but I think the lower frequency has a toll on that.

I've also quickly hacked NumPy's memory allocation to ensure pages are allocated with `np.empty`, and it did make a difference. Below are some results.

i7 workstation (Not writing memory pages):
`alloc:  2.1219253540039062e-05 , dev2host:  0.5319631099700928 , total:  0.5319843292236328`
i7 workstation (Writing memory pages in Python):
`alloc:  0.3025069236755371 , dev2host:  0.1667640209197998 , total:  0.4692709445953369`
i7 workstation (Writing memory pages in C):
`alloc:  0.07596516609191895 , dev2host:  0.166917085647583 , total:  0.24288225173950195`

DGX-1 (Not writing memory pages):
`alloc:  1.7642974853515625e-05 , dev2host:  0.5940659046173096 , total:  0.5940835475921631`
DGX-1 (Writing memory pages in Python):
`alloc:  0.4045259952545166 , dev2host:  0.1753249168395996 , total:  0.5798509120941162`
DGX-1 (Writing memory pages in C):
`alloc:  0.22144031524658203 , dev2host:  0.09378528594970703 , total:  0.31522560119628906`

To complement the results on the initial comment of this PR, I reran the test on the latest commit, it took 233-235 seconds, and adding in the NumPy hack we go down to 213-215 seconds.

mrocklin:
We're currently at 1.5 GB/s.  If we could get to 10GB/s without a ton of effort that would be a big improvement.  Probably big enough to not worry about this until someone asked about it again.

pentschev:
> We're currently at 1.5 GB/s.

This depends heavily on the system, as results above show. The latest result (with NumPy hack) on the DGX-1 gives us a little over 3 GB/s, but to really have that, we would either have to modify NumPy or figure a way of allocating memory of a NumPy array ourselves (such as with Cython).

> If we could get to 10GB/s without a ton of effort that would be a big improvement. Probably big enough to not worry about this until someone asked about it again.

Yes, that would be the realistic bandwidth we can get with the PCIe bus, so we wouldn't have to worry about that anymore. :)

Anything beyond the results I got with NumPy pre-allocating pages will probably require some sort of host memory pooling to ensure we have the host memory pre-allocated when spilling is triggered, and that's going to be complicated too, we can't just keep a pool of memory buffers since Numba expects that the destination NumPy array matches shape and dtype of the device array.

If we decide to go to a memory pooling course, I'd say we should have an extension to Numba's `copy_to_host` that would take some sort of memory view of the pool where it could write the array's contents to without creating any specific object, just like pickle does when it returns a `bytes` object, but avoiding concatenations too.

mrocklin:
I've raised an issue upstream with NumPy here: https://github.com/numpy/numpy/issues/14177

mrocklin:
Does the 2-3GB/s match your expectations for how fast we should be able to allocate arrays in Host memory?  If not, what might be going on?  I'd encourage you both to engage with thoughts on that issue.

pentschev:
> Does the 2-3GB/s match your expectations for how fast we should be able to allocate arrays in Host memory? If not, what might be going on? I'd encourage you both to engage with thoughts on that issue.

Note that this measurement is a serial processing of memory allocation and copy to host, so we will never be able to nullify the compute time of the first. To do that, we would need to resort to some memory pooling to ensure we never allocate memory during spill.

If we want to be really truthful w.r.t. bandwidth, from the numbers I posted before, we're roughly at 5 GB/s for host memory allocation and 10 GB/s for copying. In fact, that makes sense to me, CUDA also implements the concept of "pinned memory", from [this post](https://devblogs.nvidia.com/how-optimize-data-transfers-cuda-cc/):

> Host (CPU) data allocations are pageable by default. The GPU cannot access data directly from pageable host memory, so when a data transfer from pageable host memory to device memory is invoked, the CUDA driver must first allocate a temporary page-locked, or “pinned”, host array, copy the host data to the pinned array, and then transfer the data from the pinned array to device memory, as illustrated below.

Without pinned memory, I normally see transfers at 5-6 GB/s, and 11-12 GB/s with.

That said, if we want to achieve 10-12 GB/s, we can't account for host memory allocation, so we need to rely on a memory buffer that has been preallocated, and for that we need a memory pool.

madsbk:
@pentschev, maybe try to use [`numba.cuda.pinned_array()`](https://numba.pydata.org/numba-doc/dev/cuda/memory.html#pinned-memory) and Numba to write to each memory page?

pentschev:
> @pentschev, maybe try to use numba.cuda.pinned_array() and Numba to write to each memory page?

In the context of dask-cuda, I don't think this is going to help, since we will still incur the memory allocation when we need to spill it.

Just to be clear on my previous comment, when we copy CUDA memory back to CPU where the destination is pinned memory, the cost for allocation was paid _before_ the copy, but it doesn't mean the allocation itself is faster nor free.

madsbk:
Sure, but wouldn't it prevent an extra copy inside `copy_to_host()` because of the "the GPU cannot access data directly from pageable host memory" problem? Or am I missing something?

pentschev:
> Sure, but wouldn't it prevent an extra copy inside copy_to_host() because of the "the GPU cannot access data directly from pageable host memory" problem?

Yes, but since the memory has been paged already (assuming that we do that), then it should be capable of accessing host memory.

For the sake of certainty, I tested both pinned memory and hugepage (`/sys/kernel/mm/transparent_hugepage/enabled` set to `always`, as per my comment in https://github.com/numpy/numpy/issues/14177#issuecomment-517416320).

Using hugepage:
`alloc:  1.8358230590820312e-05 , dev2host:  0.3671739101409912 , total:  0.36719226837158203`
Using pinned memory:
`alloc:  0.38730812072753906 , dev2host:  0.08667302131652832 , total:  0.4739811420440674`

I don't know why, but pinned memory is actually slower here (I expected a similar result). Also, setting hugepage to `always` has no effect with pinned memory.

pentschev:
@mrocklin tests here are passing now. If we want to pursue additional bandwidth improvements, I would like to have this PR merged and later continue the discussion in a different issue and open a new PR for that. Would you mind reviewing, and ultimately merging this?

mrocklin:
Sure.  I've reviewed this.  Merging.

PR #101, opened 2019-08-01
Update dask-cuda for dask 2.2

mrocklin:


ci/gpu/build.sh
```diff
@@ -68,7 +68,7 @@ else
     ################################################################################
 
     pip install -e .
-    pip install pytest pytest-asyncio
+    pip install pytest pytest-asyncio fsspec
```
pentschev: Shouldn't that be in `requirements.txt` instead?

pentschev:
Looks mostly good to me, could you just update `conda/recipes/dask-cuda/meta.yaml` too?

pentschev:
Thanks @mrocklin. Merging.

Issue #128, opened 2019-09-04
LocalCUDACluster() error with dask/distributed 2.3.0 & dask_cuda nightly


cjnolet:
I started noticing this error on my local workstation in a fresh conda environment (dask/distributed 2.3.0 and dask-cuda nightly) and now it appears cuML's CI is having the same problem. 
```
        while len(self.worker_spec) < n:
>           self.worker_spec.update(self.new_worker_spec())
E           TypeError: cannot convert dictionary update sequence element #0 to a sequence
```

This traces back to the instantiation of the `LocalCUDACluster()` in some of our pytests:

```
    @pytest.mark.mg
    @pytest.mark.parametrize("nrows", [1e3, 1e5, 5e5])
    @pytest.mark.parametrize("ncols", [10, 30])
    @pytest.mark.parametrize("nclusters", [5, 10])
    @pytest.mark.parametrize("n_parts", [None, 50])
    def test_end_to_end(nrows, ncols, nclusters, n_parts, client=None):
        owns_cluster = False
        if client is None:
            owns_cluster = True
>           cluster = LocalCUDACluster(threads_per_worker=1)
```

Rolling the packages back to dask/distributed=`2.2.0` and dask-cuda=`0.9.0` seems to fix it, but we probably want to be testing against, and supporting, the most up to date packages.

PR #126, opened 2019-09-02
Merge branch-0.9 into branch-0.10

pentschev:
Fixes #120, automerge failed.

PR #129, opened 2019-09-04
Change the updated new_worker_spec API for upstream

mrocklin:
This changed upstream in https://github.com/dask/distributed/pull/3000

Fixes https://github.com/rapidsai/dask-cuda/issues/128

cc @cjnolet

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -63,6 +63,9 @@ else
     # Install CuPy for tests
     pip install cupy-cuda${CUDA_REL}==6.0.0
 
+    # Install distributed@master (temporarily required due to issues with 2.3.2)
+    pip install git+https://github.com/dask/distributed.git@master
+
     ################################################################################
     # TEST - Run tests
     ################################################################################
--- a/dask_cuda/local_cuda_cluster.py
+++ b/dask_cuda/local_cuda_cluster.py
@@ -126,4 +126,4 @@ class LocalCUDACluster(LocalCluster):
             }
         )
 
-        return name, spec
+        return {name: spec}


mrocklin:
The failure here is just because this change depends on the master version of distributed.  I'm not sure if we want to pin dask-cuda to that yet.

pentschev:
What would be the issue in pinning CI only to distributed master? We've done that in the past, unless there are some changes with distributed master that may break dask-cuda somehow, on which case we would need to fix it here first.

mrocklin:
I think that I should probably make distributed more flexible in supporting
the older version here for a while.

pentschev:
Alright, I've been working on issues in this repo today, and it seems like this fix is necessary, as well as updating to upstream distributed. This is the case for https://github.com/rapidsai/dask-cuda/pull/126 too.

@mrocklin the simplest is definitely just to make sure CI uses master, but over time this becomes an issue since we have to remember of removing that close to release dates to ensure the repo still works. Any ideas on how we could better treat issues with unreleased versions of dask/distributed?

mrocklin:
One approach would be to move the dask-cuda release cycle in line with Dask rather than RAPIDS.  For the deployment work I tend to update all of the deployment solutions at each release and co-release them to avoid conflicts like this.

This pain is particular to deployment I think, which is undergoing heavy churn right now.  I'm hoping that this pain goes away in the next few months.

pentschev:
I think in this particular case, the release cycle that dask-cuda is relying on doesn't make much difference. The issue here is that the latest distributed release breaks dask-cuda. The only real solution I see for this case (which has already happened before) is to pin to a specific version of dask and distributed, and then we should probably have dask-cuda updated to newer releases manually. Would you think this is a reasonable alternative?

Regardless of the release cycle discussion, I think we should for now make sure CI uses distributed's upstream to merge this PR and #126, so that the work on this repo can resume. Would you like to do it here or prefer that I do it (#126 will still require changes from this PR, so I would suggest we fix and merge this one first).

mrocklin:
> The issue here is that the latest distributed release breaks dask-cuda.

Is it the latest release or is it that dask-cuda@master doesn't work with distributed@latest-release?

> The only real solution I see for this case (which has already happened before) is to pin to a specific version of dask and distributed, and then we should probably have dask-cuda updated to newer releases manually. 

For the other dask deployment projects (kubernetes, jobqueue) we tend to ensure that latest releases always support each other, and that master in each tends to support each other.  This seems to work out pretty well, and it doesn't require pinning environments, which folks will often avoid in Python, just because explicit pinning tends to create some havoc with complex environments.

> Would you think this is a reasonable alternative?

Sure, it's reasonable, especially during this period of deployment churn.  I'm somewhat in favor of co-releasing if that's feasible, but I'm happy to defer to your judgement here.  You've been doing most of the work here.

pentschev:
> Is it the latest release or is it that dask-cuda@master doesn't work with distributed@latest-release?

Both, because dask-cuda 0.9 and master (aka, 0.10) install distributed>=2.3.0 (which is currently 2.3.2), but they work with distributed==2.3.0.

> For the other dask deployment projects (kubernetes, jobqueue) we tend to ensure that latest releases always support each other, and that master in each tends to support each other. This seems to work out pretty well, and it doesn't require pinning environments, which folks will often avoid in Python, just because explicit pinning tends to create some havoc with complex environments.

Yeah, that's a great point that I didn't think about.

> Sure, it's reasonable, especially during this period of deployment churn. I'm somewhat in favor of co-releasing if that's feasible, but I'm happy to defer to your judgement here. You've been doing most of the work here.

Unfortunately, your prior point (creating havoc on complex environments) feels like this would be a bad choice, which I would prefer to avoid. I think for now we should just point CI to distributed@master and revert that once there's a new distributed release. Would you like to do it or prefer that I submit a PR to your branch?

mrocklin:
> Would you like to do it or prefer that I submit a PR to your branch?

As a maintainer you should have push access to my branch here.  Please feel free to push to any branch of mine to which you have access now or at any time in the future.

>  I think for now we should just point CI to distributed@master and revert that once there's a new distributed release.

Fine with me.  Once there is a distributed release (which will probably happen later this week) what happens with dask-cuda?  The latest releases of both projects will probably not be consistent.  Normally I would publish a release of dask-foo when this happens, but I'm not sure we can do that with the current RAPIDS release procedure.

pentschev:
> As a maintainer you should have push access to my branch here. Please feel free to push to any branch of mine to which you have access now or at any time in the future.

Will do.

> Fine with me. Once there is a distributed release (which will probably happen later this week) what happens with dask-cuda? The latest releases of both projects will probably not be consistent. Normally I would publish a release of dask-foo when this happens, but I'm not sure we can do that with the current RAPIDS release procedure.

We can then remove that install from CI and simply let the latest version to get installed (as should be the case since it now installs `distributed>=2.3.0`). I think this will be automatically solved when the latest release gets installed, unless users pin to a specific version (such as `distributed=2.3.2`), in which case they're out of luck.

pentschev:
Ok, this passes now, I'm merging it. Thanks for the fix @mrocklin !

PR #131, opened 2019-09-11
Update TOTAL_MEMORY to match new distributed MEMORY_LIMIT

pentschev:
Fixes dask-cuda post https://github.com/dask/distributed/pull/3039

### DIFF
--- a/dask_cuda/dgx.py
+++ b/dask_cuda/dgx.py
@@ -1,7 +1,7 @@
 import os
 
 from dask.distributed import Nanny, SpecCluster, Scheduler
-from distributed.worker import TOTAL_MEMORY
+from distributed.system import MEMORY_LIMIT
 
 from .local_cuda_cluster import cuda_visible_devices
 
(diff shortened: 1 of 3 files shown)


dask_cuda/dgx.py
```diff
@@ -1,7 +1,7 @@
 import os
 
 from dask.distributed import Nanny, SpecCluster, Scheduler
-from distributed.worker import TOTAL_MEMORY
```
mrocklin: @jcrist should we provide some compatibility buffer for this?

mrocklin: "No" is a fine answer from my perspective, but I thought I'd mention it

jcrist: I'd say no, this should never have been publicly used.

pentschev: I guess "publicly used" doesn't apply to this repo, since it extends dask/distributed, right? I'm not sure we have an alternative to doing this at the moment, but happy to change it to something safer as well if there is one.

jcrist: No, what you've moved it to should be fine. `TOTAL_MEMORY` in `distributed.worker` was more of an implementation detail, I'd say `distributed.system` should be public enough for dask-affiliated projects.

pentschev: Great, thanks for the feedback @jcrist!

pentschev:
@mrocklin if there are no other concerns, and given that https://github.com/rapidsai/dask-cuda/pull/131#discussion_r323342381 has been answered, I think this PR is good to merge at your earliest convenience. :)

mrocklin:
Thanks @pentschev !

PR #134, opened 2019-09-18
Use YYMMDD tag in nightly build

mluukkainen:


pentschev:
Generally, we use the latest commit's hash to identify the build, as most Python projects do.

@raydouglass @dillon-cullinan is this a change that you see would be useful/acceptable for RAPIDS packages?

pentschev:
I see now, wasn't aware of https://github.com/rapidsai/ops/issues/452. Looks good to me, thanks for the PR @mluukkainen, and @rlratzel for the review. Merging.

PR #138, opened 2019-09-24
Automatically determine CPU affinity

pentschev:
This PR allows automatically determining and setting CPU affinity for all GPUs during initialization.

cc @benjha as this is one of the issues we discussed on our call last week.

### DIFF
--- a/dask_cuda/tests/test_utils.py
+++ b/dask_cuda/tests/test_utils.py
@@ -2,7 +2,7 @@ import pytest
 
 import os
 
-from dask_cuda.utils import bitmask_to_list, get_cpu_affinity_list, get_n_gpus
+from dask_cuda.utils import get_cpu_affinity_list, get_n_gpus, unpack_bitmask
 
 
 def test_get_n_gpus():
@@ -30,8 +30,8 @@ def test_get_n_gpus():
         {"input": [0, 18446744073709551615], "output": [i + 64 for i in range(64)]},
     ],
 )
-def test_bitmask_to_list(params):
-    assert bitmask_to_list(params["input"]) == params["output"]
+def test_unpack_bitmask(params):
+    assert unpack_bitmask(params["input"]) == params["output"]
 
 
 def test_cpu_affinity():
--- a/dask_cuda/tests/test_utils.py
+++ b/dask_cuda/tests/test_utils.py
@@ -2,7 +2,7 @@ import pytest
 
 import os
 
-from dask_cuda.utils import get_cpu_affinity_list, get_n_gpus, unpack_bitmask
+from dask_cuda.utils import get_cpu_affinity, get_n_gpus, unpack_bitmask
 
 
 def test_get_n_gpus():
@@ -36,6 +36,6 @@ def test_unpack_bitmask(params):
 
 def test_cpu_affinity():
     for i in range(get_n_gpus()):
-        affinity = get_cpu_affinity_list(i)
+        affinity = get_cpu_affinity(i)
         os.sched_setaffinity(0, affinity)
         assert list(os.sched_getaffinity(0)) == affinity
--- a/dask_cuda/tests/test_utils.py
+++ b/dask_cuda/tests/test_utils.py
@@ -26,19 +26,12 @@ def test_get_n_gpus():
             "input": [17293823668613283840, 65535],
             "output": [i + 20 for i in range(20)] + [i + 60 for i in range(20)],
         },
-        {
-            "input": [18446744073709551615, 0],
-            "output": [i for i in range(64)],
-        },
-        {
-            "input": [0, 18446744073709551615],
-            "output": [i + 64 for i in range(64)],
-        },
+        {"input": [18446744073709551615, 0], "output": [i for i in range(64)]},
+        {"input": [0, 18446744073709551615], "output": [i + 64 for i in range(64)]},
     ],
 )
 def test_bitmask_to_list(params):
     assert bitmask_to_list(params["input"]) == params["output"]
-
 
 
 def test_cpu_affinity():
--- a/dask_cuda/tests/test_utils.py
+++ b/dask_cuda/tests/test_utils.py
@@ -1,6 +1,8 @@
+import pytest
+
 import os
 
-from dask_cuda.utils import get_n_gpus
+from dask_cuda.utils import bitmask_to_list, get_cpu_affinity_list, get_n_gpus
 
 
 def test_get_n_gpus():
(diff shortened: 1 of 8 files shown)


dask_cuda/tests/test_utils.py
```diff
@@ -11,3 +13,29 @@ def test_get_n_gpus():
...
+)
+def test_bitmask_to_list(params):
+    assert bitmask_to_list(params["input"]) == params["output"]
+
+
+def test_cpu_affinity():
+    for i in range(get_n_gpus()):
+        affinity = get_cpu_affinity_list(i)
+        os.sched_setaffinity(0, affinity)
+        assert list(os.sched_getaffinity(0)) == affinity
```
mrocklin: This seems to test the `os.sched_setaffinity`/`os.sched_getaffinity` functionality more than it tests `get_cpu_affinity_list`?

dask_cuda/utils.py
```diff
@@ -1,6 +1,57 @@
...
+
+
+class CPUAffinity:
+    def __init__(self, cores):
+        self.cores = cores
+
+    def setup(self, worker=None):
+        print("before", os.sched_getaffinity(0))
+        os.sched_setaffinity(0, self.cores)
+        print("after ", os.sched_getaffinity(0))
```
mrocklin: We might want to remove these print statements

dask_cuda/utils.py
```diff
@@ -1,6 +1,57 @@
...
+
+
+def bitmask_to_list(x, mask_bits=64):
```
mrocklin: Can I ask for a small docstring here, perhaps with a small example?

dask_cuda/utils.py
```diff
@@ -1,6 +1,57 @@
...
+    return cpu_count()
+
+
+def get_cpu_affinity_list(device_index):
+    nvmlInit()
+
+    # Result is a list of 64-bit integers, thus ceil(CPU_COUNT / 64)
+    affinity = nvmlDeviceGetCpuAffinity(
+        nvmlDeviceGetHandleByIndex(device_index), math.ceil(get_cpu_count() / 64)
+    )
```
mrocklin: I'm glad to see that this function was available in nvml!

dask_cuda/utils.py
```diff
@@ -1,6 +1,57 @@
...
+
+        bytestr = np.frombuffer(
+            bytes(np.binary_repr(mask, width=mask_bits), "utf-8"), "u1"
+        )
+        mask = np.flip(bytestr - ord("0")).astype(np.bool)
+        affinity = np.where(
+            mask, np.arange(mask_bits) + cpu_offset, np.full(mask_bits, -1)
+        )
+
+        res += affinity[(affinity >= 0)].tolist()
```
mrocklin: I like that this function is general purpose.  My guess is that terms like `affinity` here are probably leaking in from how you are using it in this case.  Maybe these should have different names?

dask_cuda/utils.py
```diff
@@ -1,6 +1,57 @@
...
+
+    return res
+
+
+@toolz.memoize
+def get_cpu_count():
+    return cpu_count()
+
+
+def get_cpu_affinity_list(device_index):
```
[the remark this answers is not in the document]
mrocklin: I think that adding a small docstring with a small example would also be good here.  I find that docstring examples are some of the nicest ways to communicate functionality.  I personally find them valuable when reviewing, or trying to figure out what internal code does later when debugging.

https://docs.dask.org/en/latest/develop.html#docstrings

dask_cuda/tests/test_utils.py
pentschev: Well, both. I'm actually verifying that what I'm setting is indeed workable. If I pass dummy values the result will be wrong and we can't predict the affinity of the machine running the test, so it's probably the best to check what we expected to set got set.

dask_cuda/utils.py
pentschev: Renamed to `unpack_bitmask` and added docstring. I think the new name is more descriptive.

dask_cuda/utils.py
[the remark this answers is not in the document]
pentschev: Done.

dask_cuda/utils.py
pentschev: Yes, it was pretty straightforward.

dask_cuda/utils.py
pentschev: Done, renamed it to `unpacked_mask`.

pentschev:
I think all the issues have been addressed, but I think tests will still fail. I'm trying to reproduce the issue, but I can't locally. Will keep trying.

dask_cuda/tests/test_utils.py
mrocklin: Ah, good point

mrocklin:
@rjzamora have you seen errors like this?

```python-traceback
/conda/envs/gdf/lib/python3.6/site-packages/tornado/ioloop.py:532: in run_sync

    return future_cell[0].result()

dask_cuda/tests/test_local_cuda_cluster.py:48: in test_with_subset_of_cuda_visible_devices

    scheduler_port=0, asynchronous=True, device_memory_limit=1

dask_cuda/local_cuda_cluster.py:111: in __init__

    self.scale(n_workers)

/conda/envs/gdf/lib/python3.6/site-packages/distributed/deploy/spec.py:408: in scale

    self.worker_spec.update(self.new_worker_spec())

dask_cuda/local_cuda_cluster.py:131: in new_worker_spec

    "plugins": {CPUAffinity(get_cpu_affinity(ii))},

dask_cuda/utils.py:88: in get_cpu_affinity

    nvmlDeviceGetHandleByIndex(device_index), math.ceil(get_cpu_count() / 64)

/conda/envs/gdf/lib/python3.6/site-packages/pynvml/nvml.py:922: in nvmlDeviceGetHandleByIndex

    check_return(ret)

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 



ret = 2



    def check_return(ret):

        if (ret != NVML_SUCCESS):

>           raise NVMLError(ret)

E           pynvml.nvml.NVMLError_InvalidArgument: Invalid Argument

```

pentschev:
I understand the error now. It means the device with the index `ii` doesn't exist. This is because on CI there's only one GPU available, but we still treat it as if there were several. I'll probably need to rework that test.

pentschev:
The issue is fixed, but there was an error on downloading packages, I'm rerunning the tests just in case.

pentschev:
rerun tests

pentschev:
All good for a new review now @mrocklin.

mrocklin:
This looks great to me.  Thanks @pentschev !

Issue #136, opened 2019-09-22
[BUG] worker memory limit sum test fails on a 6 GPU machine


ksangeek:
**Describe the bug**
Executing dask-cuda 0.9.1 `test_local_cuda_cluster()` test on a 6 GPU IBM AC922 (linux_ppc64le) machine fails -
```
>               assert sum(w.memory_limit for w in cluster.workers.values()) == TOTAL_MEMORY
E               assert 544713998334 == 544713998336
E                +  where 544713998334 = sum(<generator object test_local_cuda_cluster.<locals>.<genexpr> at 0x7fffaf1f2728>)
```

Please let me know if I any more information is required to help debug this issue. Thanks!

**Steps/Code to reproduce bug**
1. Install the dask-cuda `0.9.1` conda package which we have built for linux_ppc64le
2. Clone the `v0.9.1` code of https://github.com/rapidsai/dask-cuda.git
3. cd `dask_cuda/tests`
4. Execute `pytest test_local_cuda_cluster.py`

**Pls note** - The test only fails on a 6 GPU machine, but succeeds on a 4 GPU machine.

**Expected behavior**
The test case passes in a 4 GPU machine. It should also succeed in a 6 GPU machine!

**Environment overview (please complete the following information)**
 - Environment location: Bare-metal
 - Method of dask-cuda install: conda [Built for linux_ppc64le]

**Environment details**
The machine where it fails is an IBM AC922 machine with 6 16GB NVIDIA GPUs.

**Additional context**

```python
$ pytest test_local_cuda_cluster.py::test_local_cuda_cluster
===================================================================================== test session starts =====================================================================================
platform linux -- Python 3.6.9, pytest-5.0.1, py-1.8.0, pluggy-0.12.0
rootdir: /blah-blah/sandbox/dask-cuda-0.9.1
collected 1 item

test_local_cuda_cluster.py F                                                                                                                                                            [100%]

========================================================================================== FAILURES ===========================================================================================
___________________________________________________________________________________ test_local_cuda_cluster ___________________________________________________________________________________

    def test_func():
        with clean() as loop:
            if iscoroutinefunction(func):
                cor = func
            else:
                cor = gen.coroutine(func)
>           loop.run_sync(cor, timeout=timeout)

../../../../aconda3/envs/dask-cuda-py/lib/python3.6/site-packages/distributed/utils_test.py:761:
[...]

PR #139, opened 2019-09-24
FIX full memory use check testcase

ksangeek:
This PR is for https://github.com/rapidsai/dask-cuda/issues/136

### DIFF
--- a/dask_cuda/tests/test_local_cuda_cluster.py
+++ b/dask_cuda/tests/test_local_cuda_cluster.py
@@ -29,8 +29,10 @@ async def test_local_cuda_cluster():
                     range(utils.get_n_gpus())
                 )
 
-            # Use full memory
-            assert sum(w.memory_limit for w in cluster.workers.values()) == MEMORY_LIMIT
+            # Check workers use equal portion of full memory, ignoring FP rounding difference
+            per_worker_memory = round(MEMORY_LIMIT/utils.get_n_gpus())
+            for i in range(utils.get_n_gpus()):
+                assert cluster.workers[i].memory_limit == per_worker_memory
 
             for w, devices in result.items():
                 ident = devices[0]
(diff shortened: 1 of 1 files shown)


GPUtester:

GPUtester:

dask_cuda/tests/test_local_cuda_cluster.py
```diff
@@ -29,8 +29,10 @@ def get_visible_devices():
                     range(utils.get_n_gpus())
                 )
 
-            # Use full memory
-            assert sum(w.memory_limit for w in cluster.workers.values()) == MEMORY_LIMIT
+            # Check workers use equal portion of full memory, ignoring FP rounding difference
+            per_worker_memory = round(MEMORY_LIMIT/utils.get_n_gpus())
+            for i in range(utils.get_n_gpus()):
+                assert cluster.workers[i].memory_limit == per_worker_memory
```
pentschev: I think this approach may still cause a similar problem in some cases. I would suggest testing for some limits instead:
```python
full_mem = sum(w.memory_limit for w in cluster.workers.values())
assert full_mem >= MEMORY_LIMIT-1024 and full_mem < MEMORY_LIMIT-1024
```

mrocklin: (I think that @pentschev meant a `+` rather than a `-` in that last case.)

pentschev: Indeed, the upper limit should have a `+`, I fixed that now.

ksangeek: Thank you @pentschev for suggesting an alternative approach.
The earlier approach had worked fine, and the new one also works. I updated the PR as per your suggestion. Thanks!

pentschev: Yes, the previous approach was probably good for most cases too, but having some more relaxed condition is usually more robust.

pentschev:
ok to test

pentschev:
Thanks @ksangeek for your contribution! Merging.

PR #141, opened 2019-09-27
Pass missing local_directory to Nanny from dask-cuda-worker

pentschev:


mrocklin:
Thanks for finding and fixing this @pentschev .  This is the sort of thing that makes it more attractive to find a way to push more of this functionality upstream in the future.  Maintaining a copy of `dask-worker` like this is frustrating.

pentschev:
@mrocklin agreed, but we need CI upstream then.

Issue #146, opened 2019-10-01
Appropriate way to use spec-cluster with dask-cuda


pentschev:
Currently dask-cuda has no class similar to distributed's `Worker`, given it currently doesn't need a specialized worker. However, we have a CLI that extends `dask-worker`, but it can't be directly used with spec-cluster as we may need additional parameters to pass to each process of `dask-cuda-worker`.

Recently, a new `dask-spec-worker` was proposed https://github.com/dask/distributed/pull/3090 which seems like a good way for enabling additional functionality to `dask-cuda-worker`, but as far as I understand it, we cannot directly pass the CLI interface with that, we would require some `CUDAWorker` class for that.

@mrocklin any suggestions on what would be an appropriate solution here?

mrocklin:
> Currently dask-cuda has no class similar to distributed's Worker, given it currently doesn't need a specialized worker

Yes, this is, I think, by design.  Ideally we don't need to create a specialized `Worker` class.  I think that this is best in the long run for maintenance reasons.  The fact that we had to have a `dask-cuda-worker` is, I hope, a short term workaround.

We use the current `Nanny` class with `SpecCluster` today in `DGX`.  I think that it works well.

So, to the questions asked in this issue my initial gut reaction is "we're already done".  However I suspect that there is some deeper issue that you're trying to resolve.  Can I ask what caused you to raise this issue?

I think that there are a few different questions here:

pentschev:
Additionally, we don't want to force users to specify the options that are already available with `dask-cuda-worker`, even though it's technically possible that the user mimics it just with a regular `Worker` instance (or `Nanny`, perhaps).

mrocklin:
> Is that your main proposal?

I don't really have a proposal.  I'm not sure what the best course of action here is yet.  Thoughts are welcome.

mrocklin:
I think that having "please do everything right automatically for GPUs" user options makes sense.  I would like to reduce the amount of forks in the codebase if possible, just to reduce friction when we update things upstream.  Designing things well to achieve both of these goals simultaneously is hard :/

mrocklin:
I think that we're building out some functionality that will be useful in the final product, like `SpecCluster`, your automatic determination of CPU affinity, GPUs, and so on.  I don't know the right way to assemble everything though.

pentschev:
[...]

PR #147, opened 2019-10-01
New worker_spec function for worker recipes

pentschev:
Add a function for creation "worker recipes" that can be used with `SpecCluster`, following discussion in https://github.com/rapidsai/dask-cuda/issues/146.

### DIFF
--- a/dask_cuda/dgx.py
+++ b/dask_cuda/dgx.py
@@ -1,18 +1,18 @@
 import os
 
-from dask.distributed import Nanny, SpecCluster, Scheduler
-from distributed.system import MEMORY_LIMIT
+from dask.distributed import SpecCluster, Scheduler
 
-from .local_cuda_cluster import cuda_visible_devices
-from .utils import CPUAffinity, get_cpu_affinity
+from .worker_spec import worker_spec
 
 
 def DGX(
-    interface="ib",
+    interface="enp1s0f0",
     dashboard_address=":8787",
     threads_per_worker=1,
     silence_logs=True,
     CUDA_VISIBLE_DEVICES=None,
+    enable_infiniband=True,
+    enable_nvlink=True,
     **kwargs
 ):
     """ A Local Cluster for a DGX 1 machine
@@ -32,14 +32,20 @@ def DGX(
     Parameters
     ----------
     interface: str
-        The interface prefix for the infiniband networking cards.  This is
-        often "ib"` or "bond".  We will add the numeric suffix 0,1,2,3 as
-        appropriate.  Defaults to "ib".
+        The external interface used to connect to the scheduler.
     dashboard_address: str
         The address for the scheduler dashboard.  Defaults to ":8787".
+    threads_per_worker: int
+        Number of threads to be used for each CUDA worker process.
+    silence_logs: bool
+        Disable logging for all worker processes
     CUDA_VISIBLE_DEVICES: str
         String like ``"0,1,2,3"`` or ``[0, 1, 2, 3]`` to restrict activity to
         different GPUs
+    enable_infiniband: bool
+        Set environment variables to enable UCX InfiniBand support
+    enable_nvlink: bool
+        Set environment variables to enable UCX NVLink support
 
     Examples
     --------
@@ -48,42 +54,21 @@ def DGX(
     >>> cluster = DGX(interface='ib')
     >>> client = Client(cluster)
     """
-    if CUDA_VISIBLE_DEVICES is None:
-        CUDA_VISIBLE_DEVICES = os.environ.get("CUDA_VISIBLE_DEVICES", "0,1,2,3,4,5,6,7")
-    if isinstance(CUDA_VISIBLE_DEVICES, str):
-        CUDA_VISIBLE_DEVICES = CUDA_VISIBLE_DEVICES.split(",")
-    CUDA_VISIBLE_DEVICES = list(map(int, CUDA_VISIBLE_DEVICES))
-    memory_limit = MEMORY_LIMIT / 8
-
-    spec = {
-        i: {
-            "cls": Nanny,
-            "options": {
-                "env": {
-                    "CUDA_VISIBLE_DEVICES": cuda_visible_devices(
-                        ii, CUDA_VISIBLE_DEVICES
-                    ),
-                    # 'UCX_NET_DEVICES': 'mlx5_%d:1' % (i // 2)
-                    "UCX_TLS": "rc,cuda_copy,cuda_ipc",
-                },
-                "interface": interface + str(i // 2),
-                "protocol": "ucx",
-                "nthreads": threads_per_worker,
-                "data": dict,
-                "preload": ["dask_cuda.initialize_context"],
-                "dashboard_address": ":0",
-                "plugins": [CPUAffinity(get_cpu_affinity(i))],
-                "silence_logs": silence_logs,
-                "memory_limit": memory_limit,
-            },
-        }
-        for ii, i in enumerate(CUDA_VISIBLE_DEVICES)
-    }
+    spec = worker_spec(
+        interface=interface,
+        dashboard_address=dashboard_address,
+        threads_per_worker=threads_per_worker,
+        silence_logs=silence_logs,
+        CUDA_VISIBLE_DEVICES=CUDA_VISIBLE_DEVICES,
+        enable_infiniband=enable_infiniband,
+        enable_nvlink=enable_nvlink,
+        **kwargs,
+    )
 
     scheduler = {
         "cls": Scheduler,
         "options": {
-            "interface": interface + str(CUDA_VISIBLE_DEVICES[0] // 2),
+            "interface": interface,
             "protocol": "ucx",
             "dashboard_address": dashboard_address,
         },
--- a/dask_cuda/dgx.py
+++ b/dask_cuda/dgx.py
@@ -61,6 +61,7 @@ def DGX(
         silence_logs=silence_logs,
         CUDA_VISIBLE_DEVICES=CUDA_VISIBLE_DEVICES,
         enable_infiniband=enable_infiniband,
+        ucx_net_device=lambda i: "mlx5_%d:1" % (i // 2),
         enable_nvlink=enable_nvlink,
         **kwargs,
     )
--- a/dask_cuda/dgx.py
+++ b/dask_cuda/dgx.py
@@ -6,7 +6,7 @@ from .worker_spec import worker_spec
 
 
 def DGX(
-    interface="enp1s0f0",
+    interface=None,
     dashboard_address=":8787",
     threads_per_worker=1,
     silence_logs=True,
@@ -32,7 +32,8 @@ def DGX(
     Parameters
     ----------
     interface: str
-        The external interface used to connect to the scheduler.
+        The external interface used to connect to the scheduler, usually
+        the ethernet interface is used for connection (not the InfiniBand!).
     dashboard_address: str
         The address for the scheduler dashboard.  Defaults to ":8787".
     threads_per_worker: int
--- a/dask_cuda/tests/test_worker_spec.py
+++ b/dask_cuda/tests/test_worker_spec.py
@@ -1,0 +1,59 @@
+import pytest
+
+from distributed import Nanny
+
+from dask_cuda.worker_spec import worker_spec
+
+
+def _check_option(spec, k, v):
+    assert all([s["options"][k] == v for s in spec.values()])
+
+def _check_env_key(spec, k, enable):
+    if enable:
+        assert all([k in s["options"]["env"] for s in spec.values()])
+    else:
+        assert all([k not in s["options"]["env"] for s in spec.values()])
+
+def _check_env_value(spec, k, v):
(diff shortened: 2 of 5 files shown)


pentschev:
@mrocklin this is more or less the idea I suggested in #146, please let me know what you think.

This is still unfinished though, given that we need to pass some environment variables to the scheduler as well. Do you have an idea of the proper way to do it? Unlike the `Workers` forked by `Nanny`, the `Scheduler` seems to be forked by `SpecCluster` itself, so the `SpecCluster` would have to know about the environment variables we need to pass.

dask_cuda/dgx.py
```diff
@@ -1,18 +1,18 @@
...
+from dask.distributed import SpecCluster, Scheduler
 
-from .local_cuda_cluster import cuda_visible_devices
-from .utils import CPUAffinity, get_cpu_affinity
+from .worker_spec import worker_spec
 
 
 def DGX(
-    interface="ib",
+    interface="enp1s0f0",
```
mrocklin: I think that Python usually finds the ethernet interface by default.  I wonder if maybe now we can drop this keyword altogether.

dask_cuda/worker_spec.py
```diff
@@ -0,0 +1,116 @@
...
+       'plugins': [<dask_cuda.utils.CPUAffinity at 0x7f2d47f9fc50>],
+       'silence_logs': True,
+       'memory_limit': 135263611392.0}}}
+    """
+    if CUDA_VISIBLE_DEVICES is None:
+        CUDA_VISIBLE_DEVICES = os.environ.get("CUDA_VISIBLE_DEVICES", "0,1,2,3,4,5,6,7")
+    if isinstance(CUDA_VISIBLE_DEVICES, str):
+        CUDA_VISIBLE_DEVICES = CUDA_VISIBLE_DEVICES.split(",")
+    CUDA_VISIBLE_DEVICES = list(map(int, CUDA_VISIBLE_DEVICES))
+    memory_limit = MEMORY_LIMIT / 8
```
mrocklin: The focus here on 8 GPUs seems DGX specific.  I wonder if we might detect this automatically somehow.

dask_cuda/worker_spec.py
```diff
@@ -0,0 +1,116 @@
...
+
+    ucx_env = {}
+    if enable_infiniband:
+        ucx_env["UCX_SOCKADDR_TLS_PRIORITY"] = "sockcm"
+        ucx_env["UCX_TLS"] = "rc,tcp,sockcm"
+        ucx_env["UCXPY_IFNAME"] = interface
```
mrocklin: I think that Python dicts have a `setdefault` method that might be appropriate here in case we don't want to overwrite these values if they're already present.

```python
In [1]: d = {}

In [2]: d = {"x": 1}

In [3]: d.setdefault("y", 2)
Out[3]: 2

In [4]: d
Out[4]: {'x': 1, 'y': 2}

In [5]: d.setdefault("x", 3)  # respect original value
Out[5]: 1

In [6]: d
Out[6]: {'x': 1, 'y': 2}
```

dask_cuda/worker_spec.py
```diff
@@ -0,0 +1,116 @@
...
+
+    spec = {
+        i: {
+            "cls": Nanny,
+            "options": {
+                "env": {
+                    "CUDA_VISIBLE_DEVICES": cuda_visible_devices(
+                        ii, CUDA_VISIBLE_DEVICES
+                    ),
+                    "UCX_NET_DEVICES": "mlx5_%d:1" % (i // 2),
```
mrocklin: This is also DGX specific, right?

mrocklin:
> This is still unfinished though, given that we need to pass some environment variables to the scheduler as well. Do you have an idea of the proper way to do it? Unlike the Workers forked by Nanny, the Scheduler seems to be forked by SpecCluster itself, so the SpecCluster would have to know about the environment variables we need to pass.

Currently we are launching the scheduler from this process, so we have no opportunity to set environment variables.  To do what you want we would need a Nanny process for the Scheduler.  We don't have that currently, but we could make one.

pentschev:
> I'm curious if your intent for the `worker_spec` function is to be specific to DGX's or not.

It's not going to be specific to DGX, even though it is right now. My idea is that we can expand and make it more generic, but this was just to give us an initial view. Ideally, we would have support for things like device memory spilling and better ways to specify the InfiniBand interfaces (and perhaps do that automatically).

dask_cuda/dgx.py
pentschev: I think we can make it `None` by default, but some machines may have multiple ethernet interfaces available, so it's generally good to be able to specify that.

pentschev:
> Currently we are launching the scheduler from this process, so we have no opportunity to set environment variables. To do what you want we would need a Nanny process for the Scheduler. We don't have that currently, but we could make one.

Do you think we have a short-term solution/workaround for this? It's important that people have a way to easily start up a dask cluster to test UCX properly. We have limitations on passing environment variables on `SSHCluster`, `dask-spec-worker` and here. I'm honestly not sure how to best handle this, but I need some other people to be able to debug issues as well.

dask_cuda/dgx.py
pentschev: Addressed in the latest commit.

dask_cuda/worker_spec.py
pentschev: Addressed in the latest commit.

dask_cuda/worker_spec.py
pentschev: Addressed in the latest commit.

dask_cuda/worker_spec.py
pentschev: I'm not sure where you're specifically suggesting that `setdefault` is used. Note that we're not really overwriting anything since `ucx_env` is a `dict` created inside of `worker_spec`.

mrocklin: Oh I see, right.  I guess we would want to check `os.environ`.

I think broadly my question is "should we respect user-provided environment variables if they are present"

mrocklin:
> Do you think we have a short-term solution/workaround for this?

It is not very hard to create a Scheduler-like class that creates a new process and passes in all keyword arguments to that new process.  

We need to implement this interface

https://github.com/dask/distributed/blob/ad37905521c8c133707b44a4c7676ee4735e20f0/distributed/deploy/spec.py#L22

And make a new process, probably with the API in `distributed.process` (although it's also possibly needlessly complicated).  We could also try to `multiprocessing.Process` API, which is simpler, but possibly also error prone in some exotic ways.

I don't know of a shorter route.

pentschev:
> > Do you think we have a short-term solution/workaround for this?
> 
> It is not very hard to create a Scheduler-like class that creates a new process and passes in all keyword arguments to that new process.
> 
> We need to implement this interface
> 
> https://github.com/dask/distributed/blob/ad37905521c8c133707b44a4c7676ee4735e20f0/distributed/deploy/spec.py#L22
> 
> And make a new process, probably with the API in `distributed.process` (although it's also possibly needlessly complicated). We could also try to `multiprocessing.Process` API, which is simpler, but possibly also error prone in some exotic ways.
> 
> I don't know of a shorter route.

Yes, I think this seems like a good and quick solution. That's mainly what I needed, some guidance on what existing tools to use. Thanks @mrocklin, I will work on it!

dask_cuda/worker_spec.py
pentschev: Probably makes sense in a future iteration, once we are certain of the variables we would need in various cases and we can sync worker/scheduler in the same manner. For now I think this is more verbose and thus safer to understand the behavior.

dask_cuda/dgx.py
mrocklin: Did we want this to be `None`?

dask_cuda/tests/test_worker_spec.py
```diff
@@ -0,0 +1,59 @@
...
+
+@pytest.mark.parametrize("num_devices", [1, 4])
+@pytest.mark.parametrize("cls", [Nanny])
+@pytest.mark.parametrize("interface", [None, "eth0", "enp1s0f0"])
+@pytest.mark.parametrize("protocol", [None, "tcp", "ucx"])
+@pytest.mark.parametrize("dashboard_address", [None, ":0", ":8787"])
+@pytest.mark.parametrize("threads_per_worker", [1, 8])
+@pytest.mark.parametrize("silence_logs", [False, True])
+@pytest.mark.parametrize("enable_infiniband", [False, True])
+@pytest.mark.parametrize("enable_nvlink", [False, True])
```
mrocklin: This is around 1000 tests total.  Maybe this is too much?

dask_cuda/tests/test_worker_spec.py
```diff
@@ -0,0 +1,59 @@
...
+            enable_infiniband=enable_infiniband, enable_nvlink=enable_nvlink)
+
+    assert len(spec) == num_devices
+    assert all([s["cls"] == cls for s in spec.values()])
+
+    _check_option(spec, "interface", interface)
+    _check_option(spec, "protocol", protocol)
+    _check_option(spec, "dashboard_address", dashboard_address)
+    _check_option(spec, "nthreads", threads_per_worker)
+    _check_option(spec, "silence_logs", silence_logs)
```
mrocklin: I would prefer inlining these.  I know it makes it less pleasant to write and it's a bit repetitive, but it makes it easier for others to jump into this in the future.

dask_cuda/worker_spec.py
```diff
@@ -0,0 +1,128 @@
...
+    enable_nvlink: bool
+        Set environment variables to enable UCX NVLink support
+
+    Examples
+    --------
+    >>> from dask_cuda.worker_spec import worker_spec
+    >>> worker_spec(interface="enp1s0f0", CUDA_VISIBLE_DEVICES=[0, 1])
+    {0: {'cls': distributed.nanny.Nanny,
+      'options': {'env': {'CUDA_VISIBLE_DEVICES': '0,1',
+        'UCX_NET_DEVICES': 'mlx5_0:1'},
```
mrocklin: This output may not be correct?  It looks like `UCX_NET_DEVICES` may have been carried over from a previous example.

dask_cuda/worker_spec.py
```diff
@@ -0,0 +1,128 @@
...
+        Disable logging for all worker processes
+    CUDA_VISIBLE_DEVICES: str
+        String like ``"0,1,2,3"`` or ``[0, 1, 2, 3]`` to restrict activity to
+        different GPUs
+    enable_infiniband: bool
+        Set environment variables to enable UCX InfiniBand support
+    ucx_net_device: str or callable
+        A string with the interface name to be used for all devices (empty
+        string means use default), or a callable function taking an integer
+        identifying the GPU index.
```
mrocklin: This seems very specific.   I wonder if there is some more general solution, like a function that produces an environment or something that might be more flexible to future use cases.

pentschev:
> Did we want this to be None?

I'm not sure if we want `interface` to be `None` by default, I think this is more verbose now and makes easier for people to understand they should select the ethernet interface instead of the infiniband.

dask_cuda/tests/test_worker_spec.py
pentschev: You're right in number, but it runs in 3 seconds, so maybe it's ok for now?

mrocklin:
> I'm not sure if we want interface to be None by default, I think this is more verbose now and makes easier for people to understand they should select the ethernet interface instead of the infiniband.

Most people won't read the documentation string.  They'll just copy-paste some example text:

```python
from dask_cuda import DGX

cluster = DGX()
```

If their machine has a different ethernet interface name then things will break for them.

dask_cuda/tests/test_worker_spec.py
pentschev: Do you mind if we leave it for next release? I'd rather not waste time on things like this at the moment.

dask_cuda/worker_spec.py
pentschev: I'll check this.

dask_cuda/worker_spec.py
pentschev: We need to find something more generic, but I don't really have the time to do it now. So I think this is an ok-ish solution until we can come up with something that works better.

mrocklin:
> Do you mind if we leave it for next release? 

You've been handling most of the maintenance for this repository, so I'm happy to defer to your judgement.

dask_cuda/worker_spec.py
pentschev: Fixed in latest commits.

pentschev:
> > I'm not sure if we want interface to be None by default, I think this is more verbose now and makes easier for people to understand they should select the ethernet interface instead of the infiniband.
> 
> Most people won't read the documentation string. They'll just copy-paste some example text:
> 
> ```python
> from dask_cuda import DGX
> 
> cluster = DGX()
> ```
> 
> If their machine has a different ethernet interface name then things will break for them.

Fixed in latest commits.

pentschev:
Thanks @mrocklin, merging.

PR #149, opened 2019-10-02
[WIP] Add new Scheduler class supporting environment variables

pentschev:
This new scheduler is needed to pass environment variables related to UCX.

### DIFF
--- a/dask_cuda/__init__.py
+++ b/dask_cuda/__init__.py
@@ -1,5 +1,6 @@
 from .local_cuda_cluster import LocalCUDACluster
 from .dgx import DGX
+from .scheduler import Scheduler
 
 from ._version import get_versions
 
--- a/dask_cuda/dask_cuda_worker.py
+++ b/dask_cuda/dask_cuda_worker.py
@@ -246,6 +246,7 @@ def main(
             env={"CUDA_VISIBLE_DEVICES": cuda_visible_devices(i)},
             plugins={CPUAffinity(get_cpu_affinity(i))},
             name=name if nprocs == 1 or not name else name + "-" + str(i),
+            local_directory=local_directory,
             data=(
                 DeviceHostFile,
                 {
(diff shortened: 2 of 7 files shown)


pentschev:
@mrocklin this is still not finished, I need to do some more testing still, but I know you'll have some critique to do, so better get started sooner.

Note that most of the logic was copied from https://github.com/dask/distributed/blob/master/distributed/nanny.py#L438, with adaptations where needed.

mrocklin:
This looks like it is a copy-paste-modify job from the `distribued.nanny.WorkerProcess` class.  In a sense that's great, because we already know that this code seems to work well in practice, even in a variety of exotic fail cases.  In another sense, this is bad, because no one who maintains the project today really understands this code.  It was mostly written by Antoine, who has since left the project.

My guess is that we should probably go with what we have here for now (although maybe with some testing), and in the future we might explore some other options:

1.  Maybe we can avoid duplicating this code by doing some amount of subclassing?
2.  Maybe we look into replacing all of this code, for workers to, with some other system, perhaps like [aiomultiprocessing](https://github.com/jreese/aiomultiprocess) ? (I just googled that, no idea how good it is).

I would love to replace the WorkerProcess and AsyncProcess code in distributed some day.  Maintaining it is a challenge.

Anyway, I'm +1

pentschev:
I agree with your comment @mrocklin. Given the release time constraint, I'm mainly trying to have something that works for now, but we should definitely ensure that it's maintainable long-term.

I also don't know what would be a good long-term solution, but I guess we should explore a simple and understandable way of forking processes. The current `Nanny` code seems solid, but at the same time feels like it maybe is overkill and probably there would be simpler ways to achieve the same.

After 0.10, I can try and take a look at this and have this class eventually moved to distributed as well.

pentschev:
I will still do some more work on this PR tomorrow with tests before it gets merged.

pentschev:
rerun tests

pentschev:
@mrocklin I added some new tests, I think it's good for another review.

pentschev:
@mrocklin I'm merging this to keep things going and preventing this from becoming a giant PR. I'd still appreciate if you could take a look at the new tests and see if there's something very important that we're potentially missing on those.

PR #152, opened 2019-10-04
Support for TCP over UCX

pentschev:


pentschev:
@mrocklin we have some DGX tests here that pass (but obviously, they don't run on CI). Would you like to review this PR? I'm preparing another one to move away from environment variables in favor of utilizing preload modules with UCX-Py's config interface. I would nevertheless prefer to have this merged before just to have an easy way of rolling back changes though.

dask_cuda/utils.py
```diff
@@ -129,45 +129,61 @@ def get_device_total_memory(index=0):
     ).total
 
 
-def get_ucx_env(enable_infiniband=False, interface=None, enable_nvlink=False):
+def get_ucx_env(
+    enable_tcp=True, enable_infiniband=False, interface=None, enable_nvlink=False
+):
```
mrocklin: I like that this is pulled off into a separate function.  This seems like the sort of thing that could be moved upstream easily.

(athough hopefully we can do this programmatically at some point)

dask_cuda/worker_spec.py
```diff
@@ -14,6 +14,7 @@ def worker_spec(
     threads_per_worker=1,
     silence_logs=True,
     CUDA_VISIBLE_DEVICES=None,
+    enable_tcp_over_ucx=False,
     enable_infiniband=False,
     ucx_net_devices="",
     enable_nvlink=False,
```
mrocklin: Why wouldn't all of the `enable_foo` options be `True` by default?

dask_cuda/utils.py
pentschev: Yup, hopefully most of the work here (and the upcoming PR) will be easily moveable.

dask_cuda/worker_spec.py
pentschev: Only because I don't know if they're all safe in any combinations. I guess we can fine tune this to a "fast and safe" approach when we are confident everything should work.

pentschev:
Thanks for the review @mrocklin, I'm gonna merge this now and submit another PR for UCX-Py config soon.

PR #154, opened 2019-10-09
Use UCX-Py initialization API

pentschev:
This PR replaces the explicit configuration of UCX via environment variables by UCX-Py's API for initialization.


pentschev:
cc @jacobtomlinson

mrocklin:
I'm probably being naive here, but thoughts about handling this with configuration instead?

```python
distributed:
  ucx:
    tls: rc,tcp,sockcm,cuda_copy,cuda_copy
    ...
```

We would then read these in `distributed.comm.ucx` and configure things there.  For things like IB or CUDA-Copy that don't make sense for machines without hardware, maybe we can detect that they dont use those and remove them, or maybe UCX is robust to adding mechanisms that don't apply (cc @Akshay-Venkatesh ?)

If we need to configure things per-worker like ucx-net-devices then maybe we make it easier for nannies to take in a subset of configuration and pass it down to workers.  This would be generally useful.

mrocklin:
In general preload scripts are kind of like the last resort of configuration.  They're powerful, but also a bit cumbersome.  I'm curious if we can find a way that feels more native.

mrocklin:
@pentschev I would also be very happy to chat about this some time if you're free.  Either now or earlier tomorrow.

pentschev:
> I'm probably being naive here, but thoughts about handling this with configuration instead?
> 
> ```python
> distributed:
>   ucx:
>     tls: rc,tcp,sockcm,cuda_copy,cuda_copy
>     ...
> ```
> 
> We would then read these in `distributed.comm.ucx` and configure things there. For things like IB or CUDA-Copy that don't make sense for machines without hardware, maybe we can detect that they dont use those and remove them, or maybe UCX is robust to adding mechanisms that don't apply (cc @Akshay-Venkatesh ?)

Long-term, I don't mind having them as configurations, but for now it's more difficult to have things not explicitly configured. I would also not like to pass UCX options explicitly, but rather something more descriptive like `enable_infiniband`, the UCX options are not even clear for us developers, it's going to be even more cluttered for users.

> If we need to configure things per-worker like ucx-net-devices then maybe we make it easier for nannies to take in a subset of configuration and pass it down to workers. This would be generally useful.

I think the more reasonable solution is trying to find the shortest-path IB interface automatically, but this is not done yet, as per https://github.com/rapidsai/ucx-py/issues/237. The configuration is "easy" today for us because we are generally testing on homogenous nodes, but finding proper subsets of configurations will get cumbersome as soon as you have two different types of nodes.

pentschev:
> In general preload scripts are kind of like the last resort of configuration. They're powerful, but also a bit cumbersome. I'm curious if we can find a way that feels more native.

Do you have something in mind? We already passed through environment variables and now instantiating UCX-Py with preload scripts. I have no other ideas at the moment.

> @pentschev I would also be very happy to chat about this some time if you're free. Either now or earlier tomorrow.

We can do that early tomorrow if it works for you.

mrocklin:
I agree that `enable_infiniband` is much more interpretable.  My concern is that now we own a new configuration system, and we're on the hook to explain it to people for the indefinite future.  Long term (I don't want to hold up progress here) I would like to apply pressure upstream on UCX to improve documentation of these options and then we can send users there.  We might also help them write this documentation.

pentschev:
> I agree that enable_infiniband is much more interpretable. My concern is that now we own a new configuration system, and we're on the hook to explain it to people for the indefinite future. Long term (I don't want to hold up progress here) I would like to apply pressure upstream on UCX to improve documentation of these options and then we can send users there. We might also help them write this documentation.

In that case, we should either have UCX automagically squeeze all it can from whatever is available on the system or have these high-level configuration presets there (such as `enable_infiniband`).

jakirkham:
cc @madsbk (in case you have thoughts on the ucx-py API usage here)

mrocklin:
> Do you have something in mind? We already passed through environment variables and now instantiating UCX-Py with preload scripts. I have no other ideas at the moment.

My thought was configuration.  The DGX class (or anything similar) would construct config dicts to pass to the Nannies, which would instantiate them for the workers.  Then `distributed.comm.ucx` would check that configuration and use it when calling `ucp.init(...)`

> but for now it's more difficult to have things not explicitly configured.

We can set configuration explicitly.  Dask's configuration system isn't just yaml files.  We can do things like `dask.config.set(...)` in Python code to handle things.  We could make it easy to pass config snippets through nannies, and more.

mrocklin:
In case you haven't seen this: https://docs.dask.org/en/latest/configuration.html (though I suspect that you have)

mrocklin:
Yes.  To be clear, I'm not against the work here.  I think that it makes sense to push out something pragmatic soon.

However, we can also change upstream relatively easily, and releases are pretty frequent.  Managing configuration and changing the `ucp.init` call are both things that we could do today, and then we could release tomorrow, for example.

pentschev:
> Yes. To be clear, I'm not against the work here. I think that it makes sense to push out something pragmatic soon.
> 
> However, we can also change upstream relatively easily, and releases are pretty frequent. Managing configuration and changing the `ucp.init` call are both things that we could do today, and then we could release tomorrow, for example.

Yeah, I understand it, but to be fair I just wanted the most self-contained solution due to our tight schedule this time.

pentschev:
@mrocklin I think everything is ready here. Would you mind taking a last look? Codecov is complaining despite having tests because the ucp module isn't available, but there isn't much we can do about that for now.

PR #155, opened 2019-10-10
Update changelog for v0.10 [skip ci]

pentschev:


pentschev:
@mrocklin FYI

CHANGELOG.rst
```diff
@@ -1,9 +1,22 @@
...
+- Use pynvml to get memory info without creating CUDA context (#140) `Peter Andreas Entschev`_
+- Pass missing local_directory to Nanny from dask-cuda-worker (#141) `Peter Andreas Entschev`_
+- New worker_spec function for worker recipes (#147) `Peter Andreas Entschev`_
+- Add new Scheduler class supporting environment variables (#149) `Peter Andreas Entschev`_
+- Support for TCP over UCX (#152) `Peter Andreas Entschev`_
 
+.. _`Matthew Rocklin`: https://github.com/mrocklin
+.. _`Peter Andreas Entschev`: https://github.com/pentschev
 .. _`Markku Luukkainen`: https://github.com/mluukkainen
+.. _`Sangeeth Keeriyadath`: https://github.com/ksangeek
```
mrocklin: We don't need to repeat names for every section.  Typically I just include the links at the bottom.  See https://raw.githubusercontent.com/dask/distributed/master/docs/source/changelog.rst for an example

mrocklin: So, for example, you don't need to add your name or my name here because they are already defined below.

Although what you have here is also ok, if you prefer it

pentschev: Makes sense, let me change that.

pentschev:
@mrocklin can you approve this? We need an admin to merge it due to CI not sending the status back to the PR.

Issue #159, opened 2019-10-14
[BUG] Conda install pulls a previous version


VibhuJawa:
When i do below on a fresh `conda` , `python3.7` environment,  a old version ( `19-09-24`) of `dask-cuda` seems to be pulled. 

i.e below seems to be pulls  a version from `19-09-24` .
```
 conda install -c rapidsai-nightly dask-cuda  
```
Pulls:
``` 
rapidsai-nightly/label/cuda10.0/linux-64::dask-cuda-0.10.0a190924-py37_15
```

On pinning the exact version to the latest one,  I am getting a bunch of conflicts even though this is a fresh `python3.7` environment.

#### Work Around:

To make it pull the latest version i seem to need to provide a lot of channels like below: 

```
conda install -c rapidsai-nightly -c rapidsai -c nvidia -c numba -c conda-forge dask-cuda
```

jakirkham:
@raydouglass, do you have any thoughts here on why this is happening?

raydouglass:
There is nothing wrong here. The command in the workaround is the correct way to install `dask-cuda`. No labels for CUDA should be used. It has extra channels though, so a shorter one would be:

```conda install -c rapidsai-nightly -c nvidia -c conda-forge dask-cuda```

The install instructions for `dask-cuda` on the `ucx-py` readme are incorrect. @jakirkham maybe you can update it in your PR https://github.com/rapidsai/ucx-py/pull/253

jakirkham:
Thanks @raydouglass! Have added a note there. Please let me know if it looks ok. 🙂

PR #160, opened 2019-10-15
FIX Remove all CUDA labels

mike-wendt:
The `cuda9.2` and `cuda10.0` labels are no longer needed and cause issues/confusion (see #159)

We now use community convention of `cudatoolkit=9.2` (or the appropriate CUDA version) during installs for setting the required CUDA version of a conda environment

This allows for multiple libraries requiring CUDA to work in the same environment


mike-wendt:
rerun tests

pentschev:
This is approved and looks good to me as well, merging.

PR #163, opened 2019-10-21
Set dask-cuda-worker protocol based on scheduler address

pentschev:
This is required for UCX to work properly, since using the default protocol (TCP) when scheduler is using UCX will not enable UCX.

### DIFF
--- a/dask_cuda/dask_cuda_worker.py
+++ b/dask_cuda/dask_cuda_worker.py
@@ -218,13 +218,21 @@ def main(
 
     preload_argv = kwargs.get("preload_argv", [])
     kwargs = {"worker_port": None, "listen_address": None}
-    t = Nanny
+    protocol = None
 
     if not scheduler and not scheduler_file and "scheduler-address" not in config:
         raise ValueError(
             "Need to provide scheduler address like\n"
             "dask-worker SCHEDULER_ADDRESS:8786"
         )
+    elif scheduler or "scheduler-address" in config:
+        if scheduler:
+            addr = scheduler
+        else:
+            addr = config["scheduler-address"]
+        protocol_address = addr.split("://")
+        if len(protocol_address) == 2:
+            protocol = protocol_address[0]
 
     if interface:
         if host:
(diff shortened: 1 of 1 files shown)


pentschev:
@mrocklin pinging again, can you take a look?

mrocklin:
My apologies @pentschev . I totally missed this.  Taking a look now.

mrocklin:
I see, so if I call `dask-cuda-worker ucx://...` then the Nanny will actually start up with `tcp://`?  If so, we should probably fix this upstream.  Workers and Nannies should probably use the same protocol that was handed to them.  I would have expected that this would already be resolved because of `tls://`, but I guess not.

pentschev:
> I see, so if I call dask-cuda-worker ucx://... then the Nanny will actually start up with tcp://? If so, we should probably fix this upstream. Workers and Nannies should probably use the same protocol that was handed to them. I would have expected that this would already be resolved because of tls://, but I guess not.

Exactly, this is what I'm trying to address with this PR. It actually took me a couple of days to figure this out, and it happens also with `dask-worker`. I can submit a PR to distributed, but I would like to know if you think I should address it as in https://github.com/rapidsai/dask-cuda/pull/163#issuecomment-544417846 or if you have a different idea.

mrocklin:
I think that if we get the address from the `scheduler_file` keyword, `scheduler-address` configuration value, or directly as an argument it shouldn't matter.  In all cases we should probably look at the protocol of the scheduler's address, and if our provided protocol is `None`, then use the scheduler's instead.  Does that fit with your thinking as well?

pentschev:
> I think that if we get the address from the scheduler_file keyword, scheduler-address configuration value, or directly as an argument it shouldn't matter.

Agreed.

> In all cases we should probably look at the protocol of the scheduler's address, and if our provided protocol is None, then use the scheduler's instead. Does that fit with your thinking as well?

Yes, I will then work on handling all cases then.

pentschev:
This is now solved in dask/distributed, closing.

Issue #79, opened 2019-06-26
Fix spilling test assertion that fails on rare occasions


pentschev:
See link below for an example:

https://gpuci.gpuopenanalytics.com/view/rapids-branch-gpu-matrix/job/gpuCI/job/dask-cuda/job/branches/job/dask-cuda-gpu-matrix-branch-0.8/CC_VERSION=5,CUDA=9.2,LINUX_VERSION=ubuntu16.04,PYTHON=3.6/37/console

cc @raydouglass

pentschev:
It has failed recently again:

https://gpuci.gpuopenanalytics.com/job/docker/job/tests/job/docker-test-daskcuda/CUDA_VERSION=10.0,LINUX_VERSION=ubuntu18.04,PYTHON_VERSION=3.7/76/consoleFull

cc @rlratzel

pentschev:
https://github.com/rapidsai/dask-cuda/issues/145 is another example of the failure reported here.

PR #170, opened 2019-11-06
Mark test_cudf_device_spill xfail

pentschev:
Marking test_cudf_device_spill, see https://github.com/rapidsai/dask-cuda/issues/79.


pentschev:
rerun test

pentschev:
rerun tests

pentschev:
@rlratzel there are some issues when CI builds the conda environment, apparently trying to use cudf 0.10, but that version isn't hardcoded anywhere in dask-cuda. It looks like this may be something in CI, could you take a look?

pentschev:
rerun tests

PR #174, opened 2019-11-13
Set LocalCUDACluster n_workers equal to the length of CUDA_VISIBLE_DE…

mrocklin:
…VICES


dask_cuda/local_cuda_cluster.py
```diff
@@ -54,13 +54,13 @@ def __init__(
...
     ):
-        if n_workers is None:
-            n_workers = get_n_gpus()
         if CUDA_VISIBLE_DEVICES is None:
             CUDA_VISIBLE_DEVICES = cuda_visible_devices(0)
         if isinstance(CUDA_VISIBLE_DEVICES, str):
             CUDA_VISIBLE_DEVICES = CUDA_VISIBLE_DEVICES.split(",")
         CUDA_VISIBLE_DEVICES = list(map(int, CUDA_VISIBLE_DEVICES))
+        if n_workers is None:
+            n_workers = len(CUDA_VISIBLE_DEVICES)
```
pentschev: What if the user doesn't pass `CUDA_VISIBLE_DEVICES`?

mrocklin: I believe that it is handled with the set of checks above just above.

pentschev: You're right. The `cuda_visible_devices(0)` got me confused, sorry for the false alarm.

pentschev:
rerun tests

pentschev:
rerun tests

pentschev:
Finally this passes, I'm not sure what exactly is the problem of using `LocalCUDACluster` on a sync test, but it was affecting the tests that ran after somehow. Making it async as the other `LocalCUDACluster` tests seems to solve the issue. I don't know whether this is caused by some issue in `LocalCUDACluster` or not, but if someone has an idea, please let me know.

I'm going ahead and merging this before CI changes its mind.

PR #175, opened 2019-11-15
Simplifying the code

madsbk:
This PR removes most of the preloading and makes `DGX` a class that inherit from `LocalCUDACluster`.


pentschev:
Let's not remove the `worker_spec` just yet, I introduced it as a sort of recipe that may be used with other `SpecCluster`s (such as `SSHCluster`), and we don't want to force users to write down the spec for CUDA-enabled workers by hand.

dask_cuda/dgx.py
```diff
@@ -1,116 +1,102 @@
...
-    return SpecCluster(
-        workers=spec, scheduler=scheduler, silence_logs=silence_logs, **kwargs
-    )
+    def new_worker_spec(self):
+        spec = super().new_worker_spec()
+        first_spec = next(iter(spec.values()))
+        dev = int(first_spec["options"]["env"]["CUDA_VISIBLE_DEVICES"].split(",")[0])
+        if self.set_ucx_net_devices:
+            first_spec["options"]["env"]["UCX_NET_DEVICES"] = "mlx5_%d:1" % (dev // 2)
+        return spec
```
pentschev: I think you're not configuring CPU affinity for workers here, that is required.

madsbk: Notice, I use `spec = super().new_worker_spec()` to call [`LocalCUDACluster.new_worker_spec()` ](https://github.com/rapidsai/dask-cuda/blob/953b21bb6a4226fdf5d55989568851863430490c/dask_cuda/local_cuda_cluster.py#L112). 
I think that configures the CPU affinity or  am I missing something?

pentschev: Ah I see, I missed that you subclassed `LocalCUDACluster`. Thanks for pointing it out!

madsbk:
> Let's not remove the `worker_spec` just yet, I introduced it as a sort of recipe that may be used with other `SpecCluster`s (such as `SSHCluster`), and we don't want to force users to write down the spec for CUDA-enabled workers by hand.

I have update `worker_spec` to not use the preloading anymore.

pentschev:
LGTM, thanks @madsbk !

PR #177, opened 2019-11-15
DGX inherit from LocalCUDACluster

madsbk:
This PR makes DGX inherit from `LocalCUDACluster`, which sets `UCX_NET_DEVICES` for each worker.

### DIFF
--- a/dask_cuda/tests/test_dgx.py
+++ b/dask_cuda/tests/test_dgx.py
@@ -1,85 +1,122 @@
-import pytest
-
-import psutil
 import os
-
+import multiprocessing as mp
+import pytest
+import numpy
 import dask.array as da
 
-from distributed.utils_test import gen_test
-from dask.distributed import Client
-
 from dask_cuda import DGX
-from dask_cuda.initialize import initialize
-from dask_cuda.utils import get_ucx_env
+from distributed import Client
 
-cupy = pytest.importorskip("cupy")
+mp = mp.get_context("spawn")
+ucp = pytest.importorskip("ucp")
+psutil = pytest.importorskip("psutil")
+
+if "ib0" not in psutil.net_if_addrs():
+    pytest.skip("Infiniband interface ib0 not found", allow_module_level=True)
 
 
-@pytest.mark.skipif(
-    "ib0" not in psutil.net_if_addrs(), reason="Infiniband interface ib0 not found"
-)
-@gen_test(timeout=20)
-async def test_dgx():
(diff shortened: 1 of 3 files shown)


dask_cuda/tests/test_dgx.py
```diff
@@ -1,85 +1,122 @@
...
+            res = da.from_array(numpy.arange(10000), chunks=(1000,))
+            res = res.sum().compute()
+            assert res == 49995000
+
+
+def test_tcp_only():
+    p = mp.Process(target=_test_tcp_only)
+    p.start()
+    p.join()
+    assert not p.exitcode
```
pentschev: Do we really need this? It seems to be the same as `test_default`, given that by default TCP will be used.

madsbk: What would yo8u suggest is the default protocol? Right now it is `DGX(protocol="ucx", ...)`. Should it be `tcp` instread?

pentschev: Oh sorry, I have missed that the default has been changed.

I'm not sure what the default should be, it makes sense to have UCX, but I'm not sure if it's not too early to make it default. I think I would keep TCP as default for the time being, @mrocklin do you have a suggestion to untie?

Regardless of what's the default, I see that we need both tests, so let's keep them as they are now.

madsbk: I agree, I think it is best to keep TCP as the default for now. When we are more confident that UCX works out-of-the-box we can change it to the default.

madsbk:
@pentschev @mrocklin, I think PR is ready for merging.

pentschev:
Changes look good to me. Thanks for working on this @madsbk. Merging.

dask_cuda/tests/test_dgx.py
mrocklin: > do you have a suggestion to untie?

Maybe this eventually?

```python
try:
    import ucp
except ImportError:
    protocol = "tcp"
else:
    protocol = "ucx"
```

madsbk: Yes

pentschev: Yeah, makes sense once we have stable packages and can make dask-cuda require them.

pentschev: Urm, nevermind. The suggestions is to make protocol _if_ ucp is available. That makes more sense.

PR #180, opened 2019-11-25
Set TCP for UCX tests

pentschev:


pentschev:
@madsbk do you see any downsides with the modifications of this PR? Those tests fail for me locally unless I set `UCX_TLS` environment variable without this PR.

pentschev:
Thanks @madsbk !

PR #188, opened 2019-11-29
Remove outdated xfail/importorskip test entries

pentschev:


pentschev:
@madsbk the xfail were added by you, but they don't ever fail for me, so I'm removing them. Are you ok with this change?

pentschev:
@madsbk just pinging, do you see any reason not to do this?

PR #192, opened 2019-12-03
Dask Performance Report

madsbk:
Added `--profile` to the benchmarks, which makes them write a Dask performance report.

Additionally,  I have renamed the benchmarks to `local_cupy_transpose_sum.py` and `local_cudf_merge.py`.

cc @pentschev @mrocklin


pentschev:
LGTM, thanks for adding this @madsbk

PR #194, opened 2019-12-03
Making the cupy benchmark use RMM

madsbk:
This PR makes `local_cupy_transpose_sum.py` use the RMM memory allocation pool when setting `--rmm`.

### DIFF
--- a/dask_cuda/benchmarks/local_cupy_transpose_sum.py
+++ b/dask_cuda/benchmarks/local_cupy_transpose_sum.py
@@ -11,6 +11,28 @@ from dask.distributed import Client, wait, performance_report
 from dask.utils import format_time, format_bytes, parse_bytes
 from dask_cuda.local_cuda_cluster import LocalCUDACluster
 
+try:
+    import cudf
+    from rmm._lib.device_buffer import DeviceBuffer
+
+    class RMMemory(cupy.cuda.memory.BaseMemory):
+        def __init__(self, size):
+            self.size = size
+            self.device_id = cupy.cuda.device.get_device_id()
+            if size > 0:
+                self.rmm_array = DeviceBuffer(size=size)
+                self.ptr = self.rmm_array.ptr
+            else:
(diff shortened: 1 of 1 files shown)


dask_cuda/benchmarks/local_cupy_transpose_sum.py
```diff
@@ -23,6 +45,10 @@
     ) as cluster:
         async with Client(cluster, asynchronous=True) as client:
 
+            if args.rmm:
+                await client.run(cudf.set_allocator, "default", pool=True)
+                await client.run(cupy.cuda.set_allocator, rmm_mem_allocator)
```
pentschev: What happens here if `args.rmm == True` but you couldn't import cudf or rmm? I think this will raise an error, no?

jakirkham:
Should handle this in RMM instead? ( https://github.com/rapidsai/rmm/issues/191 )

dask_cuda/benchmarks/local_cupy_transpose_sum.py
madsbk: Yes :)
Do you want me to raise a descriptive exception?

madsbk:
> Should handle this in RMM instead? ( [rapidsai/rmm#191](https://github.com/rapidsai/rmm/issues/191) )

Good point @jakirkham

dask_cuda/benchmarks/local_cupy_transpose_sum.py
pentschev: I think that would be nice, either that or stating in the output that RMM wasn't imported and thus, skipped. Otherwise, the first persons who tries that and it fails will report this as a bug. :)

jakirkham:
Now that we can hook RMM into CuPy more easily, should we update this?

madsbk:
Superseded by https://github.com/rapidsai/dask-cuda/pull/220

PR #196, opened 2019-12-03
Use ucx-py from rapidsai-nightly in CI

pentschev:


pentschev:
Thanks @mike-wendt, will merge as soon as the new build passes.

pentschev:
rerun tests

PR #200, opened 2019-12-04
LocalCUDACluster sets closest network device

madsbk:
This PR makes `LocalCUDACluster` sets the closest network device for UCX explicitly

### DIFF
--- a/dask_cuda/local_cuda_cluster.py
+++ b/dask_cuda/local_cuda_cluster.py
@@ -15,7 +15,7 @@ from .utils import (
 )
 
 
-def _ucx_net_devices(dev):
+def _ucx_net_devices(dev, ucx_net_devices):
     net_dev = None
     if callable(ucx_net_devices):
         net_dev = ucx_net_devices(dev)
@@ -86,7 +86,7 @@ class LocalCUDACluster(LocalCluster):
         CUDA_VISIBLE_DEVICES=None,
         data=None,
         local_directory=None,
-        ucx_net_devices=None
+        ucx_net_devices=None,
         **kwargs,
     ):
         if CUDA_VISIBLE_DEVICES is None:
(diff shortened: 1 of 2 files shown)


pentschev:
I would like to hold on to this for now. There are many potential issues that we haven't really figured yet. For example, what happens if we select the closest interface but that's an interface that's not configure (e.g., has no IP address or is physically disconnected)?

pentschev:
As a complement to my question above, I have to point out that `DGX` is configured as that and should work for the DGXs we have been testing on because we know the interfaces are configured, but it may not work even on other DGXs if the connections are not up.

madsbk:
> As a complement to my question above, I have to point out that `DGX` is configured as that and should work for the DGXs we have been testing on because we know the interfaces are configured, but it may not work even on other DGXs if the connections are not up.

Make sense.

pentschev:
@madsbk I was thinking about this and I think we should apply your changes but allow users to choose anyway. My idea is we have a `net_devices` kwarg that allows the user to pass a function taking the GPU index as argument that will be used to calculate the interface name based on that (as we do in https://github.com/rapidsai/dask-cuda/blob/branch-0.11/dask_cuda/dgx.py#L110), alternatively, the user can pass a string, for example `"mlx5_0:1"` (in case only one interface is configured, as in Summit) or a string `"auto"` that will try to import `TopologicalDistance`, raising an exception if it can't or use the result instead (as you've done), and finally `None` (which would be the default) that will simply not set anything.

What do you think?

pentschev:
Sorry, closed by accident.

dask_cuda/local_cuda_cluster.py
```diff
@@ -129,16 +129,29 @@ def new_worker_spec(self):
...
+
+            dev = int(visible_devices.split(",")[0])
+            net_dev = ""
+            td = TopologicalDistance()
+            ibs = td.get_cuda_distances_from_device_index(dev, "openfabrics")
+            if len(ibs) > 0:
+                net_dev += ibs[0]["name"] + ":1,"
+            ifnames = td.get_cuda_distances_from_device_index(dev, "network")
+            if len(ifnames) > 0:
+                net_dev += ifnames[0]["name"]
```
pentschev: Does it make sense to have multiple interfaces in `UCX_NET_DEVICES`, here we would have `"mlx5_0:1,enp1s0f0"`, do you know how UCX makes use of multiple interfaces? I haven't tried that so far to see what's the behavior.

madsbk: On DGX I actually gets `"mlx5_0:1,ib0"`. 
The ideas is that if UCX isn't configured with IB, it will use the TCP interface.

madsbk:
I think that is a great idea :)

dask_cuda/local_cuda_cluster.py
pentschev: You're right, `ib0` is the closest network device. Is `ib0` also valid for UCX? I've always seen the usage of `mlx5_*`, are both the same for UCX?

pentschev:
> I think that is a great idea :)

Cool, would like me to do the changes or would you like to do it yourself?

madsbk:
If you have the time, please do :)

pentschev:
@madsbk could you please take a look? The failing CI test also passes here, so let's ignore the codecov error and get this merged to unblock this repo and I'll address that in a follow-up PR.

madsbk:
Looks good :)

PR #202, opened 2019-12-05
Added --runs to merge benchmark

madsbk:
Adds `--runs` to the merge benchmark. Additionally, we now call `wait()` on the generate input, which makes a difference: 
```python
wait(ddf_base)
wait(ddf_other)
```

### Motivation
cuDF use `numba.cuda.jit`, which has a significant compile overhead first time a function is called with new input dtypes. Thus, it is useful to consider multiple runs when profiling. 
The `--profile` argument now only profiles the _last_ run.

### DIFF
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -97,33 +97,22 @@ def get_random_ddf(chunk_size, num_chunks, frac_match, chunk_type):
     return new_dd_object(graph, name, meta, divisions)
 
 
-def main(args):
-    n_workers = len(args.devs.split(","))
-
-    # Set up workers on the local machine
-    cluster = LocalCUDACluster(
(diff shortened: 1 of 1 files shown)


pentschev:
Nice feature! LGTM.

madsbk:
@pentschev, do you know why the CI is failing?

pentschev:
I am looking at it now, but it seems to be something in distributed? Not sure either.

pentschev:
@madsbk https://github.com/rapidsai/dask-cuda/pull/200/files#diff-6e16c50881bd324a8a59444ae2cd2f7f seems to have fixed it locally for me, it seems like we need `UCX_SOCKADDR_TLS_PRIORITY=sockcm` afterall.

pentschev:
For some reason `DGX` seems to work without that though. Do you think we should replace `SOCKADDR_TLS_PRIORITY=all` with `SOCKADDR_TLS_PRIORITY=sockcm` in UCX-Py?

madsbk:
Yeah, that is probably a good idea. It shouldn't effect performance anyway.

PR #205, opened 2019-12-05
Move UCX code to LocalCUDACluster and deprecate DGX

pentschev:


pentschev:
@quasiben @mrocklin @randerzander @beckernick for visibility

dask_cuda/local_cuda_cluster.py
```diff
@@ -169,15 +248,17 @@ def new_worker_spec(self):
...
-                },
+                "env": {"CUDA_VISIBLE_DEVICES": visible_devices,},
                 "plugins": {CPUAffinity(get_cpu_affinity(worker_count))},
             }
         )
 
-        net_dev = _ucx_net_devices(visible_devices.split(",")[0], self.ucx_net_devices)
-        if net_dev is not None:
-            spec["options"]["env"]["UCX_NET_DEVICES"] = net_dev
+        print(self.set_ucx_net_devices, self.ucx_net_devices)
```
madsbk: Hanging print

pentschev: Nice catch!

jakirkham:
Are we wanting to target `0.11.0` with this change or should this move to `0.12.0`?

pentschev:
I see no reason to hold this for 0.12 only, do you have any objections @jakirkham ?

jakirkham:
It just seems like a large change right before code freeze is all. I trust your judgement. Just want us to be cautious is all.

pentschev:
> It just seems like a large change right before code freeze is all. I trust your judgement. Just want us to be cautious is all.

But the whole functionality is still available, I'm just moving it around to make things cleaner and more obvious. Right now I see a lot of confusion when people should use `DGX` or `LocalCUDACluster`, which is understandable. With this PR they can use either one, but using `DGX` will just raise a warning telling people to switch.

pentschev:
@mike-wendt @raydouglass @dillon-cullinan any chance we could still get this in? I was trying to get it merged before freeze, but CI seems pretty overloaded and tests are taking longer to run. If not, that's probably ok too.

randerzander:
This would be a very good thing to get into 0.11 if possible.

We have people using nvlink on non-DGX hardware who would find it strange to use the DGX API.

pentschev:
> We have people using nvlink on non-DGX hardware who would find it strange to use the DGX API.

Yeah, that's another thing I wanted to get rid of, it seems counterintuitive to use the `DGX` class for other machines.

raydouglass:
> any chance we could still get this in?

Sure, if it can be merged before midnight PT tonight! Hopefully it passes CI and you're good to go soon.

pentschev:
Great! CI should pass, codecov is reduced because some of the tests would require a DGX to run, so we'll need to ignore those.

@madsbk could you take a last look and approve if you're good?

pentschev:
I guess @madsbk isn't available at this time anymore, so I'm going ahead and merging this before midnight comes!

PR #207, opened 2019-12-05
Update changelog for 0.11

pentschev:


pentschev:
@mike-wendt @raydouglass @dillon-cullinan can we still get this in? I was working on it as code freeze landed.

PR #212, opened 2019-12-06
[REVIEW] Add ucx-py dependency to conda recipe

raydouglass:


pentschev:
For CI, it's indeed important to have UCX-Py installed. I'm not sure if we should make it a dependency in the conda package though, since dask-cuda should be perfectly usable without it.

Ideas @madsbk @quasiben @jakirkham @mrocklin ?

quasiben:
dask-cuda as builid dependency definitely makes sense.  ucx-py is not really a dependency for dask-cuda to work properly.  ucx-py is still nascent and making it a dependency is a little risky.  Additionally,   If dask-cuda wants to continue supporting pip, depending on ucx-py would break this as it currently only supports conda

pentschev:
Indeed. @raydouglass could we keep UCX-Py as a dependency to CI but remove that from the conda package?

pentschev:
LGTM, thanks @raydouglass

PR #201, opened 2019-12-04
Expand cudf-merge benchmark

rjzamora:
It seems that the C++/UCX code that motivated this benchmark actually requires both the left and right dataframes to be dissordered such that there will be all-to-all communication between the partitions for **both** dataframes.  The current version of `local_cudf_merge.py` uses a simple `arange` operation to define the key column for the left (base) dataframe.  Therefore, the left dataframe does **not** need to be shuffled during the merge operation.

This PR modifies the data-generation stage for the left (base) dataframe, so that sorting the column will require all-to-all communication between partitions.  With that said, the data-generation alone does **not** cause the left dataframe to be sorted during the merge.  To do that, we must explicitly reset the index.  I added the `--set-index` option here to accomplish this.

Some execution examples...

[**EDIT:** Results updated to align with recent benchmark and dask changes]

**WITHOUT Index Reset** (Same as before this PR):
```
Merge benchmark
--------------------------
Chunk-size  | 100000000
Frac-match  | 0.3
Ignore-size | 1.05 MB
Protocol    | ucx
Device(s)   | 0,1,2,3
==========================
Total time  | 1.32 s
Total time  | 913.19 ms
Total time  | 789.18 ms
==========================
(w1,w2)     | 25% 50% 75% (total nbytes)
--------------------------
(00,01)     | 4.50 GB/s 5.02 GB/s 5.43 GB/s (6.60 GB)
(00,02)     | 4.65 GB/s 5.09 GB/s 5.64 GB/s (6.60 GB)
(00,03)     | 4.66 GB/s 5.21 GB/s 5.96 GB/s (6.60 GB)
(01,00)     | 4.29 GB/s 4.54 GB/s 5.89 GB/s (6.60 GB)
(01,02)     | 4.85 GB/s 5.12 GB/s 8.38 GB/s (6.60 GB)
(01,03)     | 4.37 GB/s 4.76 GB/s 5.14 GB/s (6.60 GB)
(02,00)     | 4.46 GB/s 5.24 GB/s 5.64 GB/s (4.20 GB)
(02,01)     | 4.30 GB/s 5.71 GB/s 12.88 GB/s (9.00 GB)
(02,03)     | 4.78 GB/s 4.97 GB/s 5.70 GB/s (6.60 GB)
(03,00)     | 4.59 GB/s 5.23 GB/s 14.43 GB/s (9.00 GB)
(03,01)     | 3.95 GB/s 4.70 GB/s 5.10 GB/s (4.20 GB)
[...]

PR #213, opened 2019-12-06
Follow-up revision of local_cudf_merge benchmark

rjzamora:
This is a revision of the benchmark changes in #201 

Upon a follow-up inspection of `local_cudf_merge.py`,  I realized that the key of the left (base) index needs to be shuffled differently to achieve the intended behavior.  This PR removes the `--set-index` option, and instead uses shuffling by default.  To remove the shuffling behavior, one can add the `--no-shuffle` argument.

**WITH Shuffle**:

```
Merge benchmark
--------------------------
Chunk-size  | 100000000
Frac-match  | 0.3
Ignore-size | 1.05 MB
Protocol    | ucx
Device(s)   | 0,1,2,3
==========================
Total time  | 1.65 s
Total time  | 1.57 s
Total time  | 1.66 s
==========================
--------------------------
(00,01)     | 4.31 GB/s 4.40 GB/s 4.80 GB/s (9.60 GB)
(00,02)     | 4.47 GB/s 4.73 GB/s 4.90 GB/s (7.20 GB)
(00,03)     | 5.43 GB/s 5.81 GB/s 8.34 GB/s (12.00 GB)
(01,00)     | 3.87 GB/s 4.85 GB/s 5.69 GB/s (9.60 GB)
(01,02)     | 5.45 GB/s 5.92 GB/s 6.55 GB/s (12.00 GB)
(01,03)     | 4.72 GB/s 4.79 GB/s 5.52 GB/s (7.20 GB)
(02,00)     | 4.03 GB/s 4.45 GB/s 5.05 GB/s (7.20 GB)
(02,01)     | 4.95 GB/s 5.92 GB/s 7.00 GB/s (12.00 GB)
(02,03)     | 4.55 GB/s 4.96 GB/s 5.98 GB/s (9.60 GB)
(03,00)     | 4.88 GB/s 5.25 GB/s 6.58 GB/s (12.00 GB)
(03,01)     | 4.26 GB/s 4.93 GB/s 5.55 GB/s (7.20 GB)
(03,02)     | 4.80 GB/s 5.86 GB/s 6.33 GB/s (9.60 GB)
```

**Without Shuffling**:
```
Merge benchmark
--------------------------
Chunk-size  | 100000000
Frac-match  | 0.3
Ignore-size | 1.05 MB
Protocol    | ucx
Device(s)   | 0,1,2,3
==========================
Total time  | 1.75 s
Total time  | 910.33 ms
Total time  | 901.90 ms
==========================
--------------------------
(00,01)     | 4.43 GB/s 4.92 GB/s 5.21 GB/s (4.20 GB)
(00,02)     | 4.32 GB/s 4.77 GB/s 7.80 GB/s (9.00 GB)
(00,03)     | 4.07 GB/s 4.86 GB/s 5.72 GB/s (6.60 GB)
(01,00)     | 3.66 GB/s 4.32 GB/s 5.68 GB/s (6.60 GB)
(01,02)     | 4.89 GB/s 5.11 GB/s 5.22 GB/s (4.20 GB)
(01,03)     | 4.30 GB/s 4.60 GB/s 5.34 GB/s (6.60 GB)
(02,00)     | 4.47 GB/s 5.68 GB/s 11.40 GB/s (9.00 GB)
(02,01)     | 5.39 GB/s 5.44 GB/s 6.22 GB/s (4.20 GB)
(02,03)     | 4.96 GB/s 5.77 GB/s 6.02 GB/s (4.20 GB)
(03,00)     | 4.58 GB/s 4.80 GB/s 5.13 GB/s (4.20 GB)
(03,01)     | 4.41 GB/s 4.98 GB/s 9.85 GB/s (9.00 GB)
(03,02)     | 4.63 GB/s 4.76 GB/s 4.99 GB/s (4.20 GB)
```

### DIFF
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -217,6 +217,11 @@ def main(args):
     print(f"Ignore-size | {format_bytes(args.ignore_size)}")
     print(f"Protocol    | {args.protocol}")
     print(f"Device(s)   | {args.devs}")
+    print(f"rmm-pool    | {(not args.no_pool_allocator)}")
+    if args.protocol == "ucx":
+        print(f"tcp         | {(not args.no_tcp)}")
+        print(f"ib          | {(not args.no_ib)}")
+        print(f"nvlink      | {(not args.no_nvlink)}")
     print("==========================")
     for took in took_list:
         print(f"Total time  | {format_time(took)}")
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -30,18 +30,16 @@ def generate_chunk(i_chunk, local_size, num_chunks, chunk_type, frac_match):
         #
         # "payload" column is a random permutation of the chunk_size
 
-        sub_local_size = math.ceil(local_size / num_chunks)
-        arrays = []
-        for i in range(num_chunks):
-            bgn = (local_size * i) + (sub_local_size * i_chunk)
-            end = bgn + sub_local_size
-            ar = cupy.arange(bgn, stop=end, dtype="int64")
-            arrays.append(cupy.random.permutation(ar))
-        key_array_match = cupy.concatenate(tuple(arrays), axis=0)
+        start = local_size * i_chunk
+        stop = start + local_size
+
+        parts_array = cupy.arange(num_chunks, dtype="int64")
+        suffle_array = cupy.repeat(parts_array, math.ceil(local_size / num_chunks))
 
         df = cudf.DataFrame(
             {
-                "key": cupy.random.permutation(key_array_match[:local_size]),
+                "key": cupy.arange(start, stop=stop, dtype="int64"),
+                "shuffle": cupy.random.permutation(suffle_array)[:local_size],
                 "payload": cupy.random.permutation(
                     cupy.arange(local_size, dtype="int64")
                 ),
@@ -89,10 +87,10 @@ def generate_chunk(i_chunk, local_size, num_chunks, chunk_type, frac_match):
     return df
 
 
-def get_random_ddf(chunk_size, num_chunks, frac_match, chunk_type):
+def get_random_ddf(chunk_size, num_chunks, frac_match, chunk_type, args):
 
     parts = [chunk_size for i in range(num_chunks)]
-    meta = generate_chunk(0, 4, 1, None, None)
+    meta = generate_chunk(0, 4, 1, chunk_type, None)
     divisions = [None] * (len(parts) + 1)
 
     name = "generate-data-" + tokenize(chunk_size, num_chunks, frac_match, chunk_type)
@@ -102,24 +100,31 @@ def get_random_ddf(chunk_size, num_chunks, frac_match, chunk_type):
         for i, part in enumerate(parts)
     }
 
-    return new_dd_object(graph, name, meta, divisions)
+    ddf = new_dd_object(graph, name, meta, divisions)
+
+    if chunk_type == "build":
+        if args.shuffle:
+            divisions = [i for i in range(num_chunks)] + [num_chunks]
+            return ddf.set_index("shuffle", divisions=tuple(divisions))
+        else:
+            del ddf["shuffle"]
+
+    return ddf
 
 
 def run(args, write_profile=None):
     # Generate random Dask dataframes
     ddf_base = get_random_ddf(
-        args.chunk_size, args.n_workers, args.frac_match, "build"
+        args.chunk_size, args.n_workers, args.frac_match, "build", args
     ).persist()
     ddf_other = get_random_ddf(
-        args.chunk_size, args.n_workers, args.frac_match, "other"
+        args.chunk_size, args.n_workers, args.frac_match, "other", args
     ).persist()
     wait(ddf_base)
     wait(ddf_other)
 
     # Lazy merge/join operation
     ddf_join = ddf_base.merge(ddf_other, on=["key"], how="inner")
-    if args.set_index:
-        ddf_join = ddf_join.set_index("key")
 
     # Execute the operations to benchmark
     if write_profile is not None:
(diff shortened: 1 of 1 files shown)


rjzamora:
Note:  The latest changes here have extended the optional arguments:

```
usage: local_cudf_merge.py [-h] [-d DEVS] [-p {tcp,ucx}] [-c n]
                           [--ignore-size nbytes] [--frac-match FRAC_MATCH]
                           [--no-pool-allocator] [--profile PATH]
                           [--no-shuffle] [--markdown] [--runs RUNS] [--no-ib]
                           [--no-tcp] [--no-nvlink]

Merge (dask/cudf) on LocalCUDACluster benchmark

optional arguments:
  -h, --help            show this help message and exit
  -d DEVS, --devs DEVS  GPU devices to use (default "0").
  -p {tcp,ucx}, --protocol {tcp,ucx}
                        The communication protocol to use.
  -c n, --chunk-size n  Chunk size (default 1_000_000)
  --ignore-size nbytes  Ignore messages smaller than this (default "1 MB")
  --frac-match FRAC_MATCH
                        Fraction of rows that matches (default 0.3)
  --no-pool-allocator   Disable the RMM memory pool
  --profile PATH        Write dask profile report (E.g. dask-report.html)
  --no-shuffle          Don't shuffle the keys of the left (base) dataframe.
  --markdown            Write output as markdown
  --runs RUNS           Number of runs
  --no-ib               Disable infiniband over ucx.
  --no-tcp              Disable tcp over ucx.
  --no-nvlink           Disable NVLink over ucx.
```

quasiben:
This looks good to me and I especially appreciate the the additional configs options similar to the dask-cuda-worker (--no-ib, --no-nvlink).  It looks like there are now some merge conflicts but after those are resolved I think this should be good.  @madsbk any additional thoughts ?

rjzamora:
Thanks for taking a look @quasiben ! I'll clean up the conflicts

quasiben:
With tests passing this looks good.  I @mrocklin do you have commit rights to this repo ?  If so can you review (I believe @pentschev is out on PTO)

dask_cuda/benchmarks/local_cudf_merge.py
```diff
@@ -102,16 +103,25 @@ def get_random_ddf(chunk_size, num_chunks, frac_match, chunk_type):
...
 
-    return new_dd_object(graph, name, meta, divisions)
+    ddf = new_dd_object(graph, name, meta, divisions)
+
+    if chunk_type == "build":
+        if not args.no_shuffle:
+            divisions = [i for i in range(num_chunks)] + [num_chunks]
+            return ddf.set_index("shuffle", divisions=tuple(divisions))
+        else:
+            del ddf["shuffle"]
```
mrocklin: Just so we're clear, calling `set_index` will be somewhat expensive.

If we're just generating random data then maybe we can generated it in an unordered way to start?  Or is there some particular dataset that we're using?  

If we do need to call `set_index` (which is about as expensive as a merge itself) then perhaps we should try to separate data generation from the benchmark with persist and wait calls?

rjzamora: We might not need to call `set_index` here.  However, we **are** calling `persist` and `wait` before the merge time is measured

dask_cuda/benchmarks/local_cudf_merge.py
```diff
@@ -260,6 +302,13 @@ def parse_args():
         help="Call set_index on the key column to sort the joined dataframe.",
     )
     parser.add_argument("--runs", default=3, type=int, help="Number of runs")
+    parser.add_argument(
+        "--no-ib", action="store_true", help="Disable infiniband over ucx."
+    )
+    parser.add_argument("--no-tcp", action="store_true", help="Disable tcp over ucx.")
+    parser.add_argument(
+        "--no-nvlink", action="store_true", help="Disable NVLink over ucx."
+    )
```
pentschev: I think it's usually less confusing to have variables that enable/turn on things, rather than disable/turn off. For that, I would suggest that we have variables such as `--enable-infiniband`, `--enable-tcp` and `--enable-nvlink`. This would also make things consistent with those variables in `LocalCUDACluster`.

dask_cuda/benchmarks/local_cudf_merge.py
mrocklin: OK, if that's the case then my concerns aren't that big of a deal.

dask_cuda/benchmarks/local_cudf_merge.py
pentschev: Of course, the default values for those would then be `True`, and we would then need two variables, `enable/disable`, as in https://github.com/rapidsai/dask-cuda/blob/branch-0.12/dask_cuda/dask_cuda_worker.py#L155-L169.

dask_cuda/benchmarks/local_cudf_merge.py
rjzamora: @mrcklin - You are correct that the `set_index`-based shuffle is nearly as expensive as the merge operation.   It may indeed make sense to either (1) separately report the set_index performance, or (2) be "smarter" about the data generation...  I will think a bit about this and make a note here.

dask_cuda/benchmarks/local_cudf_merge.py
rjzamora: > Of course, the default values for those would then be True, and we would then need two variables, enable/disable, as in https://github.com/rapidsai/dask-cuda/blob/branch-0.12/dask_cuda/dask_cuda_worker.py#L155-L169.

Right - I was hoping to have all transports turned on in the benchmark by default, hence the "negative" command-line options.  I am not familiar with `click` or the `enable/disable` syntax, but that might be reasonable.  Should we just change the `argparse` flags to require explicit `True/False` values (e.g. `python script.py --enable-nvlink True`)?

pentschev: I think we would be better off with something like the following. Unfortunately it allows a user to specify both (and whatever comes last is valid), but I guess that would be less cluttered for users.

```python
parser.add_argument('--enable-tcp-over-ucx', dest='enable_tcp_over_ucx', action='store_true')
parser.add_argument('--disable-tcp-over-ucx'', dest='enable_tcp_over_ucx', action='store_false')
parser.set_defaults(enable_tcp_over_ucx=True)
```

rjzamora: > I think we would be better off with something like the following...

Okay - That looks fine to me. Thanks!

pentschev: Thanks for the changes Rick!

pentschev:
LGTM, thanks for the work here @rjzamora ! Merging.

PR #218, opened 2019-12-17
[REVIEW] Add pypi upload script

raydouglass:
Adds back pip builds and uploading to pypi


pentschev:
Nice @raydouglass , this will make a lot of people happy!

PR #223, opened 2020-01-06
Fix dask-cuda-worker --interface/--net-devices docs

pentschev:


dask_cuda/dask_cuda_worker.py
```diff
@@ -78,7 +78,12 @@
     "See --interface.",
 )
 @click.option(
-    "--interface", type=str, default=None, help="Network interface like 'eth0' or 'ib0'"
+    "--interface",
+    type=str,
+    default=None,
+    help="The external interface used to connect to the scheduler, usually "
+    "the ethernet interface is used for connection, not the InfiniBand (if "
```
quasiben: "the ethernet interface is used for connection, not InfiniBand (if "

dask_cuda/dask_cuda_worker.py
```diff
@@ -171,8 +176,10 @@
     "--net-devices",
     type=str,
     default=None,
-    help="Network interface to establish UCX connection, "
-    "usually the Ethernet interface, like 'eth0' or 'enp1s0f0'",
+    help="When None (default), 'UCX_NET_DEVICES' will be left to its default. "
+    "otherwise, it must be a non-empty string with the interface name. Normally "
```
quasiben: "otherwise, it must be a non-empty string with the interface name. Normally, "

pentschev:
@quasiben added some changes, let me know if you agree with those.

PR #226, opened 2020-01-14
[REVIEW] Explicit Worker Communication 

madsbk:
This PR implements support for explicit UCX communication together with Dask. 
The idea is that Dask spawns workers and distribute data as usually while the user can submit jobs on the workers that communicate manually using UCX.

Additionally, this PR implements distributed cudf merge, which uses this new explicit communication approach.

#### DGX-1 four GPUs

More than **4 times speedup** compared the regular dask implementation of merge. 
Notice, this is **without** setting `UCX_CUDA_IPC_CACHE=y` and [`UCXPY_COMM_BUFFER=y`](https://github.com/rapidsai/ucx-py/pull/378).

##### Explicit Communication
```
$ python local_cudf_merge.py -p ucx -d 4,5,6,7 -c 100000000 --runs 10 -b explicit-comms --disable-infiniband  --markdown
Merge benchmark
-------------------------------
backend        | explicit-comms
rows-per-chunk | 100000000
protocol       | ucx
device(s)      | 4,5,6,7
rmm-pool       | True
frac-match     | 0.3
tcp            | True
ib             | False
nvlink         | True
data-processed | 12.80 GB
===============================
Wall-clock     | Throughput
-------------------------------
636.01 ms      | 20.13 GB/s
410.94 ms      | 31.15 GB/s
402.05 ms      | 31.84 GB/s
420.19 ms      | 30.46 GB/s
429.11 ms      | 29.83 GB/s
438.45 ms      | 29.19 GB/s
405.90 ms      | 31.53 GB/s
435.58 ms      | 29.39 GB/s
408.21 ms      | 31.36 GB/s
431.09 ms      | 29.69 GB/s
===============================
```

##### Regular Dask
```
$ python local_cudf_merge.py -p ucx -d 4,5,6,7 -c 100000000 --runs 10 -b dask --disable-
Merge benchmark
-------------------------------
backend        | dask
rows-per-chunk | 100000000
protocol       | ucx
device(s)      | 4,5,6,7
rmm-pool       | True
frac-match     | 0.3
tcp            | True
ib             | False
nvlink         | True
data-processed | 12.80 GB
===============================
Wall-clock     | Throughput
-------------------------------
2.17 s         | 5.91 GB/s
1.74 s         | 7.34 GB/s
1.80 s         | 7.10 GB/s
1.85 s         | 6.94 GB/s
1.69 s         | 7.55 GB/s
1.75 s         | 7.30 GB/s
1.76 s         | 7.27 GB/s
2.00 s         | 6.41 GB/s
2.28 s         | 5.61 GB/s
1.80 s         | 7.12 GB/s
===============================
```
<details>
<summary>Worker-Worker Transfer Rates</summary>

```
-------------------------------
(04,05)     | 4.28 GB/s 4.91 GB/s 6.37 GB/s (35.30 GB)
(04,06)     | 4.39 GB/s 5.76 GB/s 6.11 GB/s (28.80 GB)
(04,07)     | 4.39 GB/s 5.49 GB/s 6.46 GB/s (28.80 GB)
(05,04)     | 4.13 GB/s 5.12 GB/s 5.72 GB/s (31.90 GB)
(05,06)     | 4.36 GB/s 5.71 GB/s 6.50 GB/s (29.50 GB)
(05,07)     | 4.47 GB/s 5.45 GB/s 6.06 GB/s (29.50 GB)
(06,04)     | 4.31 GB/s 5.46 GB/s 6.26 GB/s (30.20 GB)
(06,05)     | 4.46 GB/s 5.41 GB/s 6.72 GB/s (27.10 GB)
(06,07)     | 4.88 GB/s 5.88 GB/s 7.16 GB/s (33.60 GB)
(07,04)     | 4.73 GB/s 5.81 GB/s 6.25 GB/s (31.20 GB)
(07,05)     | 4.97 GB/s 5.52 GB/s 5.99 GB/s (28.80 GB)
(07,06)     | 5.04 GB/s 5.59 GB/s 7.05 GB/s (36.00 GB)
```
</details>


mrocklin:
This is very cool too see.  Two questions: 

1. Do we know yet what is slowing down the Dask version?  I'm personally very curious to find out what that is.  I suspect that that will affect many workloads.  
2.  What is this like across eight GPUs?  Historically our performance has dropped off when we leave the island of nvlink connectivity.  I'm curious what that looks like here.

To help deal with the first question I think it might be useful to try using Dask Comm objects rather than ucx endpoints. I think that this would help us to isolate the problem by engaging about half of the Dask machinery.

Anyway, just some thoughts.  It's cool to see these results.

rjzamora:
> What is this like across eight GPUs? Historically our performance has dropped off when we leave the island of nvlink connectivity. I'm curious what that looks like here.

I was also interested in this, so I ran some quick tests on a dgx (wioth infiniband both disabled and enables).  It appears that performance still suffers significantly, but the overall performance is still much better...

**Dask backend,  Infiniband Disabled**
```
Merge benchmark
-------------------------------
backend        | dask
rows-per-chunk | 100000000
protocol       | ucx
device(s)      | 0,1,2,3,4,5,6,7
rmm-pool       | True
frac-match     | 0.3
tcp            | True
ib             | False
nvlink         | True
data-processed | 25.60 GB
===============================
Wall-clock     | Throughput
-------------------------------
22.48 s        | 1.14 GB/s
20.71 s        | 1.24 GB/s
19.72 s        | 1.30 GB/s
19.82 s        | 1.29 GB/s
14.77 s        | 1.73 GB/s
26.18 s        | 977.80 MB/s
33.76 s        | 758.39 MB/s
23.10 s        | 1.11 GB/s
19.61 s        | 1.31 GB/s
28.20 s        | 907.76 MB/s
===============================
```

**Explicit Comms,  Infiniband Disabled**
```
Merge benchmark
-------------------------------
backend        | explicit-comms
rows-per-chunk | 100000000
protocol       | ucx
device(s)      | 0,1,2,3,4,5,6,7
rmm-pool       | True
frac-match     | 0.3
tcp            | True
ib             | False
nvlink         | True
data-processed | 25.60 GB
===============================
Wall-clock     | Throughput
-------------------------------
7.83 s         | 3.27 GB/s
6.34 s         | 4.04 GB/s
6.77 s         | 3.78 GB/s
6.86 s         | 3.73 GB/s
6.50 s         | 3.94 GB/s
6.26 s         | 4.09 GB/s
6.58 s         | 3.89 GB/s
6.36 s         | 4.03 GB/s
5.98 s         | 4.28 GB/s
6.37 s         | 4.02 GB/s
===============================
```

**Dask backend,  Infiniband Enabled**
```
Merge benchmark
-------------------------------
backend        | dask
rows-per-chunk | 100000000
protocol       | ucx
device(s)      | 0,1,2,3,4,5,6,7
rmm-pool       | True
frac-match     | 0.3
tcp            | True
ib             | True
nvlink         | True
data-processed | 25.60 GB
===============================
Wall-clock     | Throughput
-------------------------------
13.21 s        | 1.94 GB/s
10.14 s        | 2.52 GB/s
9.74 s         | 2.63 GB/s
7.65 s         | 3.34 GB/s
7.28 s         | 3.52 GB/s
5.83 s         | 4.39 GB/s
4.95 s         | 5.18 GB/s
4.88 s         | 5.24 GB/s
5.45 s         | 4.70 GB/s
6.57 s         | 3.89 GB/s
===============================
```

**Explicit Comms,  Infiniband Enabled**
```
Merge benchmark
-------------------------------
backend        | explicit-comms
rows-per-chunk | 100000000
protocol       | ucx
device(s)      | 0,1,2,3,4,5,6,7
rmm-pool       | True
frac-match     | 0.3
tcp            | True
ib             | True
nvlink         | True
data-processed | 25.60 GB
===============================
Wall-clock     | Throughput
-------------------------------
4.87 s         | 5.26 GB/s
3.70 s         | 6.92 GB/s
2.76 s         | 9.29 GB/s
3.44 s         | 7.45 GB/s
2.32 s         | 11.02 GB/s
2.34 s         | 10.93 GB/s
1.92 s         | 13.32 GB/s
1.72 s         | 14.92 GB/s
1.29 s         | 19.89 GB/s
1.92 s         | 13.34 GB/s
===============================
```

dask_cuda/explicit_comms/comms.py
```diff
@@ -0,0 +1,184 @@
...
+                    self.sessionId,
+                    coroutine,
+                    args,
+                    workers=[worker],
+                    pure=False,
+                )
+            )
+        return self.client.gather(ret)
+
+    def dataframe_operation(self, coroutine, df1, df2):
```
rjzamora: Perhaps this should be more flexible, and accept a single dataframe/series as well?  I know this is only supporting merge operations for now, but i can imagine doing something similar for `set_index`/sorting.

madsbk: Agree, the plan is to support any number of inputs and more than just dataframes.

madsbk:
@mrocklin, I am trying to use Dask Comm objects but how should a message look like?

 Each worker now use `listen()` and `connect()` to initialize a new dedicated connection between each worker using the two functions:
```python
distributed.comm.listen("tcp://127.0.0.1", server_handler)
ep = await distributed.comm.connect(address)
```
That works, but when I try to send a message the serialization fails: 
```python 
>>> await ep.write(np.arange(10))
Exception: TypeError("can not serialize 'numpy.ndarray' object")

>>> await ep.write(np.arange(10), serializers=("cuda", "dask", "numpy", "pickle", "error"))
Exception: TypeError("can not serialize 'numpy.ndarray' object")
```
What am I doing wrong?

mrocklin:
I believe that you need the following:

```python
from distributed.protocol import to_serialize

comm.write(to_serialize(obj))
```

This is more commonly used as follows:

```python
msg = {"x": 123, "y": "abc", "data": to_serialize(big_thing)}
await comm.write(msg)
```

Dask uses msgack generally for smaller parts of the message, but then serializes bigger things that are explicitly marked with `to_serialize`.

madsbk:
@mrocklin, it now uses Dask Comm objects and the performance is the same :)

mrocklin:
Ah, that's super interesting.  Thanks for doing the experiment Mads.

mrocklin:
Do we have a performance report of the merge with Dask lying around somewhere?  I would be interested in seeing it.

```python
from dask.distributed import performance_report

with performance_report(filename="dask-report.html"):
    result = result.persist()
    wait(result)
```

dask_cuda/benchmarks/local_cudf_merge.py
```diff
@@ -146,6 +132,38 @@ def run(args, write_profile=None):
...
+    t1 = clock()
+    wait(explicit_comms.cudf_merge(ddf1, ddf2).persist())
+    took = clock() - t1
+    return took
+
+
+def run(client, args, write_profile=None):
+    # Generate random Dask dataframes
+    ddf_base = get_random_ddf(
+        args.chunk_size, args.n_workers, args.frac_match, "build", args
```
rjzamora: Just a note that I started experimenting with a simple [`set_index` benchmark](https://github.com/rjzamora/dask-cuda/blob/set_index_bench/dask_cuda/benchmarks/local_cudf_set_index.py) (mostly a copy of the current local_cudf_merge.py).  I noticed that the underlying `sort_index` method in cudf was the bottleneck for a 1-to-1 mapping between workers (devices) and partitions.  I added the option to create multiple partitions per worker (`args.chunks_per_worker`) and noticed a dramatic drop in performance with `args.chunks_per_worker>1` for the same total dataframe size.  In this case communication, concatenation and `group_split_cudf` become much more time-intensive.

Perhaps it makes sense to add the same argument here (to capture a more-realistic dask use-case)?

dask_cuda/explicit_comms/dataframe_merge.py
```diff
@@ -0,0 +1,145 @@
...
+        elif len(df_parts) == 1:
+            return df_parts[0]
+        else:
+            return concat(df_parts)
+
+    df1 = df_concat(df1_parts)
+    df2 = df_concat(df2_parts)
+
+    return await distributed_join(
+        s["nworkers"], s["rank"], s["eps"], df1, df2, left_on, right_on
```
rjzamora: Here the `n_chunks` argument of `distributed_join` is hard-coded to be the number of workers.  Does this means the number of output partitions must be equal to the number of workers?  If so, don't we want the number of output partitions to depend on the number of input partitions (and maybe the type of merge)?

madsbk:
rerun tests

dask_cuda/explicit_comms/comms.py
```diff
@@ -0,0 +1,238 @@
...
+                    self.sessionId,
+                    coroutine,
+                    args,
+                    workers=[worker],
+                    pure=False,
+                )
+            )
+        return self.client.gather(ret)
+
+    def dataframe_operation(self, coroutine, df_list, extra_args=tuple()):
```
rjzamora: Perhaps there is a clean way to isolate all `dataframe`-specific code within a sub-director or dedicated file (like `explicit_comms/dataframe.py`)?  This way, we could also add `array`-specific code into something like `explicit_comms/array.py` in the future (if needed/desired).  Any thoughts on this?

mrocklin: > This way, we could also add array-specific code into something like explicit_comms/array.py in the future (if needed/desired). Any thoughts on this?

Jumping in here.  I'd love to see us avoid this kind of thing in the future if we can.  I think that there are some long term drawbacks to going outside of mainline like this.  I understand that folks are pushing hard for GTC numbers, but eventually it would be good to figure out what was slowing down normal merges and fix the underlying problem rather than work around it.

rjzamora: > but eventually it would be good to figure out what was slowing down normal merges and fix the underlying problem rather than work around it.

Agree mostly.  However, it does seem that the underlying problem (in many cases) is the scheduler's ability to handle the large/complex task graphs for complex/all-to-all shuffling.  The fact that dask relies on a centralized python-based scheduler makes it very difficult to match (or even approach) the performance of something like "explicit comms" for all-to-all communication.  

In order to enable users to mix workflows that include operations that dask already does well with those that are painfully slow (many-partition set_index/join/merge), I really do believe we need to enable users to inject isolated (optimized) tasks in place of much larger grpahs in specific/isolated cases where HPC-like performance is deemed necessary.  Ultimately, this is just a way to dramatically reduce the size of the task graph for specialized algorithms that dask traditionally struggles with.

mrocklin: Ah!  So is it the scheduling overhead that's weighing this down?  How many tasks per second are we processing?  I wonder if there are ways to speed that up.  In principle I would expect Dask to be able to process a few thousand tasks per second.  I would be somewhat surprised if we were pushing up against that limit on a single DGX (but I'm often surprised these days :) )

Also, if anyone has a performance report for a dask-cudf merge I'd be very curious to see it.

rjzamora: To clarify, scheduler *processing* doesn’t limit the performance of all-to-all shuffling.  Rather, the inefficiencies of things like unnecessary data movement seem to grow with the size of the graph. I think the problem has more to do with the fact that the scheduler needs to be flexible about *where* a large number of tasks will be executed. For optimal performance, we want to do things that require us to assume a (temporarily) static set of workers.  This assumption allows optimizations like aggregating data from multiple partitions when the data is persisted on the same worker (scaling algorithm complexity with the number of workers rather than the number of tasks - and increasing the typical message size).

  There is certainly room for improvement in the shuffle-based algorithms in mainline Dask.  My point is just that we would most likely need an unrealistically smart and performant scheduler to match *explicit comms*  performance when a large task graph can be replaced with a pre-defined execution pattern.

Note that I am still very open minded about this.

mrocklin: If we're moving data around inappropriately then I think that we should figure out why that's happening and resolve that problem.  I somewhat doubt that that is happening though.

I think that there are certainly cases where we'll want to step outside of Dask for communication and computation, particularly in cases that require very low latency communication, or very fine grained control.

My intuition is that merges and sorts aren't in this class of problems.  I would expect about a 2x improvement by hand-crafting things at most.  I understand that you all are seeing significantly greater than 2x speed improvement, so I suspect that there is a bug somewhere.  I would like to try to isolate where that bug is, so that hopefully mainline dask merges with ucx-py can operate at near full speed (along with all other dask + cudf + ucx communications).  

I would love to see a performance report using dask-cudf + ucx on a DGX.  I would be very happy to look through this with folks if anyone is interested.  I suspect that we could find some surprising things that might lead us down some paths to explore.

rjzamora: I think you are right @mrocklin - My earlier comments about the *need* for something like explicit comms for shuffling was mostly coming from my experience seeing a 40x speedup for `set_index`.  However, I just realized that this speedup is only ~2-3x when communication is performed with sockets/tcp.  Therefore, you are absolutely correct that the real problem is most-likely a bug in ucx for NVLink communication of small messages.  The explicit-comms rewrite avoids small messages by temporarily aggregating partitions for the shuffle.  Some of the remaining performance improvement I see for explicit comms is obtained by avoiding the need to assign a new “_partitions” column (and thus duplicate data).  I should be able to implement this last part in the mainline algorithm.  I will gather performance reports for merge/set_index on Monday and share :)   It would be great to get your thoughts.

mrocklin: >   I will gather performance reports for merge/set_index on Monday and share :) It would be great to get your thoughts.

Woot!  I'm really looking forward to seeing these :)

pentschev: > the real problem is most-likely a bug in ucx for NVLink communication of small messages

This is not really a bug, but a known behavior when the CUDA IPC cache is disabled. However, enabling it today with explicit comms indeed triggers a [bug in UCX](https://github.com/openucx/ucx/issues/3192), and so far this explicit comms PR is the only reproducible I've seen, but we are still investigating it and perhaps we would be able to see better performance once it's fixed.

rjzamora: Thanks for clarifying @pentschev!  I should have been more precise about the "bug."  Are you saying it is now **safe** to enable the CUDA IPC cache without explicit comms?  I was working under the (probably out-of-date) assumption that this would lead to OOM.

pentschev: > Are you saying it is now safe to enable the CUDA IPC cache without explicit comms?

Yes and no. One important thing to note is that CUDA IPC cache currently requires a memory pool (such as RMM's) to avoid OOM to control the entire application's memory management, as we are then safe on memory not being allocated/freed/reallocated, and we are currently working only with that scenario. However, enabling the pool for explicit comms has currently caused the bug I mentioned before (https://github.com/openucx/ucx/issues/3192), so probably it will crash currently, but not because it goes OOM. Finally, the case where no memory pool is used must still disable the cache to avoid OOMing.

Note that we've enabled the cache back in UCX-Py by default https://github.com/rapidsai/ucx-py/pull/397 again, as the main use cases we're focusing currently do use RMM pools. It's still possible to disable it by setting `UCX_CUDA_IPC_CACHE=n`.

madsbk: @mrocklin @rjzamora, I agree we should definitely try to move optimizations into the scheduler if possible, but I just want to mentioned that I have seen benchmarks that generates more than **200k tasks** and 1k partitions!

quasiben: i filed https://github.com/dask/distributed/issues/3413 in case folks want to dig deeper into questions around dask performance and joins

rjzamora: Thanks @quasiben!  I will move future discussion about if/how something like explicit comms should be used over there :)

madsbk:
rerun tests

dask_cuda/explicit_comms/dataframe_merge.py
```diff
@@ -0,0 +1,289 @@
...
+
+async def recv_df(ep):
+    ret = await ep.read()
+    if ret == "empty":
+        return None
+    else:
+        return ret[0]
+
+
+async def barrier(rank, eps):
```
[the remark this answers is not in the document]
madsbk: No not really, but it demonstrate how to use the communication model. At some point, we should provide some common primates, which will include a barrier.

madsbk:
I think this is ready to be merged

quasiben:
Thanks @rjzamora for the review

Issue #225, opened 2020-01-08
Spill over after libcudf++ merge is causing CUDA_ERROR_OUT_OF_MEMORY  issues


VibhuJawa:
### Spill over after libcudf++ merge is causing CUDA_ERROR_OUT_OF_MEMORY issues

After the `libcudf++` merge, the spill over mechanism might be failing. 

The current hypothesis is,  in dask-cuda it looks like if it spills to disk when moving back to the GPU it will allocate via numba instead of RMM. 

Relevant code lines are: 

From Dask-cuda:

https://github.com/rapidsai/dask-cuda/blob/db07453b130e7fea082279f8bc234f6227718b5b/dask_cuda/device_host_file.py#L90

From distributed (example of how it should be handled ): 

https://github.com/dask/distributed/blob/4a8a4f3bce378406e83a099e8a12fc9bc12ef25c/distributed/comm/ucx.py#L45-L63

CC: @jakirkham  @pentschev  @kkraus14  . 

Code to recreate the issue: 

https://gist.github.com/VibhuJawa/dbf2573954db86fb193b687022a20f46

Note:
I have not run the `cleaned up` code again on `exp01` but the  issue should still be there.  (Exp-01 was busy) 

Stack Trace
```
ERROR Call to cuMemAlloc results in CUDA_ERROR_OUT_OF_MEMORY
ERROR Call to cuMemAlloc results in CUDA_ERROR_OUT_OF_MEMORY
distributed.worker - ERROR - [2] Call to cuMemAlloc results in CUDA_ERROR_OUT_OF_MEMORY
Traceback (most recent call last):
  File "/raid/vjawa/conda_install/conda_env/envs/cudf_12_8_jan/lib/python3.7/site-packages/numba/cuda/cudadrv/driver.py", line 744, in _attempt_allocation
    allocator()
  File "/raid/vjawa/conda_install/conda_env/envs/cudf_12_8_jan/lib/python3.7/site-packages/numba/cuda/cudadrv/driver.py", line 759, in allocator
    driver.cuMemAlloc(byref(ptr), bytesize)
  File "/raid/vjawa/conda_install/conda_env/envs/cudf_12_8_jan/lib/python3.7/site-packages/numba/cuda/cudadrv/driver.py", line 294, in safe_cuda_api_call
    self._check_error(fname, retcode)
[...]

PR #227, opened 2020-01-15
Use RMM for serialization when available

pentschev:
Attempt to address https://github.com/rapidsai/dask-cuda/issues/225, not tested extensively yet.

### DIFF
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -18,6 +18,12 @@ from zict.common import ZictBase
 
 from .is_device_object import is_device_object
 
+try:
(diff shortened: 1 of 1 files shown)


pentschev:
@VibhuJawa could you try this out and see if it helps #225 and potentially other issues that you may be having?

Given my recent digging on larger memory footprint of libcudf++, I'll not be surprised if this PR doesn't help much.

pentschev:
Thanks @jakirkham for approving, since it seems to have resolved our existing issues, I'm going ahead and merging this as is for now.

jakirkham:
Thanks for working on this Peter! 😄

PR #235, opened 2020-02-04
Use RMM's `DeviceBuffer` directly

jakirkham:
This makes use of `to_device` and `copy_to_host` methods added to `DeviceBuffer` recently in PR ( https://github.com/rapidsai/rmm/pull/268 ) to directly create and copy data into RMM or serialize from RMM back to host.

### DIFF
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -93,7 +93,7 @@ def device_to_host(obj: object) -> DeviceSerialized:
     header, frames = serialize(obj, serializers=["cuda", "pickle"])
     is_cuda = [hasattr(f, "__cuda_array_interface__") for f in frames]
     frames = [
-        cuda.as_cuda_array(f).copy_to_host() if ic else f
+        copy_to_host(f) if ic else f
         for ic, f in zip(is_cuda, frames)
     ]
     return DeviceSerialized(header, frames, is_cuda)
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -19,7 +19,7 @@ from zict.common import ZictBase
 from .is_device_object import is_device_object
 
 try:
-    import rmm as cuda_memory_manager
+    from rmm import DeviceBuffer as cuda_memory_manager
 except ImportError:
     import numba.cuda as cuda_memory_manager
 
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -82,6 +82,13 @@ def _(header, frames):
     return DeviceSerialized(header["main-header"], parts, header["is-cuda"])
 
 
+def copy_to_host(ary):
+    if hasattr(ary, "copy_to_host"):
+        return ary.copy_to_host()
+    else:
+        return cuda.as_cuda_array(ary).copy_to_host()
+
+
 def device_to_host(obj: object) -> DeviceSerialized:
     header, frames = serialize(obj, serializers=["cuda", "pickle"])
     is_cuda = [hasattr(f, "__cuda_array_interface__") for f in frames]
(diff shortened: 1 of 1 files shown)


jakirkham:
@pentschev, what sorts of data does `host_to_device` normally see? Am assuming here it is more-or-less `bytes`-like, but can handle other edge cases if needed. So please let me know if there is anything I should be aware of/addressing here 🙂

pentschev:
@jakirkham thanks for working on this.

> what sorts of data does host_to_device normally see? Am assuming here it is more-or-less bytes-like, but can handle other edge cases if needed.

The `host_to_device` may not be `bytes`-like. I don't recall exactly how are all types transformed to, it depends on how each of the distributed CUDA serializers dump data, those that depend on this file https://github.com/dask/distributed/blob/master/distributed/protocol/cuda.py. So in general, I think we can expect all those types, probably not more than those today. Not sure if that answers your question though.

Also, it seems that tests are failing on deserialization.

jakirkham:
Thanks Peter! 😄

Alright will take a closer look at that. Thanks 🙂

Indeed. It looks like RMM 0.12.0 is being pulled in instead of the nightly. Not sure why that is. Do you know?

jakirkham:
rerun tests

jakirkham:
Ok looks like RMM nightlies are now getting picked up. The remaining errors are legitimate will take a look at fixing these and update after.

jakirkham:
Think I've fixed the CI issues. Could you please take another look, @pentschev? 🙂

dask_cuda/device_host_file.py
```diff
@@ -82,19 +82,26 @@ def _(header, frames):
...
+        copy_to_host(f) if ic else f
         for ic, f in zip(is_cuda, frames)
     ]
     return DeviceSerialized(header, frames, is_cuda)
 
 
 def host_to_device(s: DeviceSerialized) -> object:
     frames = [
-        cuda_memory_manager.to_device(f) if ic else f
+        cuda_memory_manager.to_device(f.ravel().view("u1")) if ic else f
```
pentschev: I don't think I understand, why is this necessary? Also, is it necessary for all cases or just specific ones?

jakirkham: Because RMM `DeviceBuffer`'s require [1-D contiguous data that is `uint8`]( https://github.com/rapidsai/rmm/blob/1002230ce362dce1842f5362b1814ce247889fc5/python/rmm/_lib/device_buffer.pyx#L131 ) since they don't handle size or type.

That said, Dask serializers are able to use the header to reconstruct the shape and type given a buffer like this. For example this can be seen in [CuPy]( https://github.com/dask/distributed/blob/9af811d8f9858c63b9586bcfb78ce2dec8f5d6b3/distributed/protocol/cupy.py#L55 ) and [Numba]( https://github.com/dask/distributed/blob/9af811d8f9858c63b9586bcfb78ce2dec8f5d6b3/distributed/protocol/numba.py#L34 ). cuDF itself always works with [1-D untyped `Buffer`]( https://github.com/rapidsai/cudf/blob/ba5aaeb44a1acdb00728db138e9c82014caa92a4/python/cudf/cudf/core/buffer.py#L10-L11 ) objects [under-the-hood]( https://github.com/rapidsai/cudf/blob/a0673be1eebc7a1746bf04655648cf97aaf6c47f/python/cudf/cudf/core/column/column.py#L36-L44 ). Here's [one example of cuDF's serialization]( https://github.com/rapidsai/cudf/blob/ba5aaeb44a1acdb00728db138e9c82014caa92a4/python/cudf/cudf/core/column/column.py#L877 ). There are others that handle various specific types, but all more-or-less work the same.

pentschev: Wow, you didn't need to provide all those examples, the first sentence would have sufficed. But thanks for the long description anyway!

pentschev:
LGTM, thanks @jakirkham for the PR!

dask_cuda/device_host_file.py
jakirkham: Haha ok. 😄 Yeah I figured that was just useful info to share after having done the homework on it earlier 😉

PR #236, opened 2020-02-06
Add RMM pool support directly from dask-cuda-worker/LocalCUDACluster

### LINKED ISSUE #233: Allow initialization of RMM upon creation of workers
To ensure a smoother operation, both from the user's perspective and internal Dask multithreading, we should make sure we can initialize RMM upon the creation of each worker.

cc @jakirkham @quasiben @kkraus14

pentschev:
Note for now it doesn't add support for managed memory as we can't use it with UCX at this time. If anybody thinks we absolutely need it right away, please speak up.

Fixes #233 

cc @beckernick @jakirkham

### DIFF
--- a/dask_cuda/dask_cuda_worker.py
+++ b/dask_cuda/dask_cuda_worker.py
@@ -27,6 +27,7 @@ from .initialize import initialize
 from .local_cuda_cluster import cuda_visible_devices
 from .utils import (
     CPUAffinity,
+    RMMPool,
     get_cpu_affinity,
     get_device_total_memory,
     get_n_gpus,
@@ -114,6 +115,12 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     "(i.e., allow full device memory usage).",
 )
 @click.option(
+    "--rmm-pool-size",
+    default=None,
+    help="Initialize each worker with an RMM pool of the given size. "
+    "This can be an integer (bytes) or string (like 5GB or 5000M)."
+)
+@click.option(
     "--reconnect/--no-reconnect",
     default=True,
     help="Reconnect to scheduler if disconnected",
@@ -188,6 +195,7 @@ def main(
     name,
     memory_limit,
     device_memory_limit,
+    rmm_pool_size,
     pid_file,
     resources,
     dashboard,
@@ -271,6 +279,9 @@ def main(
         else:
             host = get_ip_interface(interface)
 
+    if rmm_pool_size is not None:
+        rmm_pool_size = parse_bytes(rmm_pool_size)
+
     initialize(
         create_cuda_context=True,
         enable_tcp_over_ucx=enable_tcp_over_ucx,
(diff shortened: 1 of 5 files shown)


beckernick:
I think hard coding managed=False is likely fine for now

pentschev:
Codecov is complaining about the `ImportError` for `rmm` that we can't really test, so we can ignore that.

dask_cuda/local_cuda_cluster.py
```diff
@@ -166,6 +171,16 @@ def __init__(
...
         self.device_memory_limit = device_memory_limit
 
+        self.rmm_pool_size = rmm_pool_size
+        if rmm_pool_size is not None:
+            try:
+                import rmm
+            except ImportError:
+                raise ImportError(
+                    "RMM pool requested but module 'rmm' is not available"
+                )
```
jakirkham: Should make coverage happier. Not sure if we need this for all lines or just the last one.

```suggestion
                )  # pragma: no cover
```

pentschev: Does that also work for codecov? I'm not sure whether it uses coverage.py or the same syntax.

jakirkham: Yeah I think all Python coverage tools use `coverage.py`.

pentschev: Thanks @jakirkham , it seems to work indeed!

pentschev:
@jakirkham could you take another look at this? Should be good now.

dask_cuda/dask_cuda_worker.py
```diff
@@ -113,6 +114,12 @@
     "'auto', or zero for no memory management "
     "(i.e., allow full device memory usage).",
 )
+@click.option(
+    "--rmm-pool-size",
+    default=None,
+    help="Initialize each worker with an RMM pool of the given size. "
+    "This can be an integer (bytes) or string (like 5GB or 5000M)."
```
jakirkham: Should we say something about it being disabled by default?

dask_cuda/local_cuda_cluster.py
```diff
@@ -166,6 +171,16 @@ def __init__(
         self.host_memory_limit = memory_limit
         self.device_memory_limit = device_memory_limit
 
+        self.rmm_pool_size = rmm_pool_size
+        if rmm_pool_size is not None:
+            try:
+                import rmm
+            except ImportError:
+                raise ValueError(
+                    "RMM pool requested but module 'rmm' is not available"
```
jakirkham: Should we mention how to install it?

dask_cuda/utils.py
```diff
@@ -16,6 +16,19 @@ def setup(self, worker=None):
...
+class RMMPool:
+    def __init__(self, nbytes):
+        self.nbytes = nbytes
+
+    def setup(self, worker=None):
+        if self.nbytes is not None:
+            import rmm
+
+            rmm.reinitialize(
+                pool_allocator=True, managed_memory=False, initial_pool_size=self.nbytes
```
jakirkham: I think you already said this. So apologies if I'm rehashing things. dask-cuda sets up device `0` to be the device used by the worker right? By default this is what `rmm` will use. If not, we might need to specify it.

dask_cuda/dask_cuda_worker.py
pentschev: Done.

dask_cuda/local_cuda_cluster.py
pentschev: Done.

dask_cuda/utils.py
pentschev: That's right, the relevant GPU will always be GPU 0 for every dask-cuda worker, so we don't need to specify it.

pentschev:
Thanks @jakirkham and @madsbk for reviews.

jakirkham:
Thanks for adding this feature Peter! 😄

PR #252, opened 2020-03-11
Remote cuDF Merge

quasiben:
PR adds a remote version of the cuDF merge benchmark.  This is intended to be used when a cluster is launched separately from the client.  I've been using this can be used to help stress test new clusters

### DIFF
--- a/dask_cuda/explicit_comms/dataframe_merge.py
+++ b/dask_cuda/explicit_comms/dataframe_merge.py
@@ -93,7 +93,7 @@ def partition_by_hash(df, columns, n_chunks):
     # Hashing `columns` in `df` and assing it to the "_partitions" column
     df["_partitions"] = partitioning_index(df[columns], n_chunks)
     # Split `df` based on the hash values in the "_partitions" column
-    ret = shuffle_group(df, "_partitions", 0, n_chunks, n_chunks)
+    ret = shuffle_group(df, "_partitions", 0, n_chunks, n_chunks, ignore_index=True)
 
     # Let's remove the partition column and return the partitions
     del df["_partitions"]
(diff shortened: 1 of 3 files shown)


quasiben:
rerun tests

quasiben:
rerun tests

dask_cuda/explicit_comms/dataframe_merge.py
```diff
@@ -93,7 +93,7 @@ def partition_by_hash(df, columns, n_chunks):
     # Hashing `columns` in `df` and assing it to the "_partitions" column
     df["_partitions"] = partitioning_index(df[columns], n_chunks)
     # Split `df` based on the hash values in the "_partitions" column
-    ret = shuffle_group(df, "_partitions", 0, n_chunks, n_chunks)
+    ret = shuffle_group(df, "_partitions", 0, n_chunks, n_chunks, ignore_index=True)
```
quasiben: @rjzamora I needed to added the `ignore_index`.  I want to double check with you that this should be set to `True`

madsbk: Fixed the missing `ignore_index` in https://github.com/rapidsai/dask-cuda/pull/253

quasiben:
I would vote for shared code (utils.py) mostly because the cluster.scheduler calls change between remote/local but I'm open to suggestion.   Do you have a preference ?

madsbk:
> I would vote for shared code (utils.py) mostly because the cluster.scheduler calls change between remote/local but I'm open to suggestion. Do you have a preference ?

Agree, having a utils.py within benchmarks for shared code is a good approach.

quasiben:
closing in favor of https://github.com/rapidsai/dask-cuda/pull/293

pentschev:
@quasiben sorry I was completely unaware or had forgotten about this. Is there still anything valuable here that you think we could move into #293 ?

quasiben:
No, your implementation is a lot better and I let this drop

Issue #242, opened 2020-02-22
Using "dask" serialization protocol in spilling


jakirkham:
Currently `device_to_host` and `host_to_device` do a bit of work to coerce data into something easy to work with. Plus they do a bit of extra handling to move data to and from device manually.

However we have since done some work to enable the `"dask"` serialization protocol on CUDA objects ( https://github.com/dask/distributed/pull/3475 ) ( https://github.com/dask/distributed/pull/3478 ) ( https://github.com/dask/distributed/pull/3482 ) ( https://github.com/rapidsai/cudf/pull/4153 ), which not only serializes CUDA objects, but also perform the DtoH and HtoD transfers for us. It might be worth exploring how we can use this.

Though it's worth noting that these features require a yet to be released version of cuDF and the very newest Dask and Distributed releases (2.11.0). If we want to maintain backwards compatibility, we could try to use the `"dask"` serialization mode (if possible) and fallback to the current workflow otherwise.

jakirkham:
cc @madsbk (IIRC this came up in last week's stand-up. Sorry for the late pointer here)

PR #256, opened 2020-03-19
Use `"dask"` serialization to move to/from host

jakirkham:
Fixes https://github.com/rapidsai/dask-cuda/issues/242

As all `"cuda"` serializable objects are now also `"dask"` serializable, switch to just using `"dask"` serialization to perform device-to-host transfers to spill to host memory. Though continue falling back to `"pickle"` if nothing better can be found (after all that will be on host too). Similarly "unspilling" from host-to-device will happen naturally as part of the deserialization step. Should simplify what Dask-CUDA needs to keep track of/do.

### DIFF
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -12,16 +12,10 @@ from distributed.protocol import (
 from distributed.utils import nbytes
 from distributed.worker import weight
 
-from numba import cuda
 from zict import Buffer, File, Func
 from zict.common import ZictBase
 
 from .is_device_object import is_device_object
-
-try:
-    from rmm import DeviceBuffer as cuda_memory_manager
-except ImportError:
-    import numba.cuda as cuda_memory_manager
 
 
 class DeviceSerialized:
@@ -35,10 +29,9 @@ class DeviceSerialized:
         which are typically NumPy arrays
     """
 
-    def __init__(self, header, parts, is_cuda):
+    def __init__(self, header, parts):
         self.header = header
         self.parts = parts
-        self.is_cuda = is_cuda
 
     def __sizeof__(self):
         return sum(map(nbytes, self.parts))
@@ -54,7 +47,7 @@ def device_serialize(obj):
         headers.append(header)
         all_frames.extend(frames)
 
-    header = {"sub-headers": headers, "is-cuda": obj.is_cuda, "main-header": obj.header}
+    header = {"sub-headers": headers, "main-header": obj.header}
 
     return header, all_frames
 
@@ -67,30 +60,16 @@ def device_deserialize(header, frames):
         part = deserialize(sub_header, frames[start:stop])
         parts.append(part)
 
-    return DeviceSerialized(header["main-header"], parts, header["is-cuda"])
-
-
-def copy_to_host(ary):
-    if hasattr(ary, "copy_to_host"):
-        return ary.copy_to_host()
-    else:
-        return cuda.as_cuda_array(ary).copy_to_host()
+    return DeviceSerialized(header["main-header"], parts)
 
 
 def device_to_host(obj: object) -> DeviceSerialized:
-    header, frames = serialize(obj, serializers=["cuda", "pickle"])
-    is_cuda = [hasattr(f, "__cuda_array_interface__") for f in frames]
-    frames = [copy_to_host(f) if ic else f for ic, f in zip(is_cuda, frames)]
-    return DeviceSerialized(header, frames, is_cuda)
+    header, frames = serialize(obj, serializers=["dask", "pickle"])
+    return DeviceSerialized(header, frames)
 
 
 def host_to_device(s: DeviceSerialized) -> object:
-    frames = [
-        cuda_memory_manager.to_device(f.ravel().view("u1")) if ic else f
-        for ic, f in zip(s.is_cuda, s.parts)
-    ]
-
-    return deserialize(s.header, frames)
+    return deserialize(s.header, s.parts)
 
 
 class DeviceHostFile(ZictBase):
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -45,7 +45,7 @@ class DeviceSerialized:
 
 
 @dask_serialize.register(DeviceSerialized)
-def _(obj):
+def device_serialize(obj):
     headers = []
     all_frames = []
     for part in obj.parts:
@@ -60,7 +60,7 @@ def _(obj):
 
 
 @dask_deserialize.register(DeviceSerialized)
-def _(header, frames):
+def device_deserialize(header, frames):
     parts = []
     for sub_header in header["sub-headers"]:
         start, stop = sub_header.pop("frame-start-stop")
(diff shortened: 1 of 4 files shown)


dask_cuda/tests/test_device_host_file.py
```diff
@@ -157,9 +157,9 @@ def test_serialize_cupy_collection(collection, length, value):
     if length > 5:
         assert obj.header["serializer"] == "pickle"
     elif length > 0:
-        assert all([h["serializer"] == "cuda" for h in obj.header["sub-headers"]])
+        assert all([h["serializer"] == "dask" for h in obj.header["sub-headers"]])
```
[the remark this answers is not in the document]
jakirkham: Generally yes.

`"dask"` serialization already existed before in Dask. We've just now implemented it for CUDA objects. Implementation details with references in issue ( https://github.com/rapidsai/dask-cuda/issues/242 ).

jakirkham:
@pentschev, any more thoughts here? 🙂

pentschev:
Sorry @jakirkham I was just testing this now, and I think we have an issue with this PR, the reason it didn't fail is because the tests are marked to xfail as they would fail randomly in CI, but I could always (or almost always) have them pass in a DGX.

We have many `Failed to deserialize` errors, even the tests from this PR show these errors if you look at the [full log](https://gpuci.gpuopenanalytics.com/blue/rest/organizations/jenkins/pipelines/rapidsai/pipelines/gpuci/pipelines/dask-cuda/pipelines/prb/pipelines/dask-cuda-gpu-build/runs/828/log/?start=0) -- I can't point an exact line in the text file, but if you search for `Failed to deserialize` you will see those errors. I should note that these errors don't happen with branch-0.13.

I'm sorry that CI masked these errors, in fact what hinted me there could be an issue was the decrease in coverage as seen in https://codecov.io/gh/rapidsai/dask-cuda/pull/256/changes.

Could you check why this is happening @jakirkham ?

pentschev:
@jakirkham I should also mention that you can see the errors on a DGX too if you run `pytest -vs dask_cuda/tests/test_spill.py` with your branch.

jakirkham:
Sure I can take a look. Should these tests not be marked as `xfail` then?

jakirkham:
Ah I overlooked `is_device_object.py`. Sorry. I'll take a look at that. My guess is that module can be dropped with this change.

There may be more than that though.

pentschev:
> Sure I can take a look. Should these tests not be marked as xfail then?

I would say this is the goal, but we haven't been able to make the tests deterministic and there's https://github.com/rapidsai/dask-cuda/issues/79 for that, but it has been pretty hard to have CI always working.

> Ah I overlooked is_device_object.py. Sorry. I'll take a look at that. My guess is that module can be dropped with this change.

I don't think we can drop that, it is used to control _when_ dask-cuda spills. IOW, it uses them to keep track of LRU cache sizes and when to move data to host and back.

jakirkham:
Have improved this a bit, but there is still an issue. My guess is this has something to do with Dask splitting and compressing frames. Will look more into this next week.

<details>

```python
  File "/conda/envs/gdf/lib/python3.7/site-packages/distributed/worker.py", line 2164, in release_key
    if key in self.data and key not in self.dep_state:
  File "/conda/envs/gdf/lib/python3.7/_collections_abc.py", line 666, in __contains__
    self[key]
  File "/var/lib/jenkins/workspace/rapidsai/gpuci/dask-cuda/prb/dask-cuda-gpu-build/dask_cuda/device_host_file.py", line 122, in __getitem__
    return self.device_buffer[key]
  File "/conda/envs/gdf/lib/python3.7/site-packages/zict/buffer.py", line 78, in __getitem__
    return self.slow_to_fast(key)
  File "/conda/envs/gdf/lib/python3.7/site-packages/zict/buffer.py", line 65, in slow_to_fast
    value = self.slow[key]
  File "/conda/envs/gdf/lib/python3.7/site-packages/zict/func.py", line 38, in __getitem__
    return self.load(self.d[key])
  File "/var/lib/jenkins/workspace/rapidsai/gpuci/dask-cuda/prb/dask-cuda-gpu-build/dask_cuda/device_host_file.py", line 57, in host_to_device
    return deserialize(s.header, s.frames)
  File "/conda/envs/gdf/lib/python3.7/site-packages/distributed/protocol/serialize.py", line 269, in deserialize
    return loads(header, frames)
  File "/conda/envs/gdf/lib/python3.7/site-packages/distributed/protocol/serialize.py", line 51, in dask_loads
    return loads(header, frames)
  File "/conda/envs/gdf/lib/python3.7/site-packages/distributed/protocol/cupy.py", line 82, in dask_deserialize_cupy_ndarray
    frames = [dask_deserialize_cuda_buffer(header, frames)]
  File "/conda/envs/gdf/lib/python3.7/site-packages/distributed/protocol/rmm.py", line 37, in dask_deserialize_rmm_device_buffer
    (frame,) = frames
ValueError: too many values to unpack (expected 1)
```

</details>

jakirkham:
I think I've narrowed it down. It appears there may be a bug in Distributed. Basically it splits frames for compression. However it doesn't merge them back into their original structure after. Have submitted PR ( https://github.com/dask/distributed/pull/3639 ) to address this.

dask_cuda/device_host_file.py
```diff
@@ -54,43 +48,30 @@ def _(obj):
...
+    return DeviceSerialized(header["main-header"], parts)
 
 
 def device_to_host(obj: object) -> DeviceSerialized:
-    header, frames = serialize(obj, serializers=["cuda", "pickle"])
-    is_cuda = [hasattr(f, "__cuda_array_interface__") for f in frames]
-    frames = [copy_to_host(f) if ic else f for ic, f in zip(is_cuda, frames)]
-    return DeviceSerialized(header, frames, is_cuda)
+    header, frames = serialize(obj, serializers=["dask", "pickle"])
+    frames = [numpy.asarray(f) for f in frames]
```
jakirkham: Just a note here, the `frames` are `memoryviews`. Since we want to serialize `DeviceSerialized` later, which involves serializing these `frames`, and Dask doesn't know how to serialize a `memoryview`, it fails. To fix this, we construct a NumPy `ndarray` that merely views the underlying data in the `memoryview`. IOW this quite fast and zero-copy. Since Dask knows how to serialize NumPy `ndarray`s, this should proceed without issues.

jakirkham:
I've cut this back a bit to a state where it seems to work based on local testing. CI is a bit backed up, but should come through in a bit. Please let me know if anything else is needed.

dask_cuda/device_host_file.py
pentschev: Got it, this makes sense. Thanks for the description of how it works, it was really helpful to make it understand what's going on in here.

jakirkham:
Thanks for double checking Peter! 😄

I'll try adding back the other changes in another PR to see if we can simplify things further.

dask_cuda/device_host_file.py
jakirkham: Yeah ideally we wouldn't need this, but for now its a simple enough workaround 🙂

Issue #224, opened 2020-01-08
Dropping Numba sizeof code


jakirkham:
Currently [there is some code for measuring the size of Numba arrays]( https://github.com/rapidsai/dask-cuda/blob/db07453b130e7fea082279f8bc234f6227718b5b/dask_cuda/device_host_file.py#L21-L30 ). This was somewhat recently upstreamed into Dask with PR ( https://github.com/dask/dask/pull/5668 ). While we probably can't drop this code yet, figured it was worth noting that we can likely drop it in the future.

PR #257, opened 2020-03-19
Drop Numba `DeviceNDArray` code for `sizeof`

jakirkham:
Fixes https://github.com/rapidsai/dask-cuda/issues/224

As Dask 2.9.0+ has code for determining `sizeof` for Numba `DeviceNDArray`s, there is no need for us to carry around our own code for this. So go ahead and drop it.


jakirkham:
Thanks Peter! 😄

PR #259, opened 2020-03-23
DeviceHostFile: get stored value without moving it from host to device memory

madsbk:
This PR makes it possible to retrieve data from `DeviceHostFile` without moving it from host to device memory.

### DIFF
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -186,3 +186,12 @@ class DeviceHostFile(ZictBase):
     def __delitem__(self, key):
         self.device_keys.discard(key)
         del self.device_buffer[key]
+
+    def get_from_dev_or_host_buffer(self, key):
+        """Get stored value without moving it from host to device memory.
+        If the value isn't in device memory, `DeviceSerialized` is returned.
+        """
+        if key not in self.device_keys or key in self.device_func:
+            return self.__getitem__(key)
+        else:
+            return self.host_buffer[key]
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -163,7 +163,10 @@ class DeviceHostFile(ZictBase):
         self.fast = self.host_buffer.fast
 
     def __setitem__(self, key, value):
-        if is_device_object(value):
+        if isinstance(value, DeviceSerialized):
+            self.device_keys.add(key)
+            self.host_buffer[key] = value
+        elif is_device_object(value):
             self.device_keys.add(key)
             self.device_buffer[key] = value
         else:
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -195,3 +195,6 @@ class DeviceHostFile(ZictBase):
             return self.__getitem__(key)
         else:
             return self.host_buffer[key]
+
+    def __contains__(self, key):
+        return key in self.device_buffer or key in self.host_buffer


quasiben:
Would it be challenging to add a test for `get_from_dev_or_host_buffer` which asserts data has not been moved to the device ?

dask_cuda/device_host_file.py
```diff
@@ -186,3 +189,15 @@ def __iter__(self):
...
         del self.device_buffer[key]
+
+    def get_from_dev_or_host_buffer(self, key):
+        """Get stored value without moving it from host to device memory.
+        If the value isn't in device memory, `DeviceSerialized` is returned.
+        """
+        if key not in self.device_keys or key in self.device_func:
+            return self.__getitem__(key)
+        else:
+            return self.host_buffer[key]
```
pentschev: Not sure if I'm missing some condition here, but wouldn't it be simpler to write:
```python
        if key in self.host_buffer:
            return self.host_buffer[key]
        else:
            return self.__getitem__(key)
```

jakirkham:
Is PR ( https://github.com/dask/distributed/pull/3624 ) also relevant to this change?

dask_cuda/device_host_file.py
jakirkham: Could you please clarify what this is suppose to do that is different from `__getitem__`?

pentschev: `__getitem__` will get from the "top" of cache, meaning it will attempt to get from `device_buffer` and if it has been spilled it will retrieve from host/disk and store it back to the GPU and only then return the object.

jakirkham: Sorry maybe I'm still confused. Wouldn't this function more simply be just `return self.host_buffer[key]`?

pentschev: I think the purpose of the function is not to retrieve a spilled buffer, but also return the buffer in case it hasn't been spilled, thus the `__getitem__` call which would be equivalent to `return.device_buffer[key]`. However, I just realized that if the buffer has been spilled to host and ultimately to disk, `__getitem__` will make the buffer be retrieved to host and then to device, which is probably not what @madsbk wants. Maybe for the case where the buffer has been spilled to disk there should be a mechanism to retrieve it to the host and return that?

jakirkham: More generally I'm wondering if there is a way for us to avoid the addition of this method. It requires changes upstream, which probably won't be accepted ( https://github.com/dask/distributed/pull/3624 ). Trying to use `__getitem__` would be better if possible. Maybe using multiple Zict layers could be another way to get around this need.

pentschev: I see your point @jakirkham . I'm not sure whether there's much that can be done here without chaging upstream other than modifying the default behavior of this class. I guess generally we could opt to change what `__getitem__` does here, but we would have to check carefully whether there would be any drawbacks to that, for one changing it just to do what this function does would require an extra step to move it to device when required, which is probably undesirable.

I particularly think that the spilling functionality is very difficult to get right, it took us some good few weeks with the original implementation, so we should probably discuss upstream what alternatives we have and only then consider changing default behavior here.

pentschev:
@madsbk what's the status of this PR, do you think it's still going to be useful or you're working on a different solution?

pentschev:
Maybe the idea is that https://github.com/rapidsai/dask-cuda/pull/353 supersedes this PR?

madsbk:
Yes, #353 supersedes this.

PR #269, opened 2020-03-29
Fix behavior for `memory_limit=0`

pentschev:
This ensures behavior for `memory_limit=0` matches that of dask/distributed, meaning unlimited host memory, implying no spilling to disk.

### DIFF
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -111,7 +111,7 @@ class DeviceHostFile(ZictBase):
             serialize_bytelist, deserialize_bytes, File(self.disk_func_path)
         )
         if memory_limit == 0:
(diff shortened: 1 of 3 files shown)


pentschev:
cc @mt-jones @quasiben

pentschev:
> Not that it needs to be done in this PR (or perhaps not even at all), but would it make sense to do something similar with device_memory_limit?

That's a good question, I was thinking about that earlier today, thanks for raising that question as well. Short answer is: I'm not sure. I've raised https://github.com/rapidsai/dask-cuda/issues/270 to have a broader discussion with people that may have opinions on that matter. With that said, this won't be done in this PR, but we can open a new PR for that in the future.

quasiben:
Merging in

PR #272, opened 2020-03-30
Raise serialization errors when spilling

jakirkham:
The default behavior of `serialize` and thus `serialize_bytes` is to capture any serialization errors as messages. The result is the error gets raises on deserialization. However for spilling, this means that we fail to spill and raise that error after trying to restore that data. Instead go ahead and raise the error when trying to spill to avoid thinking content has spilled successfully (when that is not the case) and thus losing it.

### DIFF
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -1,3 +1,4 @@
+import functools
 import os
 
 import numpy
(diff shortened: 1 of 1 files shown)


jakirkham:
Should add this is what Dask does normally as well. So this just aligns us with that behavior.

pentschev:
LGTM. Just one question, do we need to do the same for deserialization?

jakirkham:
Nope. That will just raise the error if it fails.

PR #286, opened 2020-05-01
Remove deprecated DGX class

pentschev:


quasiben:
Yay for code removal!

You'll need to remove from `__init__.py` as well: https://github.com/rapidsai/dask-cuda/blob/afaec84dfd3a56f79f20b8a61619d2013eb6a634/dask_cuda/__init__.py#L2

pentschev:
Thanks for noting that @quasiben , I saw it before and still forgot. :)

dask_cuda/tests/test_device_host_file.py
```diff
@@ -154,9 +154,7 @@ def test_serialize_cupy_collection(collection, length, value):
     else:
         obj = device_to_host(collection((x,) * length))
 
-    if length > 5:
-        assert obj.header["serializer"] == "pickle"
-    elif length > 0:
+    if length > 0:
```
quasiben, line 157: can you remind me what's going on here ?  did we use to serialize with pickle for small objects ?

pentschev: Yes, but this magic number was now fixed in https://github.com/dask/distributed/pull/3689 , so everything is serialized using the default `"dask"` serializer.

pentschev:
Codecov is complaining about the NVTX annotation code, not sure why this is being triggered now but it's safe to ignore.

I also committed changes to the benchmarks that weren't meant for this PR but are useful to have nevertheless. I just did one last change to the benchmarks to add `--ucx-net-devices` and make that change complete then. It should be good to merge once tests pass.

dask_cuda/tests/test_device_host_file.py
quasiben: Thanks!

quasiben:
@rapidsai/admins do you know why I or @pentschev can't merge this PR ?  We aren't hitting and CI files

kkraus14:
Not sure why you couldn't, but I merged to unblock things in the short term.

pentschev:
Thanks @kkraus14 !

jakirkham:
Thought we bumped the Distributed requirement, but appears we missed that. Adding this in PR ( https://github.com/rapidsai/dask-cuda/pull/306 ). Since most users are likely on the latest version anyways and [other RAPIDS packages require Distributed 2.15.0+]( https://github.com/rapidsai/cudf/blob/b34b838625139bb14f4754db558112e8fd9293c8/conda/recipes/dask-cudf/meta.yaml#L31 ), I don't think this matters much, but probably still a good idea to update the requirement.

PR #290, opened 2020-05-05
Docs Setup

quasiben:
cc @pentschev


docs/source/worker.rst
```diff
@@ -0,0 +1,106 @@
...
+New configuration options::
+
+    --device-memory-limit
+    --rmm-pool-size
+    --enable-tcp-over-ucx / --disable-tcp-over-ucx
+    -enable-infiniband / --disable-infiniband
+    --enable-nvlink / --disable-nvlink
+    --net-devices
+
+Full details ``dask-cuda-worker`` options
```
pentschev, line 27: I think it's irrelevant to paste the full output of `dask-cuda-worker --help` here. We should definitely elaborate on those options that are specific to `dask-cuda-worker` (like the ones above this), but having a full output will also make it harder to keep in sync. My suggestion is to remove this and later elaborate on the important options.

quasiben: 👍

quasiben:
Thanks @pentschev .  I'll get this configured the rest of the way on RTD then we can keep iterating

quasiben:
@rapidsai/admins can you merge ?

conda/environments/builddocs_py37.yml
```diff
@@ -0,0 +1,14 @@
...
+dependencies:
+# required for building docs
+- sphinx
+- sphinx_rtd_theme
+- sphinxcontrib-websupport
+- nbsphinx
+- numpydoc
+- recommonmark
+- pandoc=<2.0.0
+- pip
```
mike-wendt, line 14: This is fine, but we have a package that is consistently updated for all of RAPIDS docs `rapids-doc-env` specified [here](https://github.com/rapidsai/integration/blob/branch-0.14/conda/recipes/rapids-doc-env/meta.yaml)

Instead of adding yet another env file with it's own dependencies, I'd like to see this use the env pkg above. This way we're not chasing so many version updates and changes across a dozen plus repos

quasiben: Thanks @mike-wendt -- didn't know about `rapids-doc-env` in #7d5bf84 I moved everything out and used `rapids-doc-env` as we don't do anything outside of standard sphinx things for now

quasiben:
@mike-wendt can you review again when you have time ?

jakirkham:
Thanks Ben! 😄

jakirkham:
RTD build is running out of memory creating the environment. Filed issue ( https://github.com/readthedocs/readthedocs.org/issues/7036 ) to follow up on that.

jakirkham:
Woot! It lives 😄

https://dask-cuda.readthedocs.io/

jakirkham:
Linking out to RTD from the README in PR ( https://github.com/rapidsai/dask-cuda/pull/292 ).

quasiben:
Thanks for pinging RTD. Woohoo!  Now onto lots more content!

Issue #289, opened 2020-05-03
Make `ucx_net_devices="auto"` more resilient


pentschev:
Currently `ucx_net_devices="auto"` will return the topologically closest devices regardless of their state. I don't think this is currently resilient in case the interfaces are not configured/connected properly.

It's probably a bit too difficult and even maybe out-of-scope for dask-cuda to ensure that devices are in a good state. However, we could do something to make that more resilient, some alternatives are:

1. Raising an exception if `"auto"` is used when `enable_infiniband=False`;
1. Check at least that the closest interface has an IP configured.

I think 1 should be done anyway, but 2 may limit usability given one could use IB without IBoIP.

PR #291, opened 2020-05-05
Raise ValueError when ucx_net_devices="auto" and IB is disabled

pentschev:
Fixes https://github.com/rapidsai/dask-cuda/issues/289

### DIFF
--- a/dask_cuda/dask_cuda_worker.py
+++ b/dask_cuda/dask_cuda_worker.py
@@ -192,9 +192,17 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     type=str,
     default=None,
     help="When None (default), 'UCX_NET_DEVICES' will be left to its default. "
-    "Otherwise, it must be a non-empty string with the interface name. Normally "
-    "used only with --enable-infiniband to specify the interface to be used by "
-    "the worker, such as 'mlx5_0:1' or 'ib0'.",
+    "Otherwise, it must be a non-empty string with the interface name, such as "
+    "such as 'eth0' or 'auto' to allow for automatically choosing the closest "
+    "interface based on the system's topology. Normally used only with "
+    "--enable-infiniband to specify the interface to be used by the worker, "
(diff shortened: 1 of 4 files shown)


dask_cuda/dask_cuda_worker.py
```diff
@@ -192,9 +192,17 @@
     type=str,
     default=None,
     help="When None (default), 'UCX_NET_DEVICES' will be left to its default. "
-    "Otherwise, it must be a non-empty string with the interface name. Normally "
-    "used only with --enable-infiniband to specify the interface to be used by "
-    "the worker, such as 'mlx5_0:1' or 'ib0'.",
+    "Otherwise, it must be a non-empty string with the interface name, such as "
```
quasiben, line 195: This a rather lengthy message for CLI output.  What do you think about redirecting them to documentation on how to properly use ?

pentschev: Do you mean linking to https://github.com/rapidsai/dask-cuda/blob/branch-0.14/dask_cuda/local_cuda_cluster.py#L98-L101, or do you have something else in mind that's immediately available?

quasiben: Nothing that's immediately available.  But we could add this to a configuration section on how to properly use IB within dask-cuda.  If you would prefer to keep this as I can submit a follow up PR and we can discuss there

pentschev: Certainly there's a lot to be documented, and this is one of those things. The idea here is just not to leave that uncovered for now. Personally, I would leave it as is just to provide a very descriptive information of what's supposed to be used -- not that I've seen anybody who actually read this anyway -- and address this together with the upcoming documentation improvements that we're working on.

pentschev: Feel free to disagree with me though and suggest a different solution. I'm considering this as temporary for now.

quasiben: Agreed with all that

PR #293, opened 2020-05-07
Multi-node benchmarks

pentschev:
This PR adds the capability to run the existing benchmarks in multi-node mode, using `SSHCluster`.

There are still a few issues that probably can't be resolved in dask-cuda:

1. We have to sleep for some time to allow workers to spin and connect to the scheduler;
1. We have to manually call `client.shutdown()`/`client.close()` to force the cluster to terminate;
1. We see some errors when the cluster is being terminated using the method from the previous item.

I haven't spent any time trying to fix the issues above yet, but this should give us at least some room to benchmarks things more easily.

### DIFF
--- a/dask_cuda/benchmarks/utils.py
+++ b/dask_cuda/benchmarks/utils.py
@@ -5,9 +5,7 @@ from dask_cuda.local_cuda_cluster import LocalCUDACluster
 
 
 def parse_benchmark_args(description="Generic dask-cuda Benchmark", args_list=[]):
-    parser = argparse.ArgumentParser(
-        description=description
-    )
+    parser = argparse.ArgumentParser(description=description)
     parser.add_argument(
         "-d", "--devs", default="0", type=str, help='GPU devices to use (default "0").'
     )
@@ -176,8 +174,6 @@ def setup_memory_pool(pool_size=None, disable_pool=False):
     import cupy
 
     rmm.reinitialize(
-        pool_allocator=not disable_pool,
-        devices=0,
-        initial_pool_size=pool_size,
+        pool_allocator=not disable_pool, devices=0, initial_pool_size=pool_size,
     )
     cupy.cuda.set_allocator(rmm.rmm_cupy_allocator)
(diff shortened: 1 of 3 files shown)


dask_cuda/benchmarks/utils.py
```diff
@@ -0,0 +1,179 @@
...
+
+def get_scheduler_workers(dask_scheduler=None):
+    return dask_scheduler.workers
+
+
+def setup_memory_pool(pool_size=None, disable_pool=False):
+    import rmm
+    import cupy
+
+    rmm.reinitialize(
```
quasiben, line 176: If the the `disable_pool` is True should we still set the cupy allocator ?

quasiben:
Thank you for taking this on @pentschev !  Would you be willing to provide a CLI example of using with dask-ssh ?

dask_cuda/benchmarks/utils.py
pentschev: TBH I just moved this function, I didn't change anything. I think it's fine (perhaps even preferred) to use the same allocator, even if we don't enable the pool it's fine to use RMM as the allocator for CuPy, that shouldn't cause problems.

pentschev:
> Thank you for taking this on @pentschev ! Would you be willing to provide a CLI example of using with dask-ssh ?

Sure, do you mean just as part of the command-line help, probably in https://github.com/rapidsai/dask-cuda/pull/293/files#diff-0ad9c1b7b91321ef0bc3a6a28ccb7402R85-R93 ?

quasiben:
Either in the help or here in the PR or in the new docs we have :)

pentschev:
> Either in the help or here in the PR or in the new docs we have :)

I'll for now add to the help and later move to docs. The docs will require some work and I don't intend to have everything in the same PR. :)

quasiben:
Sounds good to me.  If @madsbk has time to comment that would be great otherwise let's plan to merge in tomorrow ?

pentschev:
> Sounds good to me. If @madsbk has time to comment that would be great otherwise let's plan to merge in tomorrow ?

Sure, that's fine by me. I've added the example, it doesn't look very nice but we will soon improve that with proper docs.

PR #294, opened 2020-05-07
Add docs for UCX

pentschev:


docs/source/ucx.rst
```diff
@@ -0,0 +1,122 @@
+UCX Integration
+===============
+
+Communication can be a major bottleneck in distributed systems, and that's no different in Dask and Dask-CUDA. To address that, Dask-CUDA provides integration with `UCX-Py <https://ucx-py.readthedocs.io/>`_, a Python interface for the `UCX <https://www.openucx.org/>`_ communication framework. UCX is a low-level library that provides high-performance networking, supporting several transports, including NVLink and InfiniBand for systems that have such specialized hardware, as well as TCP for those that do not.
```
quasiben, line 4:
```suggestion
Communication can be a major bottleneck in distributed systems, and that's no different in Dask and Dask-CUDA. To address that, Dask-CUDA provides integration with `UCX-Py <https://ucx-py.readthedocs.io/>`_, a Python interface for the `UCX <https://www.openucx.org/>`_ communication framework. UCX is a low-level library that provides high-performance networking and supports several transports including: NVLink and InfiniBand, for systems that have such specialized hardware, as well as TCP for those that do not.
```

docs/source/ucx.rst
```diff
@@ -0,0 +1,122 @@
...
+
+
+Relevant Settings
+-----------------
+
+In this list below we shall see all the Dask configurations relevant for using with Dask-CUDA currently. Note that these options can also be used with mainline Dask/Distributed if you can't use Dask-CUDA for some reason. Note that many improvments for running GPU code is readily available in Dask-CUDA -- but not in mainline Dask/Distributed -- thus, we recommend using Dask-CUDA when possible.
```
quasiben, line 10:
```suggestion
In the list below we describe the relevant Dask configurations for using UCX with Dask-CUDA.  Note: these options can also be used with mainline Dask/Distributed outside of Dask-CUDA, however, we recommend using Dask-CUDA when possible.
```

pentschev: I want to keep the information about the advantages of using Dask-CUDA, removing that to just say "we recommend that" seems like it's just because we want to force that upon people for no reason, which is far from true. For that reason, I added another section describing benefits of using Dask-CUDA. I'm still looking at how to cross-reference sections, but will shortly push a change for that.

pentschev:
@quasiben I have everything in for this PR, please take another look when you can.

docs/source/specializations.rst
```diff
@@ -0,0 +1,24 @@
+Specializations for GPU Usage
+=============================
+
+It is known that main line Dask and Distributed packages can be used to leverage GPU computing, utilizing libraries such as cuDF, CuPy and Numba. So why use Dask-CUDA instead? This section aims to answer this question.
+
+Automatic Instantiation of One-Worker-Per-GPU
+---------------------------------------------
+
+Using the ``dask-cuda-worker`` or ``LocalCUDACluster`` will automatically launch one worker for each GPU available on the node where that is running, avoiding the need for users to select GPUs in their application and thus reducing code complexity.
```
quasiben, line 9:
```sugggestion
Using the ``dask-cuda-worker`` or ``LocalCUDACluster`` will automatically launch one worker for each GPU available on the node from where it was executed, avoiding the need for users to select GPUs in their application and thus reducing code complexity.
```

docs/source/specializations.rst
```diff
@@ -0,0 +1,24 @@
...
+
+CPU Affinity
+------------
+
+To improve performance, setting CPU affinity for each GPU is automatically done, preventing memory transfers from taking sub-optimal paths.
+
+Automatic Selection of InfiniBand Device
+----------------------------------------
+
+When UCX is activated, Dask-CUDA is also capable of selecting the topologically closest InfiniBand device to each GPU, thus ensuring optimal path and improving performance even further by using Remote Direct Memory Access (RDMA) when available. See the :doc:`UCX <ucx>` page for more details.
```
quasiben, line 24:
```
When InfiniBand is activated, Dask-CUDA is also capable of selecting the topologically closest InfiniBand device to each GPU, thus ensuring optimal path and improving performance even further by using GPU Remote Direct Memory Access (RDMA) when available. See the :doc:`UCX <ucx>` page for more details.
```

quasiben:
Made two small suggestions but otherwise I think looks good!

pentschev:
@quasiben suggestions applied, please merge at will if you have no further comments.

quasiben:
Thanks again @pentschev !

PR #295, opened 2020-05-08
Add `--runs` argument to CuPy benchmark

pentschev:


pentschev:
Thanks @quasiben for approving, merging now.

PR #301, opened 2020-05-29
Initialize dask-cuda-worker parent process' UCX configuration

pentschev:
This ensures the main process is also using the correct UCX transports for a homogeneous cluster.


quasiben:
Does this mean if I use LocalCUDACluster in a notebook this will appropriately configure UCX for the client as well ?

pentschev:
Hmm, that's a good question. I'll check that tomorrow and make changes if necessary.

pentschev:
We need to initialize `LocalCUDACluster` too, the latest commit addresses that. Thanks @quasiben for the catch.

PR #310, opened 2020-06-05
Fix tests related to latest RMM changes

pentschev:
Ensure we use Numba memory manager or xfail tests for which there's not yet a fix, may change with https://github.com/rapidsai/rmm/pull/391 .

cc @jakirkham

### DIFF
--- a/dask_cuda/tests/test_dask_cuda_worker.py
+++ b/dask_cuda/tests/test_dask_cuda_worker.py
@@ -55,7 +55,6 @@ def test_cuda_visible_devices_and_memory_limit(loop):  # noqa: F811
         del os.environ["CUDA_VISIBLE_DEVICES"]
 
 
-@pytest.mark.xfail(reason="rmm.get_info removed by https://github.com/rapidsai/rmm/pull/363")
 def test_rmm_pool(loop):  # noqa: F811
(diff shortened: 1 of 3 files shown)


jakirkham:
Thanks Peter! This seems reasonable. I think we need to run `black` on this maybe?

pentschev:
Thanks @jakirkham for the work on https://github.com/rapidsai/rmm/pull/391, I updated tests to use that capability so we don't have to xfail them anymore! :)

pentschev:
@rapidsai/ops-codeowners  it seems like CI is still picking rapidsai packages from 0.14, could you check?

raydouglass:
> @rapidsai/ops-codeowners it seems like CI is still picking rapidsai packages from 0.14, could you check?

Should be fixed now.

rerun tests

jakirkham:
Thanks Ray! 😄

jakirkham:
May need to wait for packages to be published with the fix in PR ( https://github.com/rapidsai/ucx-py/pull/544 ).

pentschev:
rerun tests

pentschev:
Thanks @raydouglass for fixing this, and @jakirkham for reviewing. I'm merging this.

jakirkham:
Thanks Peter! 😄

PR #315, opened 2020-06-10
Close CUDA context before running test_get_device_total_memory

### LINKED ISSUE #313: `test_get_device_total_memory` fails
The test currently fails due to changes in RMM, and only fails when another test involving RMM is executed prior to this. We have to check whether it's possible to really rollback the memory manager to Numba's default to make this pass, or find another way of checking GPU memory.

```python
======================================================== FAILURES ========================================================
______________________________________________ test_get_device_total_memory ______________________________________________

    def test_get_device_total_memory():
        # Ensure Numba is using its own memory manager, rather than RMM's
        cuda.set_memory_manager(cuda.cudadrv.driver.NumbaCUDAMemoryManager)

        for i in range(get_n_gpus()):
            with cuda.gpus[i]:
>               assert (
                    get_device_total_memory(i)
                    == cuda.current_context().get_memory_info()[1]
                )

dask_cuda/tests/test_utils.py:64:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../miniconda3/envs/rn-102-0.15.0b200608/lib/python3.7/site-packages/numba/cuda/cudadrv/driver.py:1023: in get_memory_info
    return self.memory_manager.get_memory_info()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <rmm.rmm.RMMNumbaManager object at 0x7fa4b52ed210>

    def get_memory_info(self):
>       raise NotImplementedError()
E       NotImplementedError

../../miniconda3/envs/rn-102-0.15.0b200608/lib/python3.7/site-packages/rmm/rmm.py:225: NotImplementedError
```

pentschev:
Fixes #313

### DIFF
--- a/dask_cuda/tests/test_utils.py
+++ b/dask_cuda/tests/test_utils.py
@@ -56,17 +56,11 @@ def test_cpu_affinity():
 
 
 def test_get_device_total_memory():
-    # Close CUDA context and Ensure Numba is using its own memory manager,
-    # rather than RMM's
-    cuda.close()
-    cuda.set_memory_manager(cuda.cudadrv.driver.NumbaCUDAMemoryManager)
-
     for i in range(get_n_gpus()):
         with cuda.gpus[i]:
-            assert (
-                get_device_total_memory(i)
-                == cuda.current_context().get_memory_info()[1]
-            )
+            total_mem = get_device_total_memory(i)
+            assert type(total_mem) is int
+            assert total_mem > 0
 
 
 @pytest.mark.parametrize("enable_tcp", [True, False])
--- a/dask_cuda/tests/test_utils.py
+++ b/dask_cuda/tests/test_utils.py
@@ -55,9 +55,10 @@ def test_cpu_affinity():
         assert list(os.sched_getaffinity(0)) == affinity
 
 
-@pytest.mark.xfail(reason="https://github.com/rapidsai/dask-cuda/issues/313")
 def test_get_device_total_memory():
-    # Ensure Numba is using its own memory manager, rather than RMM's
+    # Close CUDA context and Ensure Numba is using its own memory manager,
+    # rather than RMM's
+    cuda.close()
     cuda.set_memory_manager(cuda.cudadrv.driver.NumbaCUDAMemoryManager)
 
     for i in range(get_n_gpus()):


jakirkham:
I wonder if this will cause us issues in other tests that still expect Numba to be using RMM for allocations.

dask_cuda/tests/test_utils.py
```diff
@@ -55,9 +55,10 @@ def test_cpu_affinity():
         assert list(os.sched_getaffinity(0)) == affinity
 
 
-@pytest.mark.xfail(reason="https://github.com/rapidsai/dask-cuda/issues/313")
 def test_get_device_total_memory():
-    # Ensure Numba is using its own memory manager, rather than RMM's
+    # Close CUDA context and Ensure Numba is using its own memory manager,
+    # rather than RMM's
+    cuda.close()
     cuda.set_memory_manager(cuda.cudadrv.driver.NumbaCUDAMemoryManager)
```
jakirkham, line 62: Should we call `close` again and set back to RMM at the end of the test?

```python
cuda.set_memory_manager(rmm.RMMNumbaManager)
```

pentschev:
> I wonder if this will cause us issues in other tests that still expect Numba to be using RMM for allocations.

Arguably, a test should never implicitly rely on some other test's state, that would definitely be a flaw in the test design. If that happens, we should fix that test and ensure it is self-contained instead.

dask_cuda/tests/test_utils.py
gmarkall: Would it be better if the behaviour of `cuda.close()` also set the memory manager back to the default? (The default being either the Numba one, or the one specified in the environment variable `NUMBA_CUDA_MEMORY_MANAGER`)

Changing that behaviour obviously wouldn't help with this PR for right now, but I wonder if this would be the behaviour that you'd hope for / expect?

pentschev: As per my previous comment, I don't think we should undo changes in a test so that some other test will be in some expected state. Instead of resetting the memory manager to RMM, another tests that require that should do it if necessary.

pentschev: > Would it be better if the behaviour of cuda.close() also set the memory manager back to the default? (The default being either the Numba one, or the one specified in the environment variable NUMBA_CUDA_MEMORY_MANAGER)

I think that's not a bad idea, what would people generally expect from such a thing? If it's expected that `cuda.close()` leaves Numba in a "clean" state, then that change would be desirable. Otherwise, maybe there could be a `cuda.reset()` or something similar to roll Numba back to its initial state?

gmarkall: > I think that's not a bad idea, what would people generally expect from such a thing? If it's expected that `cuda.close()` leaves Numba in a "clean" state, then that change would be desirable. Otherwise, maybe there could be a `cuda.reset()` or something similar to roll Numba back to its initial state?

Thanks for the thoughts... I'm not quite sure what people would expect generally - `close()` is for destroying all contexts, so I would be inclined to not also reset the memory manager there. Despite what the documentation says, it turns out to be possible to change the memory manager at any time - however, existing contexts will still use their instances of memory managers, and the newly-set memory manager will only be instantiated for new contexts.

pentschev: I agree with that. I think at this time we don't have any strong argument in favor of such a change. Besides that, being explicit is probably better for sensitive functionalities like this one, so I think the way it is currently is perfectly fine.

jakirkham:
Well it's not really another test's state per-se. The issue is `import cudf` will [change the allocator]( https://github.com/rapidsai/cudf/blob/9544b7d7ce19a1a3db4bd998efd8c11590482989/python/cudf/cudf/__init__.py#L47 ). So we are merely restoring what happens when `import cudf` occurs.

pentschev:
> Well it's not really another test's state per-se. The issue is import cudf will change the allocator. So we are merely restoring what happens when import cudf occurs.

I see your point now. To avoid doing these potentially dangerous changes, I think it's best to just simplify that test not to rely on `get_memory_info` anymore. The latest commit reflects that. And thanks for detailing that concern! :)

pentschev:
Any other comments here @jakirkham ?

PR #316, opened 2020-06-15
Fix and improve DGX tests

pentschev:
Pass `UCX_NET_DEVICES` to scheduler and client processes, as RDMACM requires an IB device to work. Ensures that `UCX_NET_DEVICES` test also works for systems with many GPUs, such as a DGX-2. Add support for DGX A100 where already possible.

### DIFF
--- a/dask_cuda/tests/test_dgx.py
+++ b/dask_cuda/tests/test_dgx.py
@@ -228,6 +228,8 @@ def _test_dask_cuda_worker_ucx_net_devices(enable_rdmacm):
     sched_env = os.environ.copy()
     sched_env["DASK_UCX__INFINIBAND"] = "True"
     sched_env["DASK_UCX__TCP"] = "True"
+    sched_env["DASK_UCX__CUDA_COPY"] = "True"
+    sched_env["DASK_UCX__NET_DEVICES"] = openfabrics_devices[0]
 
     if enable_rdmacm:
         sched_env["DASK_UCX__RDMACM"] = "True"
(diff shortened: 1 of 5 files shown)


dask_cuda/tests/test_dgx.py
```diff
@@ -272,11 +277,16 @@ def _test_dask_cuda_worker_ucx_net_devices(enable_rdmacm):
...
                 start = time()
                 while True:
-                    if len(client.scheduler_info()["workers"]) == get_gpu_count():
+                    n_gpus = get_n_gpus()
+                    if len(client.scheduler_info()["workers"]) == n_gpus:
                         break
-                    else:
-                        assert time() - start < 10
-                        sleep(0.1)
+                    elif time() - start > 2 * n_gpus:
```
jakirkham, line 283: How did we determine this amount of time?

pentschev: Trial and error, this seems to be enough for a DGX-2, but the previous 10 seconds that was hardcoded wasn't.

pentschev:
All tests should now pass (or get skipped) on DGX-1, DGX-2 and DGX A100. Could you take another look when you have a chance @jakirkham ?

jakirkham:
Thanks Peter! This seems sensible. Though it appears we are seeing CI failures here. Do we know what might be happening?

pentschev:
> Thanks Peter! This seems sensible. Though it appears we are seeing CI failures here. Do we know what might be happening?

Not really, I can't reproduce this locally, I'll revert the change that is causing that and try again.

jakirkham:
Debugging with a dummy PR ( https://github.com/rapidsai/dask-cuda/pull/319 )

pentschev:
rerun tests

pentschev:
Thansk @jakirkham , merging before something else breaks! 😄

PR #320, opened 2020-06-19
Fix errors when TLS files are not specified

pentschev:
This PR should fix the issues that were caused by changes from https://github.com/dask/distributed/pull/3887 .


pentschev:
@rapidsai/admins the change mentioned in the description broke 0.14 release as well. I think we'll need to backport this to 0.14 and do a 0.14.1 release.

pentschev:
Thanks @jakirkham !

PR #323, opened 2020-06-23
Use RMM `pool-size` from config (if unspecified)

jakirkham:
Builds off of PR ( https://github.com/dask/distributed/pull/3515 )

If the user does not specify the RMM `pool-size`, try grabbing it from the `distributed.yaml` config. By default, that is disabled as well.

### DIFF
--- a/dask_cuda/local_cuda_cluster.py
+++ b/dask_cuda/local_cuda_cluster.py
@@ -156,7 +156,7 @@ class LocalCUDACluster(LocalCluster):
         )
         self.device_memory_limit = device_memory_limit
 
-        self.rmm_pool_size = rmm_pool_size
+        self.rmm_pool_size = rmm_pool_size or dask.config.get("rmm.pool-size")
         if rmm_pool_size is not None:
             try:
                 import rmm  # noqa F401


dask_cuda/local_cuda_cluster.py
```diff
@@ -156,7 +156,7 @@ def __init__(
         )
         self.device_memory_limit = device_memory_limit
 
-        self.rmm_pool_size = rmm_pool_size
+        self.rmm_pool_size = rmm_pool_size or dask.config.get("rmm.pool-size")
```
pentschev, line 159: Why is this necessary? `rmm.pool-size` was not intended to be used like this and was introduced specifically to allow adding an RMM pool to the scheduler when `protocol="ucx"`, as we weren't allowed to add it as a parameter to the scheduler's class/CLI. There's two critical side-effects in doing this, specifically when using `protocol="ucx"`:

1. `rmm.reinitialize` will be called a second time for workers in https://github.com/dask/distributed/blob/44b2358e33a0738c4c70ca96db4242636245e07d/distributed/comm/ucx.py#L103-L108 -- probably of lesser concern;
1. The scheduler (and maybe the client too, but I'm not sure) will also use an RMM pool of the same size as the workers, meaning that creating a large enough pool will cause an OOM.

quasiben: @jakirkham was there a workflow you had in mind ?

pentschev:
This PR has gone stale, @jakirkham could you check https://github.com/rapidsai/dask-cuda/pull/323#discussion_r444777550 and let us know if there's still some use case for when this is important or if we should close the PR?

jakirkham:
Let's revisit after the 0.15 release

pentschev:
@jakirkham is this still relevant? If so, could you check https://github.com/rapidsai/dask-cuda/pull/323#discussion_r444777550 ?

pentschev:
This has stalled and I believe this isn't necessary, I'm closing it but feel free to reopen if you would wish to discuss this further @jakirkham .

PR #324, opened 2020-06-24
Refactor worker into CUDAWorker class

jacobtomlinson:
As we are looking to move SSH Cluster to use spec (dask/distributed#3907) I've moved the `dask-cuda-worker` logic into a `CUDAWorker` class.

The CLI option still exists and now just handles signalling and starting the loop, it then creates a `CUDAWorker` and awaits it.

### DIFF
--- a/dask_cuda/dask_cuda_worker.py
+++ b/dask_cuda/dask_cuda_worker.py
@@ -20,7 +20,6 @@ from distributed.worker import parse_memory_limit
 
 import click
 from toolz import valmap
-from tornado import gen
 from tornado.ioloop import IOLoop, TimeoutError
 
 from .device_host_file import DeviceHostFile
--- a/dask_cuda/dask_cuda_worker.py
+++ b/dask_cuda/dask_cuda_worker.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
+import asyncio
 import atexit
 import logging
 import multiprocessing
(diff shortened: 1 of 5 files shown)


dask_cuda/dask_cuda_worker.py
```diff
@@ -244,160 +244,233 @@ def main(
...
+        dashboard_address,
+        local_directory,
+        scheduler_file,
+        interface,
+        death_timeout,
+        preload,
+        dashboard_prefix,
+        tls_ca_file,
+        tls_cert,
+        tls_key,
```
mrocklin, line 318: Most other classes accept a `security=` keyword that can be either a Security object or a dict of kwargs to expand into a Security object.  Thoughts on conforming to that API here?  (or changing the upstream API)

dask_cuda/dask_cuda_worker.py
```diff
@@ -244,160 +244,233 @@ def main(
...
+        enable_tcp_over_ucx,
+        enable_infiniband,
+        enable_nvlink,
+        enable_rdmacm,
+        net_devices,
+        **kwargs,
+    ):
```
mrocklin, line 325: It might be nice for some of these keywords to have default values.  This class might be hard to use from Python otherwise.

dask_cuda/dask_cuda_worker.py
jacobtomlinson: Yes good point. I've moved these options up into the CLI function to create the security object and am now passing that into the `CUDAWorker` class via the `security` kwarg.

dask_cuda/dask_cuda_worker.py
jacobtomlinson: Ah yeah, I've now set defaults in line with the CLI defaults.

jacobtomlinson:
@pentschev agree breaking them up would make sense.

How about moving the CLI code to either `cli.py` or `cli/worker.py` to be in line with upstream projects. And then renaming `dask_cuda_worker.py` to `cuda_worker.py`?

pentschev:
> How about moving the CLI code to either `cli.py` or `cli/worker.py` to be in line with upstream projects. And then renaming `dask_cuda_worker.py` to `cuda_worker.py`?

Creating a `cli` directory makes sense. I think to match more closely the naming scheme of distributed we could name them `dask_cuda/cli/dask_cuda_worker.py` (distributed uses `distributed/cli/dask_worker.py`) and `dask_cuda/cuda_worker.py` for the class (similar to `distributed/worker.py`).

jacobtomlinson:
@pentschev good idea. Done!

jacobtomlinson:
Whoops thanks @pentschev!

PR #329, opened 2020-06-30
add dask distributed gpu tests

quasiben:
resolves https://github.com/rapidsai/ucx-py/issues/523

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -101,4 +101,18 @@ else
     cd $WORKSPACE
     ls dask_cuda/tests/
     UCXPY_IFNAME=eth0 UCX_WARN_UNUSED_ENV_VARS=n UCX_MEMTYPE_CACHE=n py.test -vs --cache-clear --junitxml=${WORKSPACE}/junit-dask-cuda.xml --cov-config=.coveragerc --cov=dask_cuda --cov-report=xml:${WORKSPACE}/dask-cuda-coverage.xml --cov-report term dask_cuda/tests/
+
+    logger "Running dask.distributed GPU tests"
+    # Test downstream packages, which requires Python v3.7
+    if [ $(python -c "import sys; print(sys.version_info[1])") -ge "7" ]; then
+        logger "TEST OF DASK/UCX..."
+        py.test --cache-clear -vs `python -c "import distributed.protocol.tests.test_cupy as m;print(m.__file__)"`
+        py.test --cache-clear -vs `python -c "import distributed.protocol.tests.test_numba as m;print(m.__file__)"`
+        py.test --cache-clear -vs `python -c "import distributed.protocol.tests.test_rmm as m;print(m.__file__)"`
+        py.test --cache-clear -vs `python -c "import distributed.protocol.tests.test_collection_cuda as m;print(m.__file__)"`
+        py.test --cache-clear -vs `python -c "import distributed.comm.tests.test_ucx as m;print(m.__file__)"`
+        py.test --cache-clear -vs `python -c "import distributed.tests.test_nanny as m;print(m.__file__)"`
+        py.test --cache-clear -vs `python -c "import distributed.tests.test_gpu_metrics as m;print(m.__file__)"`
+        py.test --cache-clear -m "slow" -vs `python -c "import distributed.comm.tests.test_ucx as m;print(m.__file__)"`
+    fi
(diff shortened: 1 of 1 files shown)


ci/gpu/build.sh
```diff
@@ -101,4 +101,18 @@ else
...
+
+    logger "Running dask.distributed GPU tests"
+    # Test downstream packages, which requires Python v3.7
+    if [ $(python -c "import sys; print(sys.version_info[1])") -ge "7" ]; then
+        logger "TEST OF DASK/UCX..."
+        py.test --cache-clear -vs `python -c "import distributed.protocol.tests.test_cupy as m;print(m.__file__)"`
+        py.test --cache-clear -vs `python -c "import distributed.protocol.tests.test_numba as m;print(m.__file__)"`
+        py.test --cache-clear -vs `python -c "import distributed.protocol.tests.test_rmm as m;print(m.__file__)"`
+        py.test --cache-clear -vs `python -c "import distributed.protocol.tests.test_collection_cuda as m;print(m.__file__)"`
+        py.test --cache-clear -vs `python -c "import distributed.comm.tests.test_ucx as m;print(m.__file__)"`
```
pentschev, line 113: I think UCX tests should remain in UCX-Py, we will want to catch errors caused by changes in UCX-Py there at build time rather than here when something breaks.

ci/gpu/build.sh
```diff
@@ -101,4 +101,18 @@ else
...
+        py.test --cache-clear -vs `python -c "import distributed.tests.test_nanny as m;print(m.__file__)"`
+        py.test --cache-clear -vs `python -c "import distributed.tests.test_gpu_metrics as m;print(m.__file__)"`
+        py.test --cache-clear -m "slow" -vs `python -c "import distributed.comm.tests.test_ucx as m;print(m.__file__)"`
```
pentschev, line 116: Ditto.

ci/gpu/build.sh
quasiben: I'm waffling a bit.  What you are saying makes total sense, however, it would be nice to keep all the dask.distributed tests in one CI .  Perhaps mental overhead of tracking which tests are where is not too bad -- dask-ucx things in ucx-py and everything else in dask-cuda.

jakirkham: Would it be reasonable to run them in both places? How long do these tests take to run?

pentschev: I think the most important thing is to ensure we don't lose the ability to test UCX-Py potential breakages of distributed's UCX tests in PRs. I would normally prefer to reduce redundancy instead of testing it in both places, but I'm fine with keeping it in both places if people think this is important.

quasiben: Ok, let's keep the tests unchanged in ucx-py.  This PR replicates all the the test from ucx-py and adds `test_gpu_metrics`.  @pentschev does that work for you ?

pentschev: I don't really see the point in doing that, if all tests remain in UCX-Py, I would then argue that we don't need this PR here. I'm fine keeping UCX tests in both, but we should leave the distributed tests that don't depend on UCX in only one place.

quasiben: Oh, I thought you wanted to see tests moved here: https://github.com/rapidsai/ucx-py/issues/523 .  I am good with keeping everything in ucx-py

pentschev: To be fair, what I think makes more sense and what was my intent with the comment in https://github.com/rapidsai/ucx-py/issues/523:

1. Move distributed tests here into dask-cuda;
1. Keep `distributed.comm.tests.test_ucx` in UCX-Py.

UCX-Py is not in any way depending on distributed so I don't think that should be tested there, but dask-cuda does depend on distributed so it makes sense to have tests here. The UCX tests in distributed could be in UCX-Py (to ensure we don't break those with any PRs there), but it would be ok to be here as well, given dask-cuda is adding support to UCX as well.

quasiben: Done in https://github.com/rapidsai/dask-cuda/pull/329/commits/b971ef8d8ea3c34e23d3bbe4cc14aa6e05fc6278 .  I'll make an accompanying PR to ucx-py when this is merged in

pentschev: Thanks @quasiben . :)

PR #330, opened 2020-07-01
Make sure `del` clears keys from `host_buffer`

jakirkham:
As `host_buffer` can still contain the spilled value, make sure to remove that when `del` is called. Since it is possible the value was not spilled to the `host_buffer`, make sure a `KeyError` that could be thrown here is ignored.

### DIFF
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -1,3 +1,4 @@
+import contextlib
 import functools
 import os
 
@@ -146,3 +147,5 @@ class DeviceHostFile(ZictBase):
     def __delitem__(self, key):
         self.device_keys.discard(key)
         del self.device_buffer[key]
+        with contextlib.suppress(KeyError):
+            del self.host_buffer[key]
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -146,6 +146,7 @@ class DeviceHostFile(ZictBase):
 
     def __delitem__(self, key):
         self.device_keys.discard(key)
-        del self.device_buffer[key]
+        with contextlib.suppress(KeyError):
+            del self.device_buffer[key]
         with contextlib.suppress(KeyError):
             del self.host_buffer[key]


quasiben:
If the value was not spilled we shouldn't we still see the key in the `device_buffer` ?  I thought zict was always keeping track of all keys.

jakirkham:
FWICT we are [setting values in `host_buffer`]( https://github.com/rapidsai/dask-cuda/blob/e93ca001a2e8e2679c18824e8c5f8efb3d601c0e/dask_cuda/device_host_file.py#L129-L130 ) and [getting values from `host_buffer`]( https://github.com/rapidsai/dask-cuda/blob/e93ca001a2e8e2679c18824e8c5f8efb3d601c0e/dask_cuda/device_host_file.py#L135-L136 ). IOW we are managing the content of `host_buffer`. Shouldn't we be doing that with deletion as well?

dask_cuda/device_host_file.py
```diff
@@ -145,4 +146,7 @@ def __iter__(self):
 
     def __delitem__(self, key):
         self.device_keys.discard(key)
-        del self.device_buffer[key]
+        with contextlib.suppress(KeyError):
+            del self.device_buffer[key]
+        with contextlib.suppress(KeyError):
+            del self.host_buffer[key]
```
pentschev, line 152: This isn't necessary, here `device_buffer` is `fast` and `host_buffer` -- being the dict in `host_func` -- is `slow`. `zict.Buffer` will delete keys from both `fast` and `slow`: https://github.com/dask/zict/blob/3a9ea886583ba9920be8ab8ecec4adde23401a1a/zict/buffer.py#L93-L99

pentschev:
> FWICT we are setting values in host_buffer and getting values from host_buffer. IOW we are managing the content of host_buffer. Shouldn't we be doing that with deletion as well?

As per my comment above, zict is doing the management itself, so it would be redundant to do it again in `DeviceHostFile`.

dask_cuda/device_host_file.py
jakirkham: So I'm curious about this point. Why do we need to specify [`__setitem__`]( https://github.com/rapidsai/dask-cuda/blob/e93ca001a2e8e2679c18824e8c5f8efb3d601c0e/dask_cuda/device_host_file.py#L125 ) and [`__getitem__`]( https://github.com/rapidsai/dask-cuda/blob/e93ca001a2e8e2679c18824e8c5f8efb3d601c0e/dask_cuda/device_host_file.py#L132 ) and handle `host_buffer` in those cases and not here?

jakirkham: Additionally do we actually need `__delitem__` if this is being handled for us?

pentschev: > So I'm curious about this point. Why do we need to specify [`__setitem__`](https://github.com/rapidsai/dask-cuda/blob/e93ca001a2e8e2679c18824e8c5f8efb3d601c0e/dask_cuda/device_host_file.py#L125) and [`__getitem__`](https://github.com/rapidsai/dask-cuda/blob/e93ca001a2e8e2679c18824e8c5f8efb3d601c0e/dask_cuda/device_host_file.py#L132) and handle `host_buffer` in those cases and not here?

We need it because zict doesn't know about device or host buffers based on their type. By doing this we make sure that buffers will go directly to the appropriate dict (`device_buffer` if it's a CUDA-based buffer, `host_buffer` if CPU). When we define a maximum size for each dict (based on `device_memory_limit`/`memory_limit`) zict is able to spill once the actual buffer gets to that limit, otherwise we wouldn't be able to separate and spill according to whether device or host memory buffers are at the limit. Think for example of when we do `__getitem__`, if we just get a host item and use the `device_buffer` it would work but zict would try to "unspill" it from `host_buffer` back to `device_buffer`, which we don't want to do.

> Additionally do we actually need __delitem__ if this is being handled for us?

That's a good question, I think we need it only because of `self.device_keys.discard(key)`, otherwise it would probably be fine to drop `__delitem__` entirely.

pentschev:
@jakirkham as per https://github.com/rapidsai/dask-cuda/pull/330#discussion_r448199014 and https://github.com/rapidsai/dask-cuda/pull/330#issuecomment-652237685, I don't think this PR is necessary. I'm tentatively closing it since it has gone stale. Please reopen if you still have any concerns.

PR #331, opened 2020-07-01
Handle host buffers in iter and len

jakirkham:
It seems we handle host buffers in other cases, but that doesn't seem to be the case in `__iter__` and `__len__`. So adjust `__iter__` and `__len__` to handle both device and host buffers.

### DIFF
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -1,4 +1,5 @@
 import functools
+import itertools
 import os
 
 import dask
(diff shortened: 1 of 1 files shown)


pentschev:
@jakirkham I don't think this PR is necessary. Note that `self.device_buffer` _contains_ `self.host_buffer`. The reason we have special casing for `self.device_buffer` and `self.host_buffer` is mentioned in https://github.com/rapidsai/dask-cuda/pull/330#discussion_r448199014 . If there's a case for which the current behavior without this PR is wrong, could you add a test for that so that we can verify this PR really fixes it? Otherwise if you think that this may not be necessary, should we close it?

jakirkham:
Let's revisit after the 0.15 release

pentschev:
@jakirkham is this still relevant? Can you check https://github.com/rapidsai/dask-cuda/pull/331#issuecomment-675574202 ?

pentschev:
This has stalled and I believe https://github.com/rapidsai/dask-cuda/pull/331#issuecomment-675574202 is correct, I'm closing it but feel free to reopen if you still think we have a bug here @jakirkham .

PR #338, opened 2020-07-07
CPU Bench

quasiben:
This is generalizes the merge benchmark to optionally test merges on CPUs (pandas)

### DIFF
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -2,15 +2,12 @@ import math
 from collections import defaultdict
 from time import perf_counter as clock
 
-import cupy
 import numpy
 
 from dask.base import tokenize
 from dask.dataframe.core import new_dd_object
 from dask.distributed import Client, performance_report, wait
 from dask.utils import format_bytes, format_time, parse_bytes
-
-import cudf
 
 from dask_cuda import explicit_comms
 from dask_cuda.benchmarks.utils import (
@@ -24,17 +21,17 @@ from dask_cuda.benchmarks.utils import (
 # <https://gist.github.com/rjzamora/0ffc35c19b5180ab04bbf7c793c45955>
 
 
-def generate_chunk(i_chunk, local_size, num_chunks, chunk_type, frac_match,
-        gpu):
+def generate_chunk(i_chunk, local_size, num_chunks, chunk_type, frac_match, gpu):
     # Setting a seed that triggers max amount of comm in the two-GPU case.
     if gpu:
         import cupy as xp
+
         import cudf as xdf
     else:
         import numpy as xp
         import pandas as xdf
 
-    xp.random.seed(2**32 - 1)
+    xp.random.seed(2 ** 32 - 1)
 
     chunk_type = chunk_type or "build"
     frac_match = frac_match or 1.0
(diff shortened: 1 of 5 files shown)


dask_cuda/benchmarks/local_cudf_merge.py
```diff
@@ -227,6 +238,7 @@ def main(args):
     print("Merge benchmark")
     print("-------------------------------")
     print(f"backend        | {args.backend}")
+    print(f"backend        | {args.type}")
```
pentschev, line 241:
```suggestion
    print(f"merge type     | {args.type}")
```

dask_cuda/benchmarks/utils.py
```diff
@@ -176,8 +183,11 @@ def get_scheduler_workers(dask_scheduler=None):
 
 
 def setup_memory_pool(pool_size=None, disable_pool=False):
-    import rmm
-    import cupy
+    try:
+        import rmm
+        import cupy
```
pentschev, line 188: It doesn't make sense to have `import cupy` inside the `try` block for this test, for any GPU test CuPy is required.

dask_cuda/benchmarks/utils.py
```diff
@@ -176,8 +183,11 @@ def get_scheduler_workers(dask_scheduler=None):
...
+    except ImportError:
+        return None
```
pentschev, line 190: I think it doesn't make sense to just `return None` if RMM isn't installed, I hate this kind of silent fallbacks because it makes it very hard to debug. IMO, we should error out if RMM can't be imported and `pool_size != None or disable_pool is not False`.

dask_cuda/explicit_comms/dataframe_merge.py
```diff
@@ -3,7 +3,10 @@
 from dask.dataframe.shuffle import partitioning_index, shuffle_group
 from distributed.protocol import to_serialize
 
-import cudf
+try:
+    import cudf
+except ImportError:
+    pass
```
pentschev, line 9: What happens when GPU type is requested but cuDF isn't installed?

dask_cuda/benchmarks/utils.py
quasiben: That's true, however, when loading utils in `local_cudf_merge.py` this will fail unless cupy is installed or in the try block

dask_cuda/benchmarks/utils.py
quasiben: That's a good point.  My goal was minimal changes in the merge code .  Would you be ok if I added in some more gpu/cpu conditionals  ?

quasiben:
@pentschev when you have time could you look this over again ?

jakirkham:
This is great! Thanks for working on this Ben and Peter for reviewing. 😄

Just gave this a try on a DGX-1 with CPU and GPU cases. Have included the results below. This is using the very latest RAPIDS 0.15 nightlies.

CPU:

```
$ python dask_cuda/benchmarks/local_cudf_merge.py -t cpu
Merge benchmark
-------------------------------
backend        | dask
merge type     | cpu
rows-per-chunk | 1000000
protocol       | tcp
device(s)      | 0
rmm-pool       | True
frac-match     | 0.3
data-processed | 32.00 MB
===============================
Wall-clock     | Throughput
-------------------------------
467.18 ms      | 68.50 MB/s
483.35 ms      | 66.20 MB/s
484.23 ms      | 66.08 MB/s
===============================
-------------------------------
```

GPU:

```
$ python dask_cuda/benchmarks/local_cudf_merge.py -t gpu
Merge benchmark
-------------------------------
backend        | dask
merge type     | gpu
rows-per-chunk | 1000000
protocol       | tcp
device(s)      | 0
rmm-pool       | True
frac-match     | 0.3
data-processed | 32.00 MB
===============================
Wall-clock     | Throughput
-------------------------------
10.38 ms       | 3.08 GB/s
8.86 ms        | 3.61 GB/s
12.14 ms       | 2.64 GB/s
===============================
-------------------------------
```

PR #343, opened 2020-07-09
Add support for RMM managed memory

### LINKED ISSUE #341: [FEA] Support RMM managed memory during cluster startup via the command line
Today, I can set up a Dask cluster via the CLI that includes a configurable RMM pool size via `--rmm-pool-size`. This ultimately calls down to the RMMPool class to initialize RMM. This class explicitly sets up an RMM pool with `managed_memory=False`.

https://github.com/rapidsai/dask-cuda/blob/b24567d0cc3be7570f19b3bd5fe8d31d4591b623/dask_cuda/utils.py#L31-L41

It would be very nice if we could also configure RMM to use managed memory via the command line.

pentschev:
Add support for RMM managed memory via `dask-cuda-worker`/`LocalCUDACluster` parameters.

Fixes #341 .

### DIFF
--- a/dask_cuda/cli/dask_cuda_worker.py
+++ b/dask_cuda/cli/dask_cuda_worker.py
@@ -100,6 +100,14 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     "an integer (bytes) or string (like 5GB or 5000M).",
 )
 @click.option(
+    "--rmm-managed-memory/--no-rmm-managed-memory",
+    default=False,
+    help="If enabled, initialize each worker with RMM and set it to "
+    "use managed memory. If disabled, RMM may still be used if "
+    "--rmm-pool-size is specified, but in that case with default "
+    "(non-managed) memory type.",
+)
+@click.option(
     "--reconnect/--no-reconnect",
     default=True,
     help="Reconnect to scheduler if disconnected",
@@ -188,6 +196,7 @@ def main(
     memory_limit,
     device_memory_limit,
     rmm_pool_size,
+    rmm_managed_memory,
     pid_file,
     resources,
     dashboard,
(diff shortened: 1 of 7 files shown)


pentschev:
cc @beckernick

pentschev:
This is the first time I see the test fail in https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/rapidsai%2Fgpuci%2Fdask-cuda%2Fprb%2Fdask-cuda-gpu-build/detail/dask-cuda-gpu-build/1273/pipeline#log-1522 . I don't think it's deterministic and is also unrelated to this PR. We better keep an eye on it in any case, it seems to be due to UCX.

pentschev:
rerun tests

pentschev:
rerun tests

pentschev:
rerun tests

pentschev:
rerun tests

pentschev:
This finally passes, 3 different issues had to be fixed, all unrelated to this PR:

1. Something changed that caused issues with missing `cuda_copy` with `LocalCluster` (note it's not `LocalCUDACluster`), fixed by https://github.com/rapidsai/dask-cuda/pull/343/commits/42594c8ecac2a9b389d3f17fea1a2c86fc4ab16c ;
1. Changes due to pickle5 serialization, fixed in: https://github.com/rapidsai/cudf/pull/5684 ;
1. A new `reuse-endpoints` variable was introduced in https://github.com/dask/distributed/pull/3953 which needs to be added to dask-cuda, fixed in https://github.com/rapidsai/dask-cuda/pull/343/commits/d7de57b36f81b58f93095b86a192a52a82192b35 .

I'm merging this before anything else breaks.

quasiben:
Thank you @pentschev for working through these annoying issues

beckernick:
➕ 1️⃣  thanks a lot Peter

PR #350, opened 2020-07-24
Fix isort and black line length

jakirkham:
Appears that isort and black (via flake8) had different line lengths, which clashed with each other resulting in the lint always failing. This makes sure isort uses the same line length that black (and flake8) use. Also reruns isort, black, and flake8 to make sure everything is copacetic.

### DIFF
--- a/ci/checks/style.sh
+++ b/ci/checks/style.sh
@@ -12,7 +12,7 @@ PATH=/conda/bin:$PATH
 source activate gdf
 
 # Run isort and get results/return code
-ISORT=`isort --recursive --check-only .`
+ISORT=`isort --check-only .`
 ISORT_RETVAL=$?
 
 # Run black and get results/return code
--- a/dask_cuda/explicit_comms/comms.py
+++ b/dask_cuda/explicit_comms/comms.py
@@ -5,11 +5,7 @@ import uuid
 
 import distributed.comm
 from distributed import default_client, get_worker
-from distributed.comm.addressing import (
-    parse_address,
-    parse_host_port,
-    unparse_address,
-)
+from distributed.comm.addressing import parse_address, parse_host_port, unparse_address
 
 from . import utils
 
--- a/dask_cuda/tests/test_device_host_file.py
+++ b/dask_cuda/tests/test_device_host_file.py
@@ -8,11 +8,7 @@ import dask
 from dask import array as da
 from distributed.protocol import deserialize_bytes, serialize_bytelist
 
-from dask_cuda.device_host_file import (
-    DeviceHostFile,
-    device_to_host,
-    host_to_device,
-)
+from dask_cuda.device_host_file import DeviceHostFile, device_to_host, host_to_device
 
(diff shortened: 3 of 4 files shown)


jakirkham:
This really weird. I'm not seeing this issue locally...

```

>>>> Detected in repo script, using 'ci/checks/style.sh' for build...
All done! ✨ 🍰 ✨
32 files would be left unchanged.


>>>> PASSED: isort style check




>>>> PASSED: black style check




>>>> FAILED: flake8 style check; begin output





>>>> FAILED: flake8 style check; end output
```

https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/rapidsai%2Fgpuci%2Fdask-cuda%2Fprb%2Fdask-cuda-style/detail/dask-cuda-style/458/pipeline

pentschev:
Changes here seem to be differing from other RAPIDS projects, e.g.:

- https://github.com/rapidsai/cudf/blob/6857a43cfc59ca68442eec31af5f841feea91422/python/cudf/setup.cfg#L16
- https://github.com/rapidsai/rmm/blob/ba4436b17d4a0c94469db2ca364c6f961d263889/python/setup.cfg#L16
- https://github.com/rapidsai/ucx-py/blob/3994037b5bb44458bc85c3e9505138b5d3f6a3f3/setup.cfg#L22

Is that really what we want?

jakirkham:
Would suggest running `isort` and `black` locally. What I've seen is they both have a different way of handling this line. Depending on which one is run, this either becomes one line or is broken by a `\`. Not sure why they handle this differently.

https://github.com/rapidsai/dask-cuda/blob/975461c68878780e18086736d4b37a37f4109bb4/dask_cuda/local_cuda_cluster.py#L211-L212

jakirkham:
rerun tests

jakirkham:
Just to be clear, I tried both options (setting to `79` and `88`). The problem I run into is `black` seems to ignore the configuration and still unwraps that line regardless of the line length specified. It may be a bug in how `black` works. Choosing `88` seems to at least get `isort` and `black` to cooperate.

pentschev:
TBH, it's still unclear to me what should be done here or why that's necessary. At least in its current state, this PR seems to be following a diverging path to that of other libraries, as per https://github.com/rapidsai/dask-cuda/pull/350#issuecomment-663865867 .

pentschev:
Do we still need this @jakirkham ? It seems like we rely on specific versions of formatting packages, it's not really clear to me what's the intent of this PR.

pentschev:
It seems to me that formatting works properly provided we locally use the same versions as CI does (see https://github.com/rapidsai/integration/blob/branch-0.17/conda/recipes/versions.yaml for reference). Therefore, I'm tentatively closing this, please feel free to reopen if there's still something that needs to be done here @jakirkham .

jakirkham:
Mads fixed this issue in PR ( https://github.com/rapidsai/dask-cuda/pull/477 ).

PR #358, opened 2020-08-12
[NO MRG] Test CI on 0.16

jakirkham:
Just creating this PR to start CI.


jakirkham:
The failure here seems transient. So won't worry about it for now unless it starts coming up more frequently.

PR #363, opened 2020-08-12
fix/rmm-init

quasiben:
Fixes https://github.com/rapidsai/dask-cuda/issues/364

PR avoids cuda driver initialization from rmm based on https://github.com/rapidsai/rmm/pull/490

Still in the middle of building an env to test the RMM PR

cc @kkraus14

### DIFF
--- a/dask_cuda/benchmarks/utils.py
+++ b/dask_cuda/benchmarks/utils.py
@@ -186,7 +186,6 @@ def get_scheduler_workers(dask_scheduler=None):
 def setup_memory_pool(pool_size=None, disable_pool=False):
     import cupy
 
-    print("FOODSFSDF")
     os.environ['RMM_NO_INITIALIZE'] = 'True'
     import rmm
 
--- a/dask_cuda/benchmarks/utils.py
+++ b/dask_cuda/benchmarks/utils.py
@@ -186,6 +186,8 @@ def get_scheduler_workers(dask_scheduler=None):
 def setup_memory_pool(pool_size=None, disable_pool=False):
     import cupy
 
+    print("FOODSFSDF")
+    os.environ['RMM_NO_INITIALIZE'] = 'True'
     import rmm
 
     rmm.reinitialize(
--- a/dask_cuda/benchmarks/utils.py
+++ b/dask_cuda/benchmarks/utils.py
@@ -1,4 +1,5 @@
 import argparse
+import os
 
 from dask.distributed import SSHCluster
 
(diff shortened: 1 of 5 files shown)


pentschev:
I already had it built, tested this PR with `dask-cuda-worker tcp://10.33.227.163:8786 --rmm-pool-size=20GB` and that worked. Would be good to ensure it works with `LocalCUDACluster` as well @quasiben . Thanks for doing this!

dask_cuda/benchmarks/utils.py
```diff
@@ -186,6 +186,8 @@ def get_scheduler_workers(dask_scheduler=None):
 def setup_memory_pool(pool_size=None, disable_pool=False):
     import cupy
 
+    print("FOODSFSDF")
```
jakirkham, line 189: Guessing this can be dropped.

```suggestion
```

jakirkham:
One question I'm curious about is, what happens if the user does this? Do they simply need to set this environment variable before starting up python then?

```python
import rmm
import dask_cuda
```

quasiben:
@jakirkham you get the following:

```python
In [1]: import rmm

In [2]: from dask_cuda import LocalCUDACluster

In [3]: cluster = LocalCUDACluster(CUDA_VISIBLE_DEVICES=[5,6])
```
```
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
|    0      7948      C   ...n/miniconda3/envs/20200812-1/bin/python   305MiB |
|    5      8215      C   ...n/miniconda3/envs/20200812-1/bin/python   305MiB |
|    6      8218      C   ...n/miniconda3/envs/20200812-1/bin/python   305MiB |
+-----------------------------------------------------------------------------+
```

Note: GPU 0 is due to the initial import of `rmm`

quasiben:
Rerun tests

jakirkham:
I think we can ignore CUDA 11.0 for the moment (talked to Mike about this offline). Going to go ahead and merge. Though let's follow up on the CUDA 11.0 failures after.

jakirkham:
As changes in RMM ( https://github.com/rapidsai/rmm/pull/493 ) make this unnecessary, am proposing reverting in PR ( https://github.com/rapidsai/dask-cuda/pull/367 ).

pentschev:
@kkraus14 wrote this on Slack: "cc @John Kirkham @Peter Entschev @Benjamin Zaitlen for visibility, but I'd prefer you keep the environment variable in dask-cuda to give us flexibility moving forward, possibly changing it to RAPIDS_NO_INITIALIZE". I agree with his statement if we have a possible future use case.

pentschev:
Sorry @john @peter @benjamin for the false alarm, I pasted this and it didn't occur to me this would happen.

PR #367, opened 2020-08-13
Revert RMM init change

jakirkham:
As RMM now delays CUDA initialization (and doesn't include this environment variable) ( https://github.com/rapidsai/rmm/pull/493 ), we are able to revert this change.

cc @quasiben @pentschev @kkraus14 @harrism

### DIFF
--- a/dask_cuda/benchmarks/utils.py
+++ b/dask_cuda/benchmarks/utils.py
@@ -1,5 +1,4 @@
 import argparse
-import os
 
 from dask.distributed import SSHCluster
 
@@ -187,7 +186,6 @@ def get_scheduler_workers(dask_scheduler=None):
 def setup_memory_pool(pool_size=None, disable_pool=False):
     import cupy
 
-    os.environ['RMM_NO_INITIALIZE'] = 'True'
     import rmm
 
     rmm.reinitialize(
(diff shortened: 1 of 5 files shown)


jakirkham:
Though we don't use this environment variable any more. Can we revisit once there is a need?

quasiben:
I think we should switch `RAPIDS_NO_INITIALIZE` rather than removing.  This allows RAPIDS to generally set/search conditions where GPU initialization will be an issue.  They then could use this variable instead of `DASK_PARENT` as they do here: https://github.com/rapidsai/cudf/blob/d8705fe9ba02f9142ef2290068dedcde5dadbe16/python/cudf/cudf/utils/gpu_utils.py#L10

pentschev:
> I think we should switch `RAPIDS_NO_INITIALIZE` rather than removing.

I'm in favor of using that variable, but when you say "switch", switch for what? Are you saying that cuDF should use that instead of `DASK_PARENT` as is done currently?

quasiben:
> I'm in favor of using that variable, but when you say "switch", switch for what? Are you saying that cuDF should use that instead of DASK_PARENT as is done currently?

Yes, I believe @kkraus14 is on board with this switch as well, correct ?

kkraus14:
Yes. Once `dask-cuda` uses `RAPIDS_NO_INITIALIZE` we'll switch to using that in cuDF as opposed to checking for the Dask environment variable.

jakirkham:
Superseded by PR ( https://github.com/rapidsai/dask-cuda/pull/371 ).

PR #376, opened 2020-08-17
Confirm DGX tests are running baremetal

pentschev:
This is helpful to run tests with `ci/local/build.sh` on a DGX machine.

### DIFF
--- a/dask_cuda/tests/test_dgx.py
+++ b/dask_cuda/tests/test_dgx.py
@@ -29,8 +29,12 @@ class DGXVersion(Enum):
 
 def _get_dgx_name():
     product_name_file = "/sys/class/dmi/id/product_name"
+    dgx_release_file = "/etc/dgx-release"
 
-    if not os.path.isfile(product_name_file):
+    # We verify `product_name_file` to check it's a DGX, and check
+    # if `dgx_release_file` exists to confirm it's not a container.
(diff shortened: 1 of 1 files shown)


dask_cuda/tests/test_dgx.py
```diff
@@ -29,26 +29,28 @@ class DGXVersion(Enum):
...
+    dgx_release_file = "/etc/dgx-release"
 
-    if not os.path.isfile(product_name_file):
+    # We verify `product_name_file` to check it's a DGX, and check
+    # if `dgx_release_file` exists to confirm it's not a container.
+    if not os.path.isfile(product_name_file) or not os.path.isfile(dgx_release_file):
         return None
 
     for line in open(product_name_file):
         return line
```
jakirkham, line 40: Should we be doing something with `dgx_release_file` here as well?

pentschev: No, having the information from only one place suffices. The `dgx_release_file` doesn't include the information we need for the A100s but it confirms it's a baremetal machine, the `product_name_file` includes the information we need but it doesn't confirm it's baremetal, thus a combination of both is required, even though it's sufficient to get the DGX type from a single place.

quasiben:
Thanks @pentschev and @madsbk / @jakirkham for the reviews

PR #380, opened 2020-08-18
Change pytest's basetemp in CI build script

pentschev:
This fixes A100 builds that are failing with errors like:

```python
E           OSError: could not create numbered dir with prefix pytest- in /tmp/pytest-of-jenkins after 10 tries
```


quasiben:
@rapidsai/admins can you merge?

pentschev:
Thanks @raydouglass !

Issue #395, opened 2020-08-28
[BUG] Enabling `rmm-managed-memory` without `rmm-pool-size`


ayushdg:
Starting a dask-cuda-worker with the `--rmm-managed-memory` flag without specifying the `--rmm-pool-size` argument does not work.  
```
dask-cuda-worker --rmm-managed-memory schedulerIP

Traceback (most recent call last):
  File "/opt/conda/envs/rapids/bin/dask-cuda-worker", line 33, in <module>
    sys.exit(load_entry_point('dask-cuda==0+untagged.1.g5389f07', 'console_scripts', 'dask-cuda-worker')())
  File "/opt/conda/envs/rapids/lib/python3.7/site-packages/dask_cuda-0+untagged.1.g5389f07-py3.7.egg/dask_cuda/cli/dask_cuda_worker.py", line 279, in go
  File "/opt/conda/envs/rapids/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/opt/conda/envs/rapids/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/opt/conda/envs/rapids/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/conda/envs/rapids/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/opt/conda/envs/rapids/lib/python3.7/site-packages/dask_cuda-0+untagged.1.g5389f07-py3.7.egg/dask_cuda/cli/dask_cuda_worker.py", line 254, in main
  File "/opt/conda/envs/rapids/lib/python3.7/site-packages/dask_cuda-0+untagged.1.g5389f07-py3.7.egg/dask_cuda/cuda_worker.py", line 148, in __init__
  File "/opt/conda/envs/rapids/lib/python3.7/site-packages/dask/utils.py", line 1211, in parse_bytes
    s = s.replace(" ", "")
AttributeError: 'NoneType' object has no attribute 'replace'
```

Rmm allows having a managed memory allocator without enabling the pool allocator while the logic [here](https://github.com/rapidsai/dask-cuda/blob/branch-0.16/dask_cuda/cuda_worker.py#L139-L148) works only when both pool and managed memory are enabled simultaneously.

jakirkham:
[...]

PR #396, opened 2020-08-29
Parse pool size only when set

quasiben:
Fixes https://github.com/rapidsai/dask-cuda/issues/395

cc @ayushdg


pentschev:
@quasiben hope you don't mind, but I pushed equivalent changes to `LocalCUDACluster` as well as some tests.

pentschev (approved):

pentschev:
Thanks @quasiben , merging.

PR #397, opened 2020-09-10
Improve CUDAWorker scheduler-address parsing and __init__

necaris:
Allows `scheduler` argument to be optional, and uses `dask.config` to check the `scheduler-address` rather than dictionary lookup (this allows e.g. environment variables like `DASK_SCHEDULER_ADDRESS` to be correctly parsed)

### DIFF
--- a/dask_cuda/cuda_worker.py
+++ b/dask_cuda/cuda_worker.py
@@ -47,7 +47,7 @@ def _get_interface(interface, host, cuda_device_index, ucx_net_devices):
 class CUDAWorker:
     def __init__(
         self,
-        scheduler,
+        scheduler=None,
         host=None,
         nthreads=0,
         name=None,
--- a/dask_cuda/cuda_worker.py
+++ b/dask_cuda/cuda_worker.py
@@ -9,6 +9,7 @@ import warnings
 from toolz import valmap
 from tornado.ioloop import IOLoop
 
+import dask
 from distributed import Nanny
 from distributed.config import config
 from distributed.proctitle import (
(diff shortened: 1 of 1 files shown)


GPUtester:

GPUtester:

GPUtester:

jrbourbeau:
For context, setting the `DASK_SCHEDULER_ADDRESS` environment variable puts `scheduler_address` (note the underscore) into Dask's config system. We generally try to use `dask.config.get` when interacting with config values as it treats underscores and hyphens the same (xref https://github.com/dask/distributed/pull/2676) 

The `scheduler=None` change is so `CUDAWorker` can better handle other ways in which the scheduler address can be specified (e.g. `scheduler_file` or `scheduler-address` configuration option). There's additional logic here

https://github.com/rapidsai/dask-cuda/blob/381195162564be133339d82b033f58949e400941/dask_cuda/cuda_worker.py#L130-L134

that will still catch cases when no scheduler address is specified

quasiben:
ok to test

dask_cuda/cuda_worker.py
```diff
@@ -9,6 +9,7 @@
 from toolz import valmap
 from tornado.ioloop import IOLoop
 
+import dask
 from distributed import Nanny
 from distributed.config import config
```
quasiben, line 14:
```suggestion
```

Failing due to a linting issue:
> dask_cuda/cuda_worker.py:14:1: F401 'distributed.config.config' imported but unused

mrocklin: The canonical way to do this is to `import dask` and then use `dask.config.get`

mrocklin: Ah, I see that this is already done, and that you're just suggesting removing this.  Woot.

quasiben:
@necaris i made this change on your behalf -- hope you don't mind

quasiben (approved):
Thanks again @necaris

necaris:
@quasiben very welcome!

PR #398, opened 2020-09-11
CI TEST

madsbk:


jakirkham:
rerun tests

jakirkham:
Yeah there’s another issue in Distributed. Seems to just be out-of-date testing code. We have a fix in PR ( https://github.com/dask/distributed/pull/4154 ). It seems to work on the DGX-1. Though Ben ran into an issue on a workstation, the plan is to look at that issue in the morning.

pentschev:
It isn't clear to me, but I think this is just a PR to force CI runs, is that correct? Can we close it?

PR #402, opened 2020-09-25
Skip pyNVML to support Tegra devices

### LINKED ISSUE #400: pyNVML won't work on a Jetson, is there a workaround
There is no NVML library on aarch64 NVIDIA Jetson  That will break many libraries relying on this library, such as cuxfilter.  The geospatial and cuxfilter libraries are among the most requested for Jetson and I'd love to make it work.  Is there a way to use Numba functions to replace pyNVML in this library?

pentschev:
Fixes #400

### DIFF
--- a/dask_cuda/utils.py
+++ b/dask_cuda/utils.py
@@ -139,21 +139,24 @@ def get_cpu_affinity(device_index):
      40, 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 58, 59,
      60, 61, 62, 63, 64, 65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79]
     """
-    pynvml.nvmlInit()
-
-    try:
-        # Result is a list of 64-bit integers, thus ceil(get_cpu_count() / 64)
-        affinity = pynvml.nvmlDeviceGetCpuAffinity(
-            pynvml.nvmlDeviceGetHandleByIndex(device_index),
-            math.ceil(get_cpu_count() / 64),
-        )
-        return unpack_bitmask(affinity)
-    except pynvml.NVMLError:
-        warnings.warn(
-            "Cannot get CPU affinity for device with index %d, setting default affinity"
-            % device_index
-        )
+    if _is_tegra():
         return list(range(get_cpu_count()))
+    else:
+        pynvml.nvmlInit()
+
+        try:
+            # Result is a list of 64-bit integers, thus ceil(get_cpu_count() / 64)
+            affinity = pynvml.nvmlDeviceGetCpuAffinity(
+                pynvml.nvmlDeviceGetHandleByIndex(device_index),
+                math.ceil(get_cpu_count() / 64),
+            )
+            return unpack_bitmask(affinity)
+        except pynvml.NVMLError:
+            warnings.warn(
+                "Cannot get CPU affinity for device with index %d, setting default affinity"
+                % device_index
+            )
+            return list(range(get_cpu_count()))
 
 
 def get_n_gpus():
--- a/dask_cuda/utils.py
+++ b/dask_cuda/utils.py
@@ -171,9 +171,16 @@ def get_device_total_memory(index=0):
     Return total memory of CUDA device with index
     """
     if _is_tegra():
+        from ctypes import byref, c_size_t
         import numba.cuda
 
-        return numba.cuda.current_context().get_memory_info()[1]
+        driver = numba.cuda.driver.Driver()
+
+        numba.cuda.current_context()
+        free = c_size_t()
+        total = c_size_t()
+        driver.cuMemGetInfo(byref(free), byref(total))
+        return total.value
     else:
         pynvml.nvmlInit()
         return pynvml.nvmlDeviceGetMemoryInfo(
--- a/dask_cuda/utils.py
+++ b/dask_cuda/utils.py
@@ -20,6 +20,15 @@ except ImportError:
         yield
 
 
+@toolz.memoize
+def _is_tegra():
+    import os
+
+    return os.path.isdir("/sys/class/tegra-firmware/") or os.path.isfile(
+        "/etc/nv_tegra_release"
+    )
+
+
 class CPUAffinity:
     def __init__(self, cores):
         self.cores = cores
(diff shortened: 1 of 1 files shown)


pentschev:
cc @JasonAtNvidia

jakirkham:
Would this make sense to do generally or have we found issues with this approach in some cases? Asking as we now know of 2 cases where NVML is not an option Tegra and WSL.

pentschev:
> Would this make sense to do generally or have we found issues with this approach in some cases? Asking as we now know of 2 cases where NVML is not an option Tegra and WSL.

Using Numba to do that implies creating a context, and that can be tricky in certain situations where we need the context creation to be delayed. We should evaluate context creation on a case-by-case basis as this has proven problematic in several instances in the past.

TBH, I don't know if things are 100% correct for Tegra, but it's very difficult to evaluate much further without being able to test. Long story short: I hope this PR will provide _some_ support to Tegra but we can't make any guarantees without extensive testing, same applies for WSL or any other platforms we may want to support in the future.

jakirkham:
> Using Numba to do that implies creating a context, and that can be tricky in certain situations where we need the context creation to be delayed. We should evaluate context creation on a case-by-case basis as this has proven problematic in several instances in the past.

Won't it already be tricky to initialize the context? AIUI we wanted this for multi-GPU devices.

> TBH, I don't know if things are 100% correct for Tegra, but it's very difficult to evaluate much further without being able to test. Long story short: I hope this PR will provide _some_ support to Tegra but we can't make any guarantees without extensive testing, same applies for WSL or any other platforms we may want to support in the future.

Fair enough I think we have to rely on folks being pretty engaged to keep things working. This is also why I'm wondering if we can standardize on a common code path that we are able to test in other cases at least.

dask_cuda/utils.py
```diff
@@ -96,8 +105,13 @@ def get_cpu_count():
...
+    if _is_tegra():
+        import numba.cuda
+
+        return len(numba.cuda.gpus)
+    else:
+        pynvml.nvmlInit()
+        return pynvml.nvmlDeviceGetCount()
 
 
 def get_cpu_affinity(device_index):
```
jakirkham, line 117: There is also some `pynvml` calls in this function that we would want to guard against. Currently it calls things like `pynvml.nvmlDeviceGetCpuAffinity` and `pynvml.nvmlDeviceGetHandleByIndex`.

jakirkham: Resolved in recent changes

pentschev:
On a discussion with @JasonAtNvidia and @jakirkham , we decided to move this to 0.17. One of the limitations we found is getting GPU information, such as memory, without NVML. That requires Numba, and since we replace its memory manager with RMM's and RMM doesn't implement such functionality, we get errors:

```python
============================= test session starts ==============================
platform linux -- Python 3.7.8, pytest-6.1.1, py-1.9.0, pluggy-0.13.1 -- /mnt/data/miniforge3/envs/dasktest/bin/python
cachedir: .pytest_cache
rootdir: /home/nvidia/Documents/dask-cuda
collecting ... collected 998 items / 1 error / 1 skipped / 996 selected
==================================== ERRORS ====================================
________________ ERROR collecting dask_cuda/tests/test_spill.py ________________
dask_cuda/tests/test_spill.py:18: in <module>
    if utils.get_device_total_memory() < 1e10:
/mnt/data/miniforge3/envs/dasktest/lib/python3.7/site-packages/dask_cuda-0.16.0a0+95.g04dcbb6.dirty-py3.7.egg/dask_cuda/utils.py:175: in get_device_total_memory
    return numba.cuda.current_context().get_memory_info()[1]
/mnt/data/miniforge3/envs/dasktest/lib/python3.7/site-packages/numba/cuda/cudadrv/driver.py:1040: in get_memory_info
    return self.memory_manager.get_memory_info()
/mnt/data/miniforge3/envs/dasktest/lib/python3.7/site-packages/rmm/rmm.py:214: in get_memory_info
    raise NotImplementedError()
E   NotImplementedError
=============================== warnings summary ===============================
../../../../mnt/data/miniforge3/envs/dasktest/lib/python3.7/site-packages/pandas/util/__init__.py:12
  /mnt/data/miniforge3/envs/dasktest/lib/python3.7/site-packages/pandas/util/__init__.py:12: FutureWarning: pandas.util.testing is deprecated. Use the functions in the public API at pandas.testing instead.
    import pandas.util.testing
dask_cuda/tests/test_local_cuda_cluster.py:76
  /home/nvidia/Documents/dask-cuda/dask_cuda/tests/test_local_cuda_cluster.py:76: PytestUnknownMarkWarning: Unknown pytest.mark.asyncio - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/mark.html
    @pytest.mark.asyncio
dask_cuda/tests/test_local_cuda_cluster.py:89
  /home/nvidia/Documents/dask-cuda/dask_cuda/tests/test_local_cuda_cluster.py:89: PytestUnknownMarkWarning: Unknown pytest.mark.asyncio - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/mark.html
    @pytest.mark.asyncio
-- Docs: https://docs.pytest.org/en/stable/warnings.html
=========================== short test summary info ============================
ERROR dask_cuda/tests/test_spill.py - NotImplementedError
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
=================== 1 skipped, 3 warnings, 1 error in 4.28s ====================
```

pentschev:
> Won't it already be tricky to initialize the context? AIUI we wanted this for multi-GPU devices.

Even single-GPU devices can benefit from these changes for Tegra clusters.

> Fair enough I think we have to rely on folks being pretty engaged to keep things working. This is also why I'm wondering if we can standardize on a common code path that we are able to test in other cases at least.

It seems we'll soon begin testing more with Tegra, so that will help us here. To be frank, I don't want to move away from NVML only because of Tegra, NVML is much better in various ways (e.g., no need to create a CUDA context) and supports things like gathering CPU affinity for each GPU that other tools don't, so I'd rather have two code paths.

pentschev:
It seems we'll have some Tegra devices in CI soon, we can then push this PR forward.

quasiben:
I believe we are still waiting on Tegra devices

pentschev:
@mike-wendt @raydouglass based on the discussion from Ops Demo Tuesday, this is a PR we would like to test with Tegra devices when we have it available in CI.

mike-wendt:
> @mike-wendt @raydouglass based on the discussion from Ops Demo Tuesday, this is a PR we would like to test with Tegra devices when we have it available in CI.

@pentschev @quasiben as an update @Ethyling and I are working on getting the L4T images integrated on Monday. Once we get that done we'll have a platform to easily test these builds for CUDA 10.2 as that's the only current CUDA version supported on L4T/AGX.

pentschev:
Sounds great, thanks @mike-wendt and let us know if/how we can assist.

pentschev:
The changes here along with others for SoC support are being worked further in #1505 . Closing.

Issue #404, opened 2020-10-02
Was this meant to be max?


chrisroat:
It's still early and I haven't had my coffee, but was this supposed to be max?

https://github.com/rapidsai/dask-cuda/blob/b6cfacc799e03379e1d24a0170b9240716727b01/dask_cuda/cuda_worker.py#L89

jakirkham:
Good question. Probably. @pentschev, do you have thoughts here?

quasiben:
My guess is that this should really be `nthreads=1` -- the general assumption for dask-cuda is 1 process 1 thread 1 GPU

pentschev:
Yeah, I generally agree with @quasiben , perhaps that could change in the future when we use support a per-thread default stream mode.

jakirkham:
Submitted PR ( https://github.com/rapidsai/dask-cuda/pull/405 ) to just make this `1`.

pentschev:
This has been changed in #405, closing.

PR #405, opened 2020-10-02
Set `nthreads` to `1`

jakirkham:
Fixes https://github.com/rapidsai/dask-cuda/issues/404

As we don't really support using multiple threads at this time, just set `nthreads` to `1`. This may change in the future if we support per-thread default stream (PTDS).

### DIFF
--- a/dask_cuda/cuda_worker.py
+++ b/dask_cuda/cuda_worker.py
@@ -2,7 +2,6 @@ from __future__ import absolute_import, division, print_function
 
 import asyncio
 import atexit
-import multiprocessing
 import os
 import warnings
 
@@ -86,7 +85,7 @@ class CUDAWorker:
             nprocs = get_n_gpus()
 
         if not nthreads:
-            nthreads = min(1, multiprocessing.cpu_count() // nprocs)
+            nthreads = 1
 
         memory_limit = parse_memory_limit(memory_limit, nthreads, total_cores=nprocs)
 


pentschev (approved):
LGTM, thanks @jakirkham .

mrocklin:
Ooh, this is the first I'm hearing of the term "per thread default stream",
but that sounds like a really pragmatic policy to me :)

pentschev:
"CUDA 7 introduces a new option, the per-thread default stream, that has two effects. First, it gives each host thread its own default stream. This means that commands issued to the default stream by different host threads can run concurrently. Second, these default streams are regular streams. This means that commands in the default stream may run concurrently with commands in non-default streams."

Source: https://developer.nvidia.com/blog/gpu-pro-tip-cuda-7-streams-simplify-concurrency/

jakirkham:
rerun tests

mrocklin:
Yeah, that's awesome.  I'm curious what benchmarks would look like with
nthreads=2 , just to get some overlap going on.  Anyway, sorry for the
tangent.  I'm sure that this is something that you all have been thinking
about for some time.

pentschev:
We actually need more support in libraries such as cuDF and CuPy first, but that's coming. The dask-cuda part is almost free after that.

chrisroat:
I feel this would slightly more readable if the c'tor default was changed to one.

Another use case for >1 thread is having the extra threads to do CPU tasks.  One could submit a graph with a mix of CPU and GPU work, using worker resources to allow one GPU task per worker.

pentschev:
> I feel this would slightly more readable if the c'tor default was changed to one.

That's a good point, could we do that instead @jakirkham ? That's in fact the behavior in https://github.com/rapidsai/dask-cuda/blob/b6cfacc799e03379e1d24a0170b9240716727b01/dask_cuda/local_cuda_cluster.py#L137 , which seems reasonable to me.

> Another use case for >1 thread is having the extra threads to do CPU tasks. One could submit a graph with a mix of CPU and GPU work, using worker resources to allow one GPU task per worker.

Just to be clear, we don't officially support hybrid GPU/CPU usage of dask-cuda today, but there's no harm in providing this capability IMO.

jakirkham:
Well I guess I'm wondering whether we need this argument at all considering we want it set to `1` and no other value makes sense. Should we just drop it?

pentschev:
> Well I guess I'm wondering whether we need this argument at all considering we want it set to 1 and no other value makes sense. Should we just drop it?

We don't actually need it for our current workflows. However, @chrisroat 's point is that users may want to have multiple threads, it's probably ok to allow them to still use that argument to specify the number of threads they want to use it, even though we won't support hybrid workflows at this time. Also we may have a real use for this in the near future, when we have full support for PTDS, so IMO we should keep it for now.

pentschev:
Failures here are also related to https://github.com/rapidsai/dask-cuda/pull/406#issuecomment-703683511 .

jakirkham:
Though we currently run things on stream `0` and synchronize periodically. I'm not sure one would get much benefit out of multiple threads without PTDS. It's also possible that they would be confused about that lack of performance improvement and raise issues. Seems better IMO to disallow a potentially problematic option and re-enable it in the future when those issues are fixed.

pentschev:
> Though we currently run things on stream 0 and synchronize periodically. I'm not sure one would get much benefit out of multiple threads without PTDS. It's also possible that they would be confused about that lack of performance improvement and raise issues. Seems better IMO to disallow a potentially problematic option and re-enable it in the future when those issues are fixed.

I would then say we should better document that behavior. In either case, it seems like @chrisroat has a specific use case in mind for this, it would be great if you could elaborate on your use case to help us understand what are users really doing with it. We also don't know if anybody else is currently relying on this.

I think @benjha and @jglaser tried playing with the number of threads, could you also provide some information on whether that was useful in your case, just so we have another data point here?

chrisroat:
> Just to be clear, we don't officially support hybrid GPU/CPU usage of dask-cuda today

Ah, I didn't realize that.  I ended up on these threads (pun intendend) because I was planning to to try dask-cuda with multiple threads and 1 GPU.   Currently, I'm using 1 thread per worker with dask-worker, but I was going to experiment with multiple threads -- its possible there would be an efficiency win.

I'm using dask-gateway to create a K8s cluster with 1 GPU per worker.  One of my task graph consists of two consecutive map_overlaps calls, where the 2nd one needs the information from the 1st (essentially neighboring blocks communicate with each other at their borders).  The first map_overlaps uses GPU, where the second one does not.  

I think my use case is covered with standard dask-worker, but I was trying to understand if dask-cuda was a better option.

jakirkham:
My guess is you are probably well served with the normal Dask workers with multiple threads. We do some special setup here when multiple GPUs are involved like initializing the GPUs, setting CPU affinity, configuring memory pools on workers, etc. For a single GPU, my guess is these sorts of things could be done initially before creating the Dask cluster and things would just work. I could be wrong about that though. Would give that a try first and if there are issues report back.

jakirkham:
> Failures here are also related to [#406 (comment)](https://github.com/rapidsai/dask-cuda/pull/406#issuecomment-703683511) .

Have filed as issue ( https://github.com/rapidsai/dask-cuda/issues/407 ).

jakirkham:
Moving to 0.17

pentschev:
Can we add a `--threads-per-worker` argument here instead of hardcoding `1`, @jakirkham ? This is going to be useful as we go through PTDS testing as well. If you don't have the time, I can take this over.

jakirkham:
Agreed. Please feel free to take this over 🙂

pentschev:
I thought it would be easier starting a new PR, so I opened #456 and I'm closing this one. Thanks @jakirkham for the initiative on fixing the behavior of threads per worker. :)

jakirkham:
Makes sense. Thanks for picking this up Peter! 😄

PR #411, opened 2020-10-07
Skip tests that require cuDF/UCX-Py, when not installed

pentschev:


jakirkham:
Looks like `black` and `flake8` are failing

https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/rapidsai%2Fgpuci%2Fdask-cuda%2Fprb%2Fdask-cuda-style/detail/dask-cuda-style/588/pipeline

pentschev:
Pretty bizarre, I fixed some issues, but in the current build where style checking passed, formatting is still failing https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/dask-cuda/job/prb/job/dask-cuda-style/589/display/redirect. What's even more bizarre is that my local black report is very different from that of CI, and flake8 reports nothing locally.

jakirkham:
We pin the `flake8` and `black` versions ( https://github.com/rapidsai/integration/pull/74 ). So it may be discrepancies between the latest versions and what we have pinned.

pentschev:
Aha! That makes so much more sense @jakirkham . I now fixed black formatting, but CI still complains about flake8, even though it doesn't tell us anything about the error, and locally I installed the same version as CI https://github.com/rapidsai/integration/pull/74/files#diff-16b61a6b2eb5b806543058cb7d3fca83R59 but don't see any errors.

One more interesting thing is isort complains about an error, and if you fix that, then black complains about that and fixes that, which in turn fails for isort. We really need to keep ordering consistent. :)

jakirkham:
Yeah I was running into an error like this before as well. Had opened PR ( https://github.com/rapidsai/dask-cuda/pull/350 ) to address it, but I don't think we settled on exactly how we want to resolve the conflict between these tools.

jakirkham:
rerun tests

jakirkham:
rerun tests

jakirkham (approved):

jakirkham:
Thanks Peter! 😄

PR #415, opened 2020-10-15
Fix typo in benchmark utils

manopapad:


quasiben (approved):
Thanks @manopapad ! This looks good to me.

pentschev:
I was taking a better look at this now, and it seems like `--single-node` is completely useless -- I think I added that but I now don't know why. I think the right thing to do is remove the following block entirely:

https://github.com/rapidsai/dask-cuda/blob/0411e3b1e5205cc6a2c2ceb3d2d90d03721cc58c/dask_cuda/benchmarks/utils.py#L74-L79

And then change the default of `--multi-node` to `store_false` in 

https://github.com/rapidsai/dask-cuda/blob/0411e3b1e5205cc6a2c2ceb3d2d90d03721cc58c/dask_cuda/benchmarks/utils.py#L82

Could you take care of that @manopapad ? If not, I can do that later.

manopapad:
#416 supersedes this!

PR #416, opened 2020-10-15
Fix typo in benchmark utils

quasiben:
@manopapad  it looked you branched from `main`.  I cherry picked the commit and applied it to branch-0.17 for a clean commit history 

Supersedes https://github.com/rapidsai/dask-cuda/pull/415/commits/

### DIFF
--- a/dask_cuda/benchmarks/utils.py
+++ b/dask_cuda/benchmarks/utils.py
@@ -73,7 +73,7 @@ def parse_benchmark_args(description="Generic dask-cuda Benchmark", args_list=[]
     )
     parser.add_argument(
         "--single-node",
-        action="store_true",
+        action="store_false",
         dest="multi_node",
         help="Runs a single-node cluster on the current host.",
     )
--- a/dask_cuda/benchmarks/utils.py
+++ b/dask_cuda/benchmarks/utils.py
@@ -72,12 +72,6 @@ def parse_benchmark_args(description="Generic dask-cuda Benchmark", args_list=[]
         "Ignored if protocol is 'tcp'",
     )
     parser.add_argument(
-        "--single-node",
-        action="store_false",
-        dest="multi_node",
-        help="Runs a single-node cluster on the current host.",
-    )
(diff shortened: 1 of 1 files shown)


pentschev:
Could you then do https://github.com/rapidsai/dask-cuda/pull/415#issuecomment-709320908 ?

quasiben:
> Could you then do #415 (comment) ?

Addressed in 44b8543

quasiben:
@pentschev when you have time can you give this another review ?

dask_cuda/benchmarks/utils.py
```diff
@@ -71,12 +71,6 @@ def parse_benchmark_args(description="Generic dask-cuda Benchmark", args_list=[]
...
     )
-    parser.add_argument(
-        "--single-node",
-        action="store_true",
-        dest="multi_node",
-        help="Runs a single-node cluster on the current host.",
-    )
     parser.add_argument(
         "--multi-node",
         action="store_true",
```
pentschev, line 76:
```suggestion
        action="store_false",
```

pentschev (changes requested):

pentschev:
@quasiben we want `--multi-node` to be `False` by default, added a suggestion above.

quasiben:
Dine

pentschev (approved):
Thanks @quasiben !

quasiben:
Wait, when `--multi-node` is not set it should be `False` in which case, I think we want `store_true`, right ?

pentschev:
> Wait, when `--multi-node` is not set it should be `False` in which case, I think we want `store_true`, right ?

Yes, you're right. I confused things with the whole `--enable-`/`--disable-` behavior. We do want to remove `--single-node` (and this is implicitly the default behavior), and have `--multi-node` set to `store_true`. Can you revert the latest commit?

quasiben:
Thanks for the confirmation.  Reverted the commit

pentschev:
Thanks @quasiben for noticing!

PR #419, opened 2020-10-19
More RMM options in benchmarks

quasiben:
`--no-rmm-pool` now disables the pool entirely
`--disable-rmm-pool` will reinitialize rmm, however, the pool allocator is false

These names are not great and I'd be happy to change them

### DIFF
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -331,10 +331,10 @@ def parse_args():
                 "-c",
                 "--chunk-size",
             ],
-            "default": 40_000_000,
+            "default": 1_000_000,
             "metavar": "n",
             "type": int,
-            "help": "Chunk size (default 40_000_000)",
+            "help": "Chunk size (default 1_000_000)",
         },
         {
             "name": "--ignore-size",
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -207,8 +207,9 @@ def main(args):
             client.run(setup_memory_pool, disable_pool=args.disable_rmm_pool)
             # Create an RMM pool on the scheduler due to occasional deserialization
             # of CUDA objects. May cause issues with InfiniBand otherwise.
-            client.run_on_scheduler(setup_memory_pool, 1e9,
-                    disable_pool=args.disable_rmm_pool)
+            client.run_on_scheduler(
+                setup_memory_pool, 1e9, disable_pool=args.disable_rmm_pool
+            )
 
     scheduler_workers = client.run_on_scheduler(get_scheduler_workers)
     n_workers = len(scheduler_workers)
@@ -236,7 +237,10 @@ def main(args):
         for (w1, w2), v in bandwidths.items()
     }
     total_nbytes = {
-        (scheduler_workers[w1].name, scheduler_workers[w2].name,): format_bytes(sum(nb))
+        (
+            scheduler_workers[w1].name,
+            scheduler_workers[w2].name,
+        ): format_bytes(sum(nb))
         for (w1, w2), nb in total_nbytes.items()
     }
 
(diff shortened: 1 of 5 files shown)


dask_cuda/benchmarks/local_cudf_merge.py
```diff
@@ -192,7 +192,7 @@ def main(args):
     if args.sched_addr:
         client = Client(args.sched_addr)
     else:
-        cluster = Cluster(*cluster_args, **cluster_kwargs)
+        cluster = Cluster(*cluster_args, **cluster_kwargs, rmm_pool_size="20GB")
```
pentschev, line 195: We shouldn't enforce some specific size, let's either allow the user to specify a size or use the default (half the GPU's total memory). This will immediately cause problems with 16GB GPUs, which are still very common.

dask_cuda/benchmarks/utils.py
```diff
@@ -183,6 +188,8 @@ def setup_memory_pool(pool_size=None, disable_pool=False):
     import rmm
 
     rmm.reinitialize(
-        pool_allocator=not disable_pool, devices=0, initial_pool_size=pool_size,
+        pool_allocator=not disable_pool,
+        devices=0,
+        initial_pool_size=pool_size,
```
pentschev, line 193: I think we should change `--disable-rmm-pool` to `--enable-rmm-pool`, and we can have it enabled by default. The disable behavior is confusing, as I think most people would expect it not to do anything. IOW, I would expect something like:

```python
if enable_rmm_pool:
    rmm.reinitialize(...)
```

pentschev (changes requested):

dask_cuda/benchmarks/local_cudf_merge.py
quasiben: That was a mistake .  I'll take that out

dask_cuda/benchmarks/utils.py
quasiben: Maybe we should take this out all together.  Who was using this feature ?  @madsbk or @rjzamora do you recall why this is here ?

pentschev: What feature are you talking about? Note that we have at least two different use cases:

1. Cluster created by the benchmark itself (with or without `--multi-node`): we want to set the pool;
1. Cluster started externally: we may _not_ want to set the pool.

Less important, that can be useful to test behavior if we think there may be an issue related to the pool.

quasiben: I'm talking about the flag which controls `pool_allocator=not disable_pool` .  When do we want rmm to be reinitialized without the pool ?  If we have the flag which controls the entire block of rmm I'm not sure we need the second flag

pentschev: In the current context, I don't think we should reinitialize unless we're setting a pool, thus the `if enable_rmm_pool` code above, we're not doing any modifications in that case so we don't need to reinitialize. Plus, we don't want to mess with the a cluster that's started externally, because of CUDA IPC not de-registering memory, we invalidate a cluster once we try to `rmm.reinitialize`, meaning the user would require to restart the cluster when running `local_cuda_merge.py` multiple times.

quasiben:
After chatting with @pentschev I removed the no-rmm-pool and now `disable-rmm-pool` will no longer result in a reinitialization of RMM.  Additionally, I added an option for defining the rmm pool size (default is 1/2 the GPU)

quasiben:
rerun tests

quasiben:
rerun tests

pentschev:
It all looks good to me, but the tests seem weird. I can't reproduce them locally, can you @quasiben ?

quasiben:
rerun tests

quasiben:
I can't reproduce locally either

quasiben:
rerun tests

quasiben:
rerun tests

quasiben:
@pentschev I think this can be merged in now.  The issue was that I changed the default chunk size.  Having reverted that commit, the CI now passes

pentschev:
I don't actually think this was the cause because what was failing was a test, not benchmarks. However, I agree we should merge this before something else breaks. Thanks @quasiben for working on it. :)

pentschev (approved):

Issue #403, opened 2020-09-30
[FEA] All to All Initialization


quasiben:
Awhile @kkraus14 and I were chatting about doing an all-to-all for ucx to establish various IPC handles as part of a warmup before a workflow.  With a bit of help from @jakirkham I think encoding something like the following in a function (dask_cuda.all_to_all(client)) would work (functional example below): 

```python
from itertools import combinations

from dask.distributed import Client, wait

def main():
    n_workers = 3
    client = Client(n_workers=n_workers)

    workers = list(client.scheduler_info()["workers"])
    futs = []
    for w in workers:
        bit_of_data = b"0" * 1
        data = client.map(lambda x: bit_of_data, range(1), pure=False, workers=[w])
        futs.append(data[0])

    wait(futs)

    # assert 1 key on each worker
    for w, keys in client.who_has().items():
        assert len(keys) == 1

    def f(x,y):
        pass

    new_futs = []
    for w in workers:
        for fut1, fut2 in combinations(futs, 2):
            data = client.submit(f, fut1, fut2, workers=[w], pure=False)
            new_futs.append(data)

    wait(new_futs)
    # delete noop funcs futures
    client.cancel(new_futs)

    # assert all to all has resulted in all data on every worker
    for key, w in client.has_what().items():
        assert len(w) == n_workers

if __name__ == "__main__":
    main()
```

Note we would need to replace `bit_of_data` with GPU data -- `cupy.arange(10)` or something smallish but larger thank 8192 bytes

pentschev:
This seems like a good idea, I would suggest this as an option rather than enforcing it on every start. Also note that we would have to check for rendezvous numbers instead of hardcoding 8192 as is the default in UCX-Py, see https://github.com/rapidsai/ucx-py/blob/branch-0.16/ucp/__init__.py#L31 .

quasiben:
Definitely an option!

jakirkham:
As to checking the rendezvous threshold, we should be able to just check `ucp.get_config()["RNDV_THRESH"]`, which gets the value that UCX itself is using.

PR #420, opened 2020-10-19
Add utility function to establish all-to-all connectivity upon request

quasiben:
resolves https://github.com/rapidsai/dask-cuda/issues/403

cc @pentschev

### DIFF
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -212,7 +212,8 @@ def main(args):
     n_workers = len(scheduler_workers)
     client.wait_for_workers(n_workers)
 
-    all_to_all(client)
+    if not args.all_to_all:
+        all_to_all(client)
 
     took_list = []
     for _ in range(args.runs - 1):
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -212,7 +212,7 @@ def main(args):
     n_workers = len(scheduler_workers)
     client.wait_for_workers(n_workers)
 
-    if not args.all_to_all:
+    if args.all_to_all:
         all_to_all(client)
 
     took_list = []
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -10,6 +10,7 @@ from dask.distributed import Client, performance_report, wait
 from dask.utils import format_bytes, format_time, parse_bytes
 
 from dask_cuda import explicit_comms
+from dask_cuda.utils import all_to_all
 from dask_cuda.benchmarks.utils import (
     get_cluster_options,
     get_scheduler_workers,
@@ -167,7 +168,6 @@ def run(client, args, n_workers, write_profile=None):
     ).persist()
     wait(ddf_base)
     wait(ddf_other)
-    client.wait_for_workers(n_workers)
 
     assert len(ddf_base.dtypes) == 2
     assert len(ddf_other.dtypes) == 2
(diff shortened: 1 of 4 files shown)


dask_cuda/benchmarks/local_cudf_merge.py
```diff
@@ -210,6 +210,9 @@ def main(args):
 
     scheduler_workers = client.run_on_scheduler(get_scheduler_workers)
     n_workers = len(scheduler_workers)
+    client.wait_for_workers(n_workers)
+
+    all_to_all(client)
```
pentschev, line 215: Maybe we can add a flag to enable this? I would currently prefer it disabled by default, just to be closer to what's expected from a regular user's code today. Although, this may be a fix to the issue that @quasiben and I were debugging today, still to be confirmed.

dask_cuda/utils.py
```diff
@@ -357,3 +362,38 @@ def wait_workers(
...
+    for w in workers:
+        for fut1, fut2 in combinations(futs, 2):
+            data = client.submit(f, fut1, fut2, workers=[w], pure=False)
+            new_futs.append(data)
+
+    await wait(new_futs)
+
+    if cleanup:
+        await client.cancel(futs)
+        await client.cancel(new_futs)
```
pentschev, line 393: I don't think I understand this, what's it useful for and when would you use `cleanup=False`?

quasiben: verifying in a test -- I can take it out if you don't want it

quasiben:
@manopapad I am seeing some success with this PR for cleaning up the first pass of the benchmark in both the lab and Prom.  If you have a minute can you test ?

dask_cuda/utils.py
pentschev: I would be in favor of a cleaner function, which in a few months will just be some mysterious code that will be difficult to remember. However, if that's important for the test, it would be great to at least document what it does.

dask_cuda/utils.py
```diff
@@ -357,3 +362,38 @@ def wait_workers(
...
+    await wait(futs)
+
+    def f(x, y):
+        pass
+
+    new_futs = []
+    for w in workers:
+        for fut1, fut2 in combinations(futs, 2):
+            data = client.submit(f, fut1, fut2, workers=[w], pure=False)
+            new_futs.append(data)
```
manopapad, line 387: What this seems to be doing is: for each worker `A`, for each pair of workers `(B,C)`, pull data from `B` and `C` and combine them on `A`, i.e. `O(n_workers^3)` operations. If the goal is simply to "prime" each worker-to-worker connection with a small message, then I believe it would be sufficient to do the following, which only does `O(n_workers^2)` operations:

```
def f(x):
    pass

new_futs = []
for w in workers:
    for data_in in futs:
        data_out = client.submit(f, data_in, workers=[w], pure=False)
        new_futs.append(data_out)
```

manopapad:
With this PR the `local_cudf_merge.py` benchmark no longer throws the "comm closed" error for me on prom.

pentschev:
It seems good that this makes the errors go away, I think this should still not be necessary to prevent those errors from happening, and we probably still have a bug elsewhere.

pentschev:
I was able to identify that distributed 2.22 works and this error first shows up in 2.23. After doing a bisect, I narrowed the source of the problem to https://github.com/dask/distributed/pull/4019 , but I'm still not certain of what in that PR is responsible for the errors.

dask_cuda/utils.py
quasiben: Addressed in 381f863 .  Thanks @manopapad

quasiben:
@pentschev I think this one is ready to go for another round of reviews when you have a moment

quasiben:
rerun tests

pentschev (approved):
LGTM, thanks @quasiben for working on this!

PR #428, opened 2020-10-27
Add support for connecting a CUDAWorker to a cluster object

jacobtomlinson:
While working on remote clusters like in Dask Cloudprovider or Dask Kubernetes it became apparent that you may end up wasting the local GPU.

For instance if I launch a GPU cluster on AWS from a GPU session in Sagemaker the Sagemaker GPU is not part of the cluster.

```python
from dask_cloudprovider import EC2Cluster
from dask.distributed import Client

cluster = EC2Cluster(**gpu_kwargs)
client = Client(cluster)  # Cluster with remote GPUs
```

This PR makes it possible to include the local GPU in the cluster in the same way you would connect a client.

```python
from dask_cloudprovider import EC2Cluster
from dask.distributed import Client
from dask_cuda import CUDAWorker

cluster = EC2Cluster(**gpu_kwargs)
local_worker = CUDAWorker(cluster)
client = Client(cluster)  # Cluster with remote GPUs and local GPU
```

### DIFF
--- a/dask_cuda/local_cuda_cluster.py
+++ b/dask_cuda/local_cuda_cluster.py
@@ -162,8 +162,10 @@ class LocalCUDACluster(LocalCluster):
         CUDA_VISIBLE_DEVICES = list(map(int, CUDA_VISIBLE_DEVICES))
         if n_workers is None:
             n_workers = len(CUDA_VISIBLE_DEVICES)
+        if n_workers < 1:
+            raise RuntimeError("Number of workers cannot be less than 1.")
         self.host_memory_limit = parse_memory_limit(
-            memory_limit, threads_per_worker, n_workers or 1
+            memory_limit, threads_per_worker, n_workers
         )
         self.device_memory_limit = device_memory_limit
 
@@ -283,9 +285,7 @@ class LocalCUDACluster(LocalCluster):
         visible_devices = cuda_visible_devices(worker_count, self.cuda_visible_devices)
         spec["options"].update(
             {
-                "env": {
-                    "CUDA_VISIBLE_DEVICES": visible_devices,
-                },
+                "env": {"CUDA_VISIBLE_DEVICES": visible_devices,},
                 "plugins": {
                     CPUAffinity(get_cpu_affinity(worker_count)),
                     RMMSetup(self.rmm_pool_size, self.rmm_managed_memory),
--- a/dask_cuda/local_cuda_cluster.py
+++ b/dask_cuda/local_cuda_cluster.py
@@ -163,7 +163,7 @@ class LocalCUDACluster(LocalCluster):
         if n_workers is None:
             n_workers = len(CUDA_VISIBLE_DEVICES)
         self.host_memory_limit = parse_memory_limit(
-            memory_limit, threads_per_worker, n_workers
+            memory_limit, threads_per_worker, n_workers or 1
         )
         self.device_memory_limit = device_memory_limit
 
@@ -283,7 +283,9 @@ class LocalCUDACluster(LocalCluster):
         visible_devices = cuda_visible_devices(worker_count, self.cuda_visible_devices)
         spec["options"].update(
             {
-                "env": {"CUDA_VISIBLE_DEVICES": visible_devices,},
+                "env": {
+                    "CUDA_VISIBLE_DEVICES": visible_devices,
+                },
                 "plugins": {
                     CPUAffinity(get_cpu_affinity(worker_count)),
                     RMMSetup(self.rmm_pool_size, self.rmm_managed_memory),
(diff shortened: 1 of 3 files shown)


dask_cuda/local_cuda_cluster.py
```diff
@@ -163,7 +163,7 @@ def __init__(
         if n_workers is None:
             n_workers = len(CUDA_VISIBLE_DEVICES)
         self.host_memory_limit = parse_memory_limit(
-            memory_limit, threads_per_worker, n_workers
+            memory_limit, threads_per_worker, n_workers or 1
```
jacobtomlinson, line 166: This is just to avoid a divide by zero error when creating `LocalCUDACluster` with `n_workers=0`. Which is probably a rare situation.

jacobtomlinson:
I wonder if we should also detect some of the arguments from the cluster object and pass them on to the worker. Perhaps @pentschev has thoughts?

dask_cuda/local_cuda_cluster.py
pentschev: Makes sense to me, I still don't understand why would anyone startup a `LocalCUDACluster` with 0 workers though, is there more to it besides your test case? If you're only creating the cluster for the scheduler, wouldn't it make more sense to test creating a scheduler and then a worker instead of `LocalCUDACluster` with 0 workers?

pentschev:
> I wonder if we should also detect some of the arguments from the cluster object and pass them on to the worker. Perhaps @pentschev has thoughts?

I'm not sure about this either, and to be fair, I also don't see anybody instantiating `CUDAWorker` directly, so I'm not sure if something may break, although the specialized use cases are often created given constructor arguments, such as [CPU affinity and RMM](https://github.com/rapidsai/dask-cuda/blob/fee2eee39dee88c2fdc55d6d339872a33ba257c5/dask_cuda/cuda_worker.py#L195-L198), [UCX configs](https://github.com/rapidsai/dask-cuda/blob/fee2eee39dee88c2fdc55d6d339872a33ba257c5/dask_cuda/cuda_worker.py#L201-L210) and [`DeviceHostFile` for device spilling](https://github.com/rapidsai/dask-cuda/blob/fee2eee39dee88c2fdc55d6d339872a33ba257c5/dask_cuda/cuda_worker.py#L211-L223).

dask_cuda/local_cuda_cluster.py
jacobtomlinson: Yeah I don't see this happening except for my test case.

pentschev: Sorry @jacobtomlinson , probably my fault this has stalled, but after taking another look at it, maybe we should change to `n_workers=1` in the test, and then `await client.wait_for_workers(2)`, rather than leaving room for users referring to an unsupported behavior. As a bonus, we could check and error when `n_workers < 1`. What do you think?

jacobtomlinson: Thanks @pentschev that sounds reasonable.

Updated to reflect this.

dask_cuda/local_cuda_cluster.py
```diff
@@ -162,6 +162,8 @@ def __init__(
         CUDA_VISIBLE_DEVICES = list(map(int, CUDA_VISIBLE_DEVICES))
         if n_workers is None:
             n_workers = len(CUDA_VISIBLE_DEVICES)
+        if n_workers < 1:
+            raise RuntimeError("Number of workers cannot be less than 1.")
```
pentschev, line 166:
```suggestion
            raise ValueError("Number of workers cannot be less than 1.")
```

pentschev: I think `ValueError` is a more appropriate exception here.

dask_cuda/tests/test_local_cuda_cluster.py
```diff
@@ -146,3 +146,15 @@ async def test_rmm_managed():
...
+
+
+@gen_test(timeout=20)
+async def test_cluster_worker():
+    async with LocalCUDACluster(
+        scheduler_port=0, asynchronous=True, device_memory_limit=1, n_workers=1,
+    ) as cluster:
+        assert len(cluster.workers) == 1
+        async with Client(cluster, asynchronous=True) as client:
+            new_worker = CUDAWorker(cluster)
```
pentschev, line 158:
```suggestion
            new_worker = CUDAWorker(cluster)
            await new_worker
```

pentschev: I think this should solve the failure.

pentschev (changes requested):
Thanks @jacobtomlinson , I posted a couple of comments, one of them will probably resolve the failing test.

pentschev:
Thanks @jacobtomlinson , does this need to be in 0.17 or can we retarget it to 0.18?

jacobtomlinson:
No rush here. Happy to retarget.

pentschev (approved):
Thanks for confirming, I retargeted it and all looks good to me, thanks Jacob!

jakirkham:
Thanks Jacob for working on this and Peter for reviewing! 😄

PR #430, opened 2020-10-27
Decrease data size to shorten spilling tests time

pentschev:
These tests generally take a lot of time, I think it's safe to reduce their sizes so that dask-cuda CI runs faster. Spilling tests alone used to take 228 seconds and take 72 seconds after this PR.

### DIFF
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -11,13 +11,13 @@ from dask.distributed import Client, performance_report, wait
 from dask.utils import format_bytes, format_time, parse_bytes
 
 from dask_cuda import explicit_comms
-from dask_cuda.utils import all_to_all
 from dask_cuda.benchmarks.utils import (
     get_cluster_options,
     get_scheduler_workers,
     parse_benchmark_args,
     setup_memory_pool,
 )
+from dask_cuda.utils import all_to_all
 
 # Benchmarking cuDF merge operation based on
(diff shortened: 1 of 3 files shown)


pentschev:
I've been seeing explicit-comms tests being very unstable for the last month or so. I'm not exactly sure how we should deal with that, any ideas @madsbk @rjzamora ? One alternative is just to xfail them for now, but that may cover some actual issues down the road.

rjzamora:
> I've been seeing explicit-comms tests being very unstable for the last month or so. I'm not exactly sure how we should deal with that, any ideas @madsbk @rjzamora ? One alternative is just to xfail them for now, but that may cover some actual issues down the road.

What kind of instabilities are you seeing?  xfailing these tests could be fine, since explicit comms are rarely used in practice.  However, we may just be ignoring a real problem that is merely being exposed by those tests (although it seems just as likely that something in explicit comms could be the root problem).

madsbk:
I am fine with xfailing the tests as well. Do we have an issue tracking the CI failure?

pentschev:
Sorry @rjzamora @madsbk , I should have filed an issue for that before, I did it now in https://github.com/rapidsai/dask-cuda/issues/431 . For the time being, I will xfail those tests just to unblock builds.

pentschev:
rerun tests

quasiben (approved):
Merge when Green

pentschev:
Thanks @quasiben , I guess your merge won't work as codecov failed, I'm going ahead and merging this.

PR #434, opened 2020-10-30
Move benchmarks into a top-level directory

jakirkham:
As we don't need to ship benchmarks in Conda packages or installs, move them out of the `dask_cuda` repo. This matches what we do with UCX-Py as well.


pentschev:
Wouldn't be good to instead ship them?

quasiben:
I'm not opposed to moving them but I would like to say that having them in the conda package has been quite convenient for testing and helping other folks validate a working UCX setup

Issue #334, opened 2020-07-01
[FEA] Allow spilling from device memory to host memory by default


beckernick:
In some workflows, it's common to set a `device_memory_limit` to allow spilling from device to host memory once a memory threshold is reached. By default, the `device_memory_limit` is configured to be the total memory of the GPU, which functionally means that spilling is off. For workflows that are not memory constrained, this is nice, as it lets us use the full memory availability without trying to spill at all.

Many workflows are at least partially memory constrained, though, which then with default settings will pose significant risks of out-of-memory. It would be nice to default `device_memory_limit` to a lower percentage of GPU memory to trigger spilling rather than going out-of-memory.

Anecdotally, it feels like 75 or 80% of total memory capacity might be a good default threshold. There may also be a better way to achieve spilling by default. I'm interested to hear others' thoughts.

pentschev:
Personally, I don't share the same thinking. While I see your point that going OOM is terrible users will at least be notified of that. Now if we change the default so that spilling is always enabled this will most of the time allow workflows to complete, but will definitely not deliver good performance, what may discourage newcomers from using dask-cuda in general.

To add to the problem, it is known that spilling is still very inefficient -- see https://github.com/rapidsai/dask-cuda/issues/106 .

pentschev:
It would be nice to see other opinions though.

cc @quasiben @jakirkham

beckernick:
> Now if we change the default so that spilling is always enabled this will most of the time allow workflows to complete, but will definitely not deliver good performance, what may discourage newcomers from using dask-cuda in general.

This is a good point, and something I thought about too. The performance / user experience tradeoff is important to consider. It might be valuable to learn about mainline Dask's experience setting defaults. I believe the spill-to-disk behavior is by default set to occur at 70% of CPU memory capacity. @quasiben may be able to comment more on how they selected that threshold

quasiben:
I am not sure how Dask selected the 60% default threshold (happy to ask around).  If I had to guess it may be because Dask defaults were configured to have Dask run without too many errors out of the box and then fine tuning could happen after a workflow was established. 

Perhaps this point, about whether to crash due to OOM vs complete and fine-tune should be considered.  I honestly can see arguments for both.  I would suggest leaving this issue open for some time to gather more feedback from other dask-cuda users.  If users are always setting some device memory limit it may make sense to change to some reasonable default

jakirkham:
[...]

PR #439, opened 2020-11-06
Unify `device_memory_limit` parsing and set default to 0.8

pentschev:
This should address the request to set a default value for `device_memory_limit`, as well as allowing fractions and unifying how `LocalCUDACluster` and `dask-cuda-worker` parse its value. Fixes https://github.com/rapidsai/dask-cuda/issues/334 .

cc @beckernick @rjzamora


quasiben (approved):
Thank you @pentschev .  Should we also update documentation ?

quasiben:
I mean, do you want to call out the default setting in the docs?  I'm happy to do this if you'd lke

pentschev:
> I mean, do you want to call out the default setting in the docs? I'm happy to do this if you'd like

Are you referring to some docstrings or https://dask-cuda.readthedocs.io/en/latest/ ?

quasiben:
https://dask-cuda.readthedocs.io/

pentschev:
Sure, we can update that next week. I just found out that we can't mix types with `click.option`, which causes problems with having `float`/`int`/`str` in `--device-memory-limit`, something that we do in `LocalCUDACluster`, so I'll have to find a solution for that too.

pentschev:
> Sure, we can update that next week. I just found out that we can't mix types with `click.option`, which causes problems with having `float`/`int`/`str` in `--device-memory-limit`, something that we do in `LocalCUDACluster`, so I'll have to find a solution for that too.

Hopefully this is fixed in latest commits.

pentschev:
@quasiben I updated the docs in https://github.com/rapidsai/dask-cuda/pull/439/commits/6e51a2a3b11518b6c44907ff57ce3984d993bf67, please take a look when you have a chance.

docs/source/specializations.rst
```diff
@@ -13,6 +13,8 @@ Spilling From Device
 
 For applications that do not fit in GPU memory, Dask-CUDA supports spilling from device memory to host memory when the GPU can't fit more data. The spilling mechanism is automatically triggered once the user-defined limit is reached, such limit can be set via the ``--device-memory-limit`` and ``device_memory_limit`` arguments for ``dask-cuda-worker`` and ``LocalCUDACluster``, respectively.
 
+Previously, spilling was disabled by default, but since Dask-CUDA 0.17 the default has been changed to ``0.8`` -- spilling will begin when Dask-CUDA device memory utilization reaches 80% of the device's total memory. The old behavior can still be set with ``--device-memory-limit=0`` or ``device_memory_limit=0``.
```
quasiben, line 16:
```suggestion
Previously, spilling was disabled by default, but since Dask-CUDA 0.17 the default has been changed to ``0.8`` -- spilling will begin when Dask-CUDA device memory utilization reaches 80% of the device's total memory.  Behavior can configured with ``--device-memory-limit`` flag.  Users can disable spilling by setting ``--device-memory-limit=0`` or ``device_memory_limit=0``.
```

quasiben:
Thanks @pentschev !

PR #441, opened 2020-11-11
Add notes on controlling number of workers to docs [skip ci]

quasiben:


docs/source/specializations.rst
```diff
@@ -8,6 +8,20 @@ Automatic Instantiation of One-Worker-Per-GPU
...
+
+Users can control the number of workers by explicitly defining ``CUDA_VISIBLE_DEVICES`` in either the ``dask-cuda-worker`` CLI or using ``LocalCUDACluster``.  For example, the following will launch 3 workers on devices, 1, 2, and 3:
+
+.. code-block:: bash
+
+    LocalCUDACluster(CUDA_VISIBLE_DEVICES='0,1,2')
+    or
+    CUDA_VISIBLE_DEVICES=0,1,2 dask-cuda-workewr
+
+Users can also use UUID of the device as an inputs to ``CUDA_VISIBLE_DEVICES``.  UUIDs should begin with prefix 'GPU-' or 'MIG-GPU': `GPU-9baca7f5-0f2f-01ac-6b05-8da14d6e9005`, for example.
```
pentschev, line 22: Could we add a link to the official docs https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#env-vars in here as well?

docs/source/specializations.rst
```diff
@@ -8,6 +8,20 @@ Automatic Instantiation of One-Worker-Per-GPU
 
 Using the ``dask-cuda-worker`` or ``LocalCUDACluster`` will automatically launch one worker for each GPU available on the node from where it was executed, avoiding the need for users to select GPUs in their application and thus reducing code complexity.
 
+Controlling Number of Workers
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Users can control the number of workers by explicitly defining ``CUDA_VISIBLE_DEVICES`` in either the ``dask-cuda-worker`` CLI or using ``LocalCUDACluster``.  For example, the following will launch 3 workers on devices, 1, 2, and 3:
```
pentschev, line 14: That sentence isn't correct, `CUDA_VISIBLE_DEVICES` is not defined in `dask-cuda-worker` CLI. `CUDA_VISIBLE_DEVICES` is an environment variable that will be read by `dask-cuda-worker`.

We can also say that `LocalCUDACluster` will also respect the `CUDA_VISIBLE_DEVICES` environment variable (when that's set before starting the `python` process), but if both the environment variable and the `CUDA_VISIBLE_DEVICES` kwarg are specified, the environment variable will be ignored in favor of the kwarg.

pentschev (changes requested):

docs/source/specializations.rst
quasiben: > Users can control the number of workers by explicitly defining environment variable ``CUDA_VISIBLE_DEVICES`` when using the ``dask-cuda-worker`` CLI.  When using ``LocalCUDACluster`` the environment variable or keyword argument ``CUDA_VISIBLE_DEVICES`` can be used.  If both keyword and env var are set, the env var will be ignored in favor of the kwarg.  For example, the following will launch 3 workers on devices, 1, 2, and 3:

Something like the above ?

pentschev: > Users can control the number of workers by explicitly defining environment variable `CUDA_VISIBLE_DEVICES` when using the `dask-cuda-worker` CLI.  When using `LocalCUDACluster` the environment variable `CUDA_VISIBLE_DEVICES` or keyword argument with same name can be used.  If both keyword and environment variables are set, the environment variable will be ignored in favor of the argument.  For example, the following will launch 3 workers on devices, 1, 2, and 3:

A little bit of reordering of your words. Also let's try to keep the abbreviations out as it makes it more difficult to find and for people less accustomed to them.

quasiben: Agreed about abbreviations. I'll make the changes now

pentschev (approved):
LGTM, thanks @quasiben !

pentschev:
rerun tests

pentschev:
rerun tests

pentschev:
rerun tests

PR #447, opened 2020-11-30
update builddocs dependency requirements

quasiben:
We are currently getting errors when building dask-cuda:

```
Running Sphinx v3.2.1

  File "/home/docs/checkouts/readthedocs.org/user_builds/dask-cuda/conda/latest/lib/python3.8/site-packages/dask/dataframe/__init__.py", line 2, in <module>
    from .core import (
  File "/home/docs/checkouts/readthedocs.org/user_builds/dask-cuda/conda/latest/lib/python3.8/site-packages/dask/dataframe/core.py", line 10, in <module>
    import pandas as pd
ModuleNotFoundError: No module named 'pandas'
```

This PR adds Pandas and dask requirements

I believe we will need @rapidsai/ops to approve/merge

### DIFF
--- a/conda/environments/builddocs_py37.yml
+++ b/conda/environments/builddocs_py37.yml
@@ -4,6 +4,4 @@ channels:
 - conda-forge
 dependencies:
 - rapids-doc-env
-- distributed
-- dask
 - pandas
--- a/conda/environments/builddocs_py37.yml
+++ b/conda/environments/builddocs_py37.yml
@@ -4,3 +4,6 @@ channels:
 - conda-forge
 dependencies:
 - rapids-doc-env
+- distributed
+- dask
+- pandas
--- a/conda/environments/builddocs_py37.yml
+++ b/conda/environments/builddocs_py37.yml
@@ -5,3 +5,4 @@ channels:
 dependencies:
 - rapids-doc-env
 - pandas
+- dask


conda/environments/builddocs_py37.yml
```diff
@@ -4,3 +4,6 @@ channels:
 - conda-forge
 dependencies:
 - rapids-doc-env
+- distributed
+- dask
+- pandas
```
pentschev, line 9:
```suggestion
- pandas
```

pentschev: We don't need `pandas` explicitly, it's a dependency of `dask`, see https://github.com/rapidsai/dask-cuda/blob/a27bd93aba8ad2d32003984ffa0c14fbf65e319f/conda/recipes/dask-cuda/meta.yaml#L26-L31 .

pentschev (changes requested):

conda/environments/builddocs_py37.yml
pentschev: An important note: `pandas` is a dependency of `dask`, but not `dask-core`, therefore we need the first one.

pentschev:
@quasiben why only need pandas, do we not need dask and distributed?

quasiben:
During the install it should pick up dask/distributed.  We could also change this to `dask[complete]` I think

jakirkham:
Conda doesn't support `...[...]`. `dask` already acts like `dask[complete]` for Conda

pentschev:
Regardless of `dask[complete]`, my question is whether do we want only `pandas` or `dask` (which includes `pandas`). I think it's ok-ish to only include `pandas`, but probably better to have `dask`, as `pandas` isn't itself a direct dependency in dask-cuda.

jakirkham:
Well the error also suggests this is coming from Dask and not Dask-CUDA. So yeah including `dask` makes sense.

pentschev:
I was going to say we should remove `pandas` because it isn't really required for dask-cuda, but then I saw it's actually explicitly imported in https://github.com/rapidsai/dask-cuda/blob/e77a3fe8532407016ad67e9120fc3fb91e580321/dask_cuda/tests/test_explicit_comms.py#L4, thus I think it's fine to keep it. Thanks @quasiben .

pentschev (approved):

pentschev:
@rapidsai/ops-codeowners can we get a review here?

ajschmidt8 (approved):

jakirkham:
Thanks all! 😄

Issue #336, opened 2020-07-02
Odd `__version__` reported


jakirkham:
Appears the version may not be working correctly here.

```python
In [1]: import dask_cuda                                                        
    
In [2]: dask_cuda.__version__                                                   
Out[2]: '0+unknown'
```

quasiben:
Seeing the same as well from conda and from pypi but not from source

pentschev:
I can confirm this is still an issue with conda packages, but installing from source works properly.

jakirkham:
I think PR ( https://github.com/rapidsai/dask-cuda/pull/448 ) should address this.

jakirkham:
With the new packages, this is now reported correctly.

```python
In [1]: import dask_cuda

In [2]: dask_cuda.__version__
Out[2]: '0.17.0a+46.ge77a3fe'
```

jakirkham:
Reopening as we had to revert part of PR ( https://github.com/rapidsai/dask-cuda/pull/448 ) with PR ( https://github.com/rapidsai/dask-cuda/pull/466 ) as it was breaking PyPI uploads. See issue ( https://github.com/pypa/warehouse/issues/8891 ) for more context as to why that is. Am reopening this as the issue presents (at least based on my local testing) with that reversion. Will need to figure out another way forward that doesn't breaking PyPI uploads and does include the version.

jakirkham:
This is still an issue. We tried to fix it, but the fix broke package building and uploading. So we had to revert. Need to investigate a new fix

pentschev:
Just for a quick update, this issue still remains.

jakirkham:
cc @wence-

PR #448, opened 2020-11-30
Fix versioneer

jakirkham:
Fixes https://github.com/rapidsai/dask-cuda/issues/336

* Adds missing `MANIFEST.in` to include versioneer components
* Always uses versioneer in `setup.py`

Based on local testing this should fix the issue. Also this matches with [versioneer's install instructions]( https://github.com/python-versioneer/python-versioneer/blob/master/INSTALL.md ).

cc @pentschev @quasiben

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,2 @@
+include versioneer.py
+include dask_cuda/_version.py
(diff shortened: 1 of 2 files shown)


quasiben (approved):
Thanks @jakirkham !  LGTM

pentschev (approved):
LGTM, thanks @jakirkham !

jakirkham:
Thanks all! 😄

PR #450, opened 2020-11-30
[REVIEW] FIX Move codecov upload to gpu build script

dillon-cullinan:
This moves a block from the Jenkins build into the build script. Simplifies the Jenkins builds slightly and fixes a very rare case where tests may error and not be properly caught due to EXITCODE intricacies.

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -120,6 +120,6 @@ else
 fi
 
 if [ -n "\${CODECOV_TOKEN}" ]; then
-    codecov -t \$CODECOV_TOKEN
+    codecov -t $CODECOV_TOKEN
 fi
 
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -119,7 +119,7 @@ else
     fi
 fi
 
-if [ -n "\${CODECOV_TOKEN}" ]; then
+if [ -n "${CODECOV_TOKEN}" ]; then
     codecov -t $CODECOV_TOKEN
 fi
 
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -118,3 +118,8 @@ else
         py.test --cache-clear -vs `python -c "import distributed.diagnostics.tests.test_nvml as m;print(m.__file__)"`
     fi
 fi
+
+if [ -n "\${CODECOV_TOKEN}" ]; then
+    codecov -t \$CODECOV_TOKEN
+fi
+


raydouglass (approved):

ci/gpu/build.sh
```diff
@@ -118,3 +118,8 @@ else
         py.test --cache-clear -vs `python -c "import distributed.diagnostics.tests.test_nvml as m;print(m.__file__)"`
     fi
 fi
+
+if [ -n "\${CODECOV_TOKEN}" ]; then
+    codecov -t \$CODECOV_TOKEN
```
pentschev, line 123: Is this really correct? My understand is that `\$` will escape the `$`, rather than get the variable's value. Is that really what's wanted or am I missing something?

jakirkham: We might also consider skipping the environment variable altogether. It seems `codecov` [checks this environment variable by default]( https://github.com/codecov/codecov-python/blob/713d94113e9b1189f227adc6131f04c65e4f967b/codecov/__init__.py#L327 ).

dillon-cullinan: Ah, you're correct @pentschev , this was copy pasted from a script in gpuCI that gets injected, so we were escaping characters to handle the injection.

The upload token is a secret text inside Jenkins, Jenkins will always hide it, no need to worry about that.

ci/gpu/build.sh
```diff
@@ -118,3 +118,8 @@ else
         py.test --cache-clear -vs `python -c "import distributed.diagnostics.tests.test_nvml as m;print(m.__file__)"`
     fi
 fi
+
+if [ -n "\${CODECOV_TOKEN}" ]; then
```
madsbk, line 122: I guess you have to _un-escape_ `$` here as well. The condition is always `True` otherwise.
```suggestion
if [ -n "${CODECOV_TOKEN}" ]; then
```

madsbk (changes requested):

madsbk (approved):
LGTM

madsbk:
@dillon-cullinan try merging with `branch-0.18`, it should fix the CI error

jakirkham:
rerun tests

(think CI uses a merge commit)

mike-wendt:
> (think CI uses a merge commit)
 
yeah we test the changes with the merge commit of the PR to make sure we have things up to date; that said I just changed the base to get it into 0.18

raydouglass:
@gpucibot merge

jakirkham:
Thanks all! 😀

PR #458, opened 2020-12-02
ProxyObject support of less-than, greater-than, equal, etc.

madsbk:
This PR adds support of less-than, greater-than, equal, etc.


pentschev (approved):
LGTM, thanks @madsbk !

PR #459, opened 2020-12-02
ProxyObject: support of pickle

madsbk:
This PR implements `ProxyObject.__reduce__()` in order to support pickle.

### DIFF
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -56,7 +56,7 @@ def device_deserialize(header, frames):
 
 @nvtx_annotate("SPILL_D2H", color="red", domain="dask_cuda")
 def device_to_host(obj: object) -> DeviceSerialized:
-    header, frames = serialize(obj, serializers=["dask", "pickle"])
+    header, frames = serialize(obj, serializers=["dask", "pickle"], on_error="raise")
     return DeviceSerialized(header, frames)
 
 
--- a/dask_cuda/proxy_object.py
+++ b/dask_cuda/proxy_object.py
@@ -1,6 +1,7 @@
 import operator
 import pickle
 import threading
+from collections import OrderedDict
 
 import dask
 import dask.dataframe.methods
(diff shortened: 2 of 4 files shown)


dask_cuda/proxy_object.py
```diff
@@ -155,15 +156,31 @@ def __init__(
...
+            "fixed_attr",
+            "type_serialized",
+            "typename",
+            "is_cuda_object",
+            "subclass",
+            "serializers",
+        ]
+        if include_obj:
+            args.insert(0, "obj")
+        return OrderedDict([(a, self._obj_pxy[a]) for a in args])
```
madsbk, line 183: Notice, even though `dict` is order as of Python 3.7, I think it make sense to use `OrderedDict` to make it explicit.

pentschev (approved):
I have a minor suggestion but other than that it looks good to me. In fact, that change is just a suggestion to make code a bit more understandable (at least I think it does), but feel free to ignore that either.

dask_cuda/proxy_object.py
```diff
@@ -155,15 +156,31 @@ def __init__(
...
+        args = [
+            "fixed_attr",
+            "type_serialized",
+            "typename",
+            "is_cuda_object",
+            "subclass",
+            "serializers",
+        ]
+        if include_obj:
+            args.insert(0, "obj")
```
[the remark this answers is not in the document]
madsbk: Good point fixed!

dask_cuda/device_host_file.py
```diff
@@ -56,7 +56,7 @@ def device_deserialize(header, frames):
 
 @nvtx_annotate("SPILL_D2H", color="red", domain="dask_cuda")
 def device_to_host(obj: object) -> DeviceSerialized:
-    header, frames = serialize(obj, serializers=["dask", "pickle"])
+    header, frames = serialize(obj, serializers=["dask", "pickle"], on_error="raise")
```
jakirkham, line 59: Thanks for catching this 😄

PR #462, opened 2020-12-02
Add note on RMM pool docs mentioning its per worker [skip ci]

pentschev:
I've recently seen some confusion on RMM pool being though as a cluster-wide size, this PR clarifies it's per worker.


quasiben (approved):
Thanks @pentschev

PR #477, opened 2020-12-16
CI: cleanup style check

madsbk:
This PR clean up the style check output by capturing both stdout and stderr. 
Also `isort` line-length is now 88, which is what `black` and `flake8` use.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,14 +1,14 @@
 repos:
       - repo: https://github.com/timothycrosley/isort
-        rev: 4.3.21
+        rev: 5.0.7
         hooks:
               - id: isort
       - repo: https://github.com/ambv/black
-        rev: stable
+        rev: 19.10b0
         hooks:
               - id: black
       - repo: https://gitlab.com/pycqa/flake8
-        rev: 3.7.7
+        rev: 3.8.3
         hooks:
               - id: flake8
 default_language_version:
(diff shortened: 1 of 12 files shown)


quasiben (approved):
Thanks @madsbk

jakirkham:
Thanks Mads! 😄

Tried to fix this a while back ( https://github.com/rapidsai/dask-cuda/pull/350 ), but ran into issues 😅 Guess the linter tools were too old

.pre-commit-config.yaml
```diff
@@ -1,14 +1,14 @@
...
         hooks:
               - id: isort
       - repo: https://github.com/ambv/black
-        rev: stable
+        rev: 19.10b0
         hooks:
               - id: black
       - repo: https://gitlab.com/pycqa/flake8
-        rev: 3.7.7
+        rev: 3.8.3
```
pentschev, line 11: Sorry for the late review here, but why did we pin those versions specifically instead of those from the integration repo (see https://github.com/rapidsai/integration/blob/branch-0.18/conda/recipes/versions.yaml) ?

jakirkham: I think some of these tools weren't respecting line length configuration values in some cases. Ran into similar issues as well. Though this does a raise a good point, would be good to propagate these version bumps into the integration repo.

pentschev: Ideas on what's the right approach? Pinning those exact versions in integration will probably lead all other RAPIDS repos to require some reformatting.

PR #480, opened 2020-12-30
Auto-label PRs based on their content [skip-ci]

jolorunyomi:
This PR adds the GitHub action [PR Labeler](https://github.com/actions/labeler) to auto-label PRs based on their content. 

Labeling is managed with a configuration file `.github/labeler.yml` using the following [options](https://github.com/actions/labeler#usage)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,25 +1,25 @@
 # https://github.com/actions/labeler#common-examples
-# Adapted from https://github.com/rapidsai/cudf/blob/main/.github/CODEOWNERS
+# Adapted from https://github.com/rapidsai/dask-cuda/blob/main/.github/CODEOWNERS
 # Labels culled from https://github.com/rapidsai/dask-cuda/labels
 
 Python:
- - python/**
- - notebooks/**
+ - 'dask_cuda/**'
+ - 'notebooks/**'
 
 libdaskcuda:
- - cpp/**
+ - 'cpp/**'
 
 CMake:
- - **/CMakeLists.txt
- - **/cmake/**
+ - '**/CMakeLists.txt'
+ - '**/cmake/**'
 
 Java:
- - java/**
+ - 'java/**'
 
 Ops:
- - .github/**
- - /ci/**
- - conda/**
- - **/Dockerfile
- - **/.dockerignore
- - docker/**
+ - '.github/**'
+ - 'ci/**'
+ - 'conda/**'
+ - '**/Dockerfile'
+ - '**/.dockerignore'
+ - 'docker/**'
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -5,13 +5,6 @@
 python:
  - 'dask_cuda/**'
 
-cpp:
- - 'cpp/**'
-
-CMake:
- - '**/CMakeLists.txt'
- - '**/cmake/**'
-
 ci:
   - 'ci/**'
 
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -5,6 +5,9 @@
 python:
  - 'dask_cuda/**'
 
+doc:
+ - 'docs/**'
+
 gpuCI:
   - 'ci/**'
 
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,25 @@
+# https://github.com/actions/labeler#common-examples
+# Adapted from https://github.com/rapidsai/cudf/blob/main/.github/CODEOWNERS
+# Labels culled from https://github.com/rapidsai/dask-cuda/labels
+
+Python:
+ - python/**
+ - notebooks/**
+
+libdaskcuda:
+ - cpp/**
+
+CMake:
+ - **/CMakeLists.txt
+ - **/cmake/**
+
+Java:
+ - java/**
+
+Ops:
+ - .github/**
+ - /ci/**
+ - conda/**
+ - **/Dockerfile
+ - **/.dockerignore
+ - docker/**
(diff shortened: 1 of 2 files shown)


ajschmidt8 (approved):

.github/labeler.yml
```diff
@@ -0,0 +1,25 @@
...
+
+libdaskcuda:
+ - cpp/**
+
+CMake:
+ - **/CMakeLists.txt
+ - **/cmake/**
+
+Java:
+ - java/**
```
jakirkham, line 17: Do we need these for this library? Asking as it seems unlikely we would add C++ or Java (and for that matter CMake) since the main goal is to build on top of Dask in Python.

.github/labeler.yml
```diff
@@ -0,0 +1,25 @@
+# https://github.com/actions/labeler#common-examples
+# Adapted from https://github.com/rapidsai/cudf/blob/main/.github/CODEOWNERS
+# Labels culled from https://github.com/rapidsai/dask-cuda/labels
+
+Python:
+ - python/**
```
jakirkham, line 6: This should probably just be `dask_cuda` since this just lives at the top level of the repo.

jakirkham (commented):
Made a few suggestions below. There are also a few files not handled like `.coveragerc`, `.readthedocs.yml`, `setup.py`, etc. I'm not sure how we want to handle these.

.github/labeler.yml
```diff
@@ -0,0 +1,25 @@
...
+Java:
+ - 'java/**'
+
+Ops:
+ - '.github/**'
+ - 'ci/**'
+ - 'conda/**'
+ - '**/Dockerfile'
+ - '**/.dockerignore'
+ - 'docker/**'
```
mike-wendt, line 25: This section should be removed

.github/labeler.yml
```diff
@@ -0,0 +1,25 @@
...
+
+libdaskcuda:
+ - 'cpp/**'
+
+CMake:
+ - '**/CMakeLists.txt'
+ - '**/cmake/**'
+
+Java:
+ - 'java/**'
```
mike-wendt, line 17: This section should match this repo's labels and the contents of the `.github/CODEOWNERS` file as a starting guide - PICs in this repo would then add or suggest more labels as needed

mike-wendt (changes requested):

.github/labeler.yml
```diff
@@ -0,0 +1,20 @@
+# https://github.com/actions/labeler#common-examples
+# Adapted from https://github.com/rapidsai/dask-cuda/blob/main/.github/CODEOWNERS
+# Labels culled from https://github.com/rapidsai/dask-cuda/labels
+
+Python:
```
ajschmidt8, line 5: This repo has a `python` label, but not a `Python` label.
```suggestion
python:
```

.github/labeler.yml
```diff
@@ -0,0 +1,20 @@
...
+ - 'dask_cuda/**'
+
+cpp:
```
ajschmidt8, line 8: This repo has no `cpp` label.

.github/labeler.yml
```diff
@@ -0,0 +1,20 @@
...
+ - 'cpp/**'
+
+CMake:
```
ajschmidt8, line 11: This repo has no `CMake` label.

.github/labeler.yml
```diff
@@ -0,0 +1,20 @@
...
+ - '**/CMakeLists.txt'
+ - '**/cmake/**'
+
+ci:
```
ajschmidt8, line 15: This repo has no `ci` or `gpuCI` label.

.github/labeler.yml
```diff
@@ -0,0 +1,20 @@
...
+  - 'ci/**'
+
+conda:
```
ajschmidt8, line 18: This repo has no `conda` label.

ajschmidt8 (changes requested):
Most of the labels listed in `labeler.yml` don't yet exist in the repo, so they will likely need to be added.

.github/labeler.yml
mike-wendt:
```suggestion
gpuCI:
```

I just added the missing label, so we just need to make this change

.github/labeler.yml
mike-wendt: I just added this label

mike-wendt (approved):

mike-wendt:
@ajschmidt8 can you re-review?
@jakirkham can you review for this repo?

jakirkham:
Sure. Also cc @pentschev in case you have thoughts here

.github/labeler.yml
```diff
@@ -0,0 +1,12 @@
+# https://github.com/actions/labeler#common-examples
+# Adapted from https://github.com/rapidsai/dask-cuda/blob/main/.github/CODEOWNERS
+# Labels culled from https://github.com/rapidsai/dask-cuda/labels
+
+python:
+ - 'dask_cuda/**'
```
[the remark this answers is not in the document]
pentschev: +1 to this

pentschev (commented):
I think adding `docs/` is good too, otherwise LGTM.

ajschmidt8 (approved):

.github/labeler.yml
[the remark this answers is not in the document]
mike-wendt: Added

mike-wendt:
@pentschev @jakirkham added `doc` label to `docs/` path

mike-wendt:
@gpucibot merge

Issue #472, opened 2020-12-07
Passing `--scheduler-file` will not raise error on wrong arguments


pentschev:
I just found out that for some reason passing wrong argument names to `dask-cuda-worker` will not fail when used with `--scheduler-file`, the following will start:

```
$ dask-cuda-worker --enable-tcp-over-ucx --enable-nvlink --enable-infiniband --ucx-net-devices=auto --scheduler-file scheduler.json
```

However, if you pass in the scheduler address it will fail, as the correct argument name is `--net-devices`:

```
$ dask-cuda-worker ucx://${SCHEDULER_IP}:8786 --enable-tcp-over-ucx --enable-nvlink --enable-infiniband --ucx-net-devices=auto
Usage: dask-cuda-worker [OPTIONS] [SCHEDULER] [PRELOAD_ARGV]...

Error: no such option: --ucx-net-devices=auto
```

pentschev:
After a bit of debugging it seems that the first incorrect option will be understood as being the `scheduler` argument even if it begins with `--`, thus if there's only one option that's incorrect it won't fail. This seems to be an intended consequence of `click` as it aims to be robust to options formatting and not just those starting with `--`.

PR #485, opened 2021-01-04
Fail if scheduler starts with '-' in dask-cuda-worker

pentschev:
Fixes https://github.com/rapidsai/dask-cuda/issues/472


madsbk (approved):
LGTM

PR #488, opened 2021-01-06
Extend CuPy benchmark with more operations

pentschev:


quasiben (commented):
Thanks @pentschev -- should I add this to the nightly runs ?

quasiben (approved):

pentschev:
> Thanks @pentschev -- should I add this to the nightly runs ?

I think that's not necessary for now, I'm mostly adding this as part of the PTDS work, I think it may be more useful once we have actual support for that in CuPy, RMM, etc.

jakirkham (approved):

PR #496, opened 2021-01-19
Shuffle benchmark

madsbk:
Implements a shuffle benchmark for both for explicit-comms and regular Dask

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -63,7 +63,7 @@ gpuci_conda_retry install "cudatoolkit=$CUDA_REL" \
 # `event_loop_policy`. See https://github.com/dask/distributed/pull/4212 .
 gpuci_conda_retry install "pytest-asyncio=<0.14.0"
 
-# https://docs.rapids.ai/maintainers/depmgmt/ 
+# https://docs.rapids.ai/maintainers/depmgmt/
 # gpuci_conda_retry remove -f rapids-build-env
 # gpuci_conda_retry install "your-pkg=1.0.0"
 
@@ -119,4 +119,8 @@ else
         py.test --cache-clear -vs `python -c "import distributed.tests.test_nanny as m;print(m.__file__)"`
         py.test --cache-clear -vs `python -c "import distributed.diagnostics.tests.test_nvml as m;print(m.__file__)"`
     fi
+
+    logger "Run local benchmark..."
+    python dask_cuda/benchmarks/local_cudf_shuffle.py --partition-size="1 KiB" -d 0  --runs 1 --backend dask
+    python dask_cuda/benchmarks/local_cudf_shuffle.py --partition-size="1 KiB" -d 0  --runs 1 --backend explicit-comms
 fi
--- a/dask_cuda/benchmarks/local_cudf_shuffle.py
+++ b/dask_cuda/benchmarks/local_cudf_shuffle.py
@@ -1,0 +1,257 @@
+from collections import defaultdict
+from time import perf_counter as clock
+from warnings import filterwarnings
+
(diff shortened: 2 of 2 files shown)


ci/gpu/build.sh
```diff
@@ -119,4 +119,8 @@ else
         py.test --cache-clear -vs `python -c "import distributed.tests.test_nanny as m;print(m.__file__)"`
         py.test --cache-clear -vs `python -c "import distributed.diagnostics.tests.test_nvml as m;print(m.__file__)"`
     fi
+
+    logger "Run local benchmark..."
+    python dask_cuda/benchmarks/local_cudf_shuffle.py --partition-size="1 KiB" -d 0  --runs 1 --backend dask
+    python dask_cuda/benchmarks/local_cudf_shuffle.py --partition-size="1 KiB" -d 0  --runs 1 --backend explicit-comms
```
pentschev, line 125: How long does that take? We don't run other benchmarks because they can take too long and aren't really tests. Maybe they would be more suitable for https://github.com/quasiben/dask-cuda-benchmarks ?

dask_cuda/benchmarks/local_cudf_shuffle.py
```diff
@@ -0,0 +1,243 @@
+from collections import defaultdict
```
pentschev, line 1: As a general note for this file since it overlaps a lot with https://github.com/rapidsai/dask-cuda/blob/branch-0.18/dask_cuda/benchmarks/local_cudf_merge.py, would it make sense to combine both? I did something similar for the CuPy benchmarks adding a new `--operation` option in https://github.com/rapidsai/dask-cuda/blob/7d27a0c320dcf093d0c51089c09223c17424d52a/dask_cuda/benchmarks/local_cupy.py#L211-L217 .

pentschev (commented):
Overall looks good, but I added a couple of comments to open some discussion.

ci/gpu/build.sh
madsbk: It only takes 4 second. It is just checking that the benchmark works

ajschmidt8 (approved):

dask_cuda/benchmarks/local_cudf_shuffle.py
madsbk: I thought about it but while profiling and experimenting I think it is best to have them as two individual files. 
Furthermore, I suspect that in the future merge will be using shuffle, which will make the merge benchmark less important.

pentschev: > I thought about it but while profiling and experimenting I think it is best to have them as two individual files.

Why? I think it will really make our lives easier to reduce code duplication.

madsbk: Sorry for not being clear. The thing is, the current version of the merge benchmark is fairly complex and I would like to simplify it before combining the two benchmarks. I am currently working on a redesign of explicit-comms that will make shuffle compatible with regular Dask and make it possible to write the merge benchmark as a regular Dask merge.

In short, I totally agree with your sentiment I just want to delay the unification a bit :)

pentschev: Got it, thanks for the details. At one point, I think we could even make the whole benchmark unified, creating cluster, etc., is something that's pretty much the same for all benchmarks, eventually I want to work on making that happen.

pentschev (approved):
LGTM, thanks @madsbk !

pentschev:
@gpucibot merge

madsbk:
Thanks @pentschev !

PR #500, opened 2021-01-25
Hostfile: register the removal of an existing key

madsbk:
This PR makes sure that hostfiles register the removal of an existing key. They didn't do that before when mixing CUDA and non-CUDA objects.

### DIFF
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -148,6 +148,10 @@ class DeviceHostFile(ZictBase):
         self.fast = self.host_buffer if memory_limit == 0 else self.host_buffer.fast
 
     def __setitem__(self, key, value):
+        if key in self.device_buffer:
+            # Make sure we register the removal of an existing key
+            del self[key]
+
         if is_device_object(value):
             self.device_keys.add(key)
             self.device_buffer[key] = value
--- a/dask_cuda/proxify_host_file.py
+++ b/dask_cuda/proxify_host_file.py
@@ -183,6 +183,10 @@ class ProxifyHostFile(MutableMapping):
 
     def __setitem__(self, key, value):
         with self.lock:
+            if key in self.store:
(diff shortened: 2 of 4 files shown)


dask_cuda/device_host_file.py
```diff
@@ -148,6 +148,10 @@ def __init__(
         self.fast = self.host_buffer if memory_limit == 0 else self.host_buffer.fast
 
     def __setitem__(self, key, value):
+        if key in self.device_buffer:
+            # Make sure we register the removal of an existing key
+            del self[key]
```
pentschev, line 153: I'm not sure I understand this, what is the case when this is necessary? If the key already exists in `self.device_buffer` but it now refers to a non-CUDA object that's going to be stored in `self.host_buffer`, is that right? If so, does this actually happen in practice?

madsbk:
Yes, in this case:
```python 
    dhf["x"] = b
    dhf["x"] = a
    assert set(dhf.device.keys()) == set()
    assert set(dhf.host.keys()) == set(["x"])
```
No, in the case of `DeviceHostFile` and in the current Dask/Distributed implementation, I don't think it is likely to happen in practice. 
In `ProxifyHostFile` it is way more likely to happen and I added the check in `DeviceHostFile` for robustness.

dask_cuda/device_host_file.py
madsbk: No, in the case of `DeviceHostFile` and in the current Dask/Distributed implementation, I don't think it is likely to happen in practice. 
In `ProxifyHostFile` it is way more likely to happen and I added the check in `DeviceHostFile` for robustness.

pentschev: I see, thanks for confirming.

pentschev (approved):
LGTM, thanks @madsbk .

pentschev:
@gpucibot merge

PR #503, opened 2021-01-29
ProxyObject: ignore initial fixed attribute errors

madsbk:
Fixes https://github.com/rapidsai/gpu-bdb/issues/174 by ignoring errors when initializing fixed attributes. Also adds tests of fixed attributes.

#### Rationale
When initialing a `ProxyObject` we probe the *proxied object* in order to retain a copy of the result of accessing fixed attributes like `name` and `__len__()`. Exceptions trigger by this probing is acceptable and can be safely ignored.

### DIFF
--- a/dask_cuda/proxy_object.py
+++ b/dask_cuda/proxy_object.py
@@ -55,7 +55,7 @@ def asproxy(obj, serializers=None, subclass=None) -> "ProxyObject":
                 if callable(val):
                     val = val()
                 fixed_attr[attr] = val
-            except Exception:
+            except (AttributeError, TypeError):
                 pass
 
         if subclass is None:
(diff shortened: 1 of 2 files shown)


dask_cuda/proxy_object.py
```diff
@@ -55,7 +55,7 @@ def asproxy(obj, serializers=None, subclass=None) -> "ProxyObject":
                 if callable(val):
                     val = val()
                 fixed_attr[attr] = val
-            except AttributeError:
+            except Exception:
```
pentschev, line 58: Would it make more sense to have `except (AttributeError, Type)`, i.e., the only cases being tested? I find it hard to debug this kind of catch all exceptions, and if there's a need to make something more specific for different exception types this can be very inconvenient in the future.

pentschev (commented):
I have a small question/suggestion, but otherwise LGTM.

dask_cuda/proxy_object.py
madsbk: Good point, in principle any error could be acceptable but keeping it to `AttributeError` and `TypeError` for now makes future debugging easier.

pentschev (approved):
Thanks @madsbk , everything LGTM!

pentschev:
@gpucibot merge

PR #507, opened 2021-01-29
Test broadcast merge in local_cudf_merge benchmark

rjzamora:
Includes the necessary changes to test [dask#7143](https://github.com/dask/dask/pull/7143).  More specifically, this adds the following options:

- `--base-chunks` : Number of base-DataFrame partitions (default: n_workers)
- `--other-chunks` : Number of other-DataFrame partitions (default: n_workers)
- `--broadcast-join` : Use broadcast join when possible
- `--shuffle-join` : Use shuffle join (takes precedence over '--broadcast-join')


quasiben:
run tests

rjzamora:
Now that [dask#7143](https://github.com/dask/dask/pull/7143) was merged, it is now safe to consider this PR for integration.

pentschev:
Retargeting seems to have no conflicts, let's see what the tests say. 🙂

jakirkham:
Could you please run `black` locally Rick and push the changes? Currently failing the style job. Idk if the others will run without that

pentschev:
>  Idk if the others will run without that

They still run, so we can wait another hour or so to see the results, but feel free to restyle it anyway in the meantime Rick.

jakirkham:
Ah ok. If they still run, great let's wait

Know that previously changelog (ofc no longer included) and style were required to pass before running tests. Given this, couldn't tell if the jobs here were actually queued to run or maybe we were running into some edge case in gpuCI where they only appear queued (but don't actually run).

dask_cuda/benchmarks/local_cudf_merge.py
```diff
@@ -258,6 +270,10 @@ def main(args):
         for (w1, w2), nb in total_nbytes.items()
     }
 
+    broadcast = (
+        "false" if args.shuffle_join else ("true" if args.broadcast_join else "default")
```
pentschev, line 274:
```suggestion
        False if args.shuffle_join else (True if args.broadcast_join else "default")
```

pentschev: Unless I'm misunderstanding the reason for using strings here, this will ensure the summary at the end is standardized, we have also `rmm-pool` that is using actual booleans. If there's a reason to use string, I'd prefer `"False"` and `"True"`, so as to mimic Python's conversion to string and thus still make the output look alike.

pentschev (commented):
I left a minor nitpick, but other than that LGTM. Thanks @rjzamora !

pentschev (approved):
Thanks @rjzamora !

jakirkham:
Looks like we are seeing some failures in `test_dataframe_merge` (subprocess is exiting with a non-zero exit code)

rjzamora:
>Looks like we are seeing some failures in test_dataframe_merge (subprocess is exiting with a non-zero exit code)

Hmm - It doesn't seem like that test is running the benchmark modified in this PR, is it?

jakirkham:
Was this from a change in Dask or Distributed then? Both of those are planning to be released tomorrow. So if they have issues, we should try to identify those quickly

jakirkham:
Just to update this thread, Rick did find a recent Dask change that is causing failures here as mentioned in this comment ( https://github.com/dask/dask/pull/7305#issuecomment-791129729 ) in the PR with the change. We have surfaced this in the Dask release issue ( https://github.com/dask/community/issues/129#issuecomment-791133624 ). Trying to figure out now how best to address this before the Dask + Distributed release tomorrow

pentschev:
I'm attempting to fix the issue mentioned above in https://github.com/dask/dask/pull/7325 .

jakirkham:
rerun tests

jakirkham:
@gpucibot merge

PR #511, opened 2021-02-04
Fixing fixed-attribute-proxy-object-test

madsbk:
This PR fixes the current CI error by using `SimpleNamespace()` when testing the proxy object

### DIFF
--- a/dask_cuda/tests/test_proxy.py
+++ b/dask_cuda/tests/test_proxy.py
@@ -1,5 +1,6 @@
 import operator
 import pickle
+from types import SimpleNamespace
 
 import pandas
 import pytest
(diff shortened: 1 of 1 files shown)


pentschev (commented):
Looks good to me, but I'm not sure I understand why adding a new test is fixing existing ones, is there a global state somewhere?

jakirkham:
FWIW was able to reproduce [the test failure seen originally on CI]( https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/dask-cuda/job/prb/job/dask-cuda-gpu-test/CUDA=10.1,GPU_LABEL=gpu-t4%7C%7Cgpu,OS=ubuntu16.04,PYTHON=3.7/214/consoleFull ) with NumPy alone. Have filed as issue ( https://github.com/numpy/numpy/issues/18325 ). Though there were a lot of type changes in NumPy 1.20.0. So this may intentionally no longer work

pentschev:
Ok, but the change here seems to add a completely new test with no relevant changes to existing ones. Why is that fixing the test that was failing? Is the proxy object being stored and shared among multiple tests?

madsbk:
> Ok, but the change here seems to add a completely new test with no relevant changes to existing ones. Why is that fixing the test that was failing? Is the proxy object being stored and shared among multiple tests?

My bad, I forgot to remove the name access test of numpy dtypes :)

pentschev (approved):
Thanks @madsbk for the fix!

pentschev:
@gpucibot merge

Issue #517, opened 2021-02-08
PTDS Benchmarks


quasiben:
With https://github.com/cupy/cupy/pull/4322 merged in we can begin profiling/benchmarking performance of more workers per GPU.  

## Setup 
- Build latest cupy `python -m pip install .`
- CUPY_CUDA_PER_THREAD_DEFAULT_STREAM=1 dask-cuda-worker --nthreads=2
- Use at least 4 GPUs
- test `nthreads`: 1, 2, 4, 8, 32`

## Test Operations

- SUM
- MEAN
- SVD
- Matrix Multiply
- Array Slicing 

Note @pentschev previously tested these operations on a single GPU: 
- https://github.com/pentschev/pybench/blob/89d65a6c418a1fee39d447bd11b8a999835b74a9/pybench/benchmarks/benchmark_array.py
- https://github.com/pentschev/pybench/blob/master/examples/plot_array_example.ipynb

## Example of CuPy with Dask

```python
import cupy
import dask.array as da

rs = da.random.RandomState(RandomState=cupy.random.RandomState) 
x = rs.normal(10, 1, size=(500000, 500000), chunks=(10000, 10000))
x.sum().compute()
```

cc @charlesbluca

jakirkham:
We've also used adding a matrix with it's transpose as a communication benchmark. Could be reasonable here to show the communication overhead goes to 0 while still computing work in parallel

quasiben:
That's a good idea @jakirkham .  I updated the test operations

charlesbluca:
Thanks for writing this up @quasiben!

Some questions:
- Would this benchmark be better suited as a reproducible script, a one-off notebook with some plots to reflect the results, or both? Seems like it might be leaning towards notebook since the number of GPUs is flexible.
- Could pybench be applicable in any way towards putting this together, or does it serve as more of an example in implementing the tests?

quasiben:
> Would this benchmark be better suited as a reproducible script, a one-off notebook with some plots to reflect the results, or both? Seems like it might be leaning towards notebook since the number of GPUs is flexible.

Generally, yes!  Some of these operations are already built in here:
https://github.com/rapidsai/dask-cuda/blob/branch-0.18/dask_cuda/benchmarks/local_cupy.py

> Could pybench be applicable in any way towards putting this together, or does it serve as more of an example in implementing the tests?

[...]

PR #524, opened 2021-02-12
Add operations/arguments to local CuPy array benchmark

charlesbluca:
This PR adds the following operations to the local CuPy array benchmark:

- sum
- mean
- array slicing

This also adds an additional special argument, `--benchmark-json`, which takes an optional path to dump the results of the benchmark in JSON format. This would allow us to generate plots using the output, as discussed in #517.

Some thoughts:

- Should there be an additional argument to specify the array slicing interval (which is currently fixed at 3)?
- Could the JSON output be cleaned up? Currently, a (truncated) sample output file looks like:

```json
{
  "operation": "transpose_sum",
  "size": 10000,
  "second_size": 1000,
  "chunk_size": 2500,
  "compute_size": [
    10000,
    10000
  ],
  "compute_chunk_size": [
    2500,
    2500
  ],
  "ignore_size": "1.05 MB",
  "protocol": "tcp",
  "devs": "0,1,2,3",
  "threads_per_worker": 1,
  "times": [
    {
      "wall_clock": 1.4910394318867475,
      "npartitions": 16
    }
  ],
  "bandwidths": {
    "(00,01)": {
      "25%": "136.34 MB/s",
      "50%": "156.67 MB/s",
      "75%": "163.32 MB/s",
      "total_nbytes": "150.00 MB"
    }
  }
}
```

### DIFF
--- a/dask_cuda/benchmarks/local_cupy.py
+++ b/dask_cuda/benchmarks/local_cupy.py
@@ -211,8 +211,22 @@ async def run(args):
                     "protocol": args.protocol,
                     "devs": args.devs,
                     "threads_per_worker": args.threads_per_worker,
-                    "times": took_list,
-                    "bandwiths": sorted(bandwidths.items()),
+                    "times": [
+                        {"wall_clock": took, "npartitions": npartitions}
+                        for (took, npartitions) in took_list
+                    ],
+                    "bandwidths": {
+                        f"({d1},{d2})"
+                        if args.multi_node or args.sched_addr
+                        else "(%02d,%02d)"
+                        % (d1, d2): {
+                            "25%": bw[0],
+                            "50%": bw[1],
+                            "75%": bw[2],
+                            "total_nbytes": total_nbytes[(d1, d2)],
+                        }
+                        for (d1, d2), bw in sorted(bandwidths.items())
+                    },
                 }
                 with open(args.benchmark_json, "w") as fp:
                     json.dump(d, fp, indent=2)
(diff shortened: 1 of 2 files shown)


dask_cuda/benchmarks/local_cupy.py
```diff
@@ -179,6 +197,40 @@ async def run(args):
                 )
                 print(fmt % (d1, d2, bw[0], bw[1], bw[2], total_nbytes[(d1, d2)]))
 
+            if args.benchmark_json:
+                import json
```
madsbk, line 201: Moving `import json` to the top should be fine

madsbk (approved):
Looks good to me, nice addition @charlesbluca.

> Should there be an additional argument to specify the array slicing interval (which is currently fixed at 3)?

I think fixing it to 3 is fine for now. If we need to benchmark the impact of step sizes, we can always add it. At that point, we might consider using the first position argument as a _command_ and use the following arguments  as command-specific (e.g. like how `git` works)

>  Could the JSON output be cleaned up? Currently, a (truncated) sample output file looks like:

Looks fine to me :)

dask_cuda/benchmarks/local_cupy.py
```diff
@@ -60,6 +60,24 @@ async def _run(client, args):
...
+        x = rs.random((args.size, args.size), chunks=args.chunk_size).persist()
+        await wait(x)
+        func_args = (x,)
+
+        func = lambda x: x.mean()
+    elif args.operation == "slice":
+        x = rs.random((args.size, args.size), chunks=args.chunk_size).persist()
+        await wait(x)
+        func_args = (x,)
+
```
madsbk, line 79:
```suggestion

```

dask_cuda/benchmarks/local_cupy.py
charlesbluca: Done! Also going to change this to `from json import dump`.

pentschev (approved):
LGTM as well, thanks @charlesbluca !

pentschev:
@gpucibot merge

PR #326, opened 2020-06-26
Update SSHCluster usage in benchmarks with new CUDAWorker

pentschev:
Updates usage of `SSHCluster` according to changes in https://github.com/dask/distributed/pull/5191.

PR #525, opened 2021-02-12
Add GHA to mark issues/prs as stale/rotten

Ethyling:
Issues and PRs without activity for 30d will be marked as stale.
If there is no activity for 90d, they will be marked as rotten.

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -1,0 +1,65 @@
+name: Mark stale and rotten issues and pull requests
+
+on:
+  schedule:
+    - cron: "0 * * * *"
+
+jobs:
+  mark-stale-issues:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Mark Issues as Stale
+        uses: actions/stale@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          stale-issue-message: >
+            This issue has been marked stale due to no recent activity in the past 30d.
+            Please close this issue if no further response or action is needed.
+            Otherwise, please respond with a comment indicating any updates or changes to the original issue and/or confirm this issue still needs to be addressed.
+            This issue will be marked rotten if there is no activity in the next 60d.
+          stale-issue-label: "stale"
+          days-before-issue-stale: 30
+          days-before-issue-close: -1
+  mark-stale-prs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Mark PRs as Stale
+        uses: actions/stale@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          stale-pr-message: >
+            This PR has been marked stale due to no recent activity in the past 30d.
+            Please close this PR if it is no longer required.
+            Otherwise, please respond with a comment indicating any updates.
+            This PR will be marked rotten if there is no activity in the next 60d.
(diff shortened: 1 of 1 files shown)


pentschev:
Is that the case for all PRs or will it depend on other information, such as labels? For example, I have recently marked https://github.com/rapidsai/dask-cuda/pull/402 and https://github.com/rapidsai/dask-cuda/pull/326, which are both blocked by external factors, but I wouldn't like either of those to be marked as rotten.

Ethyling:
> Is that the case for all PRs or will it depend on other information, such as labels? For example, I have recently marked #402 and #326, which are both blocked by external factors, but I wouldn't like either of those to be marked as rotten.

Currently it's for all PRs. I think we should keep it this way to be sure we do not leave PRs as "blocked" despite the blocking condition is resolved. You will just need to comment the PR to remove the `stale` label every 30 days. WDYT?

pentschev:
I wouldn't mind doing that every 90 days or so, but these PRs may be there easily for 6+ months, as is the case of https://github.com/rapidsai/dask-cuda/pull/326 , where we're blocked by a Dask feature that's currently on hold. Doing so every 30 days feels a bit annoying, that would mean we will end up having to do this roughly twice for every RAPIDS release.

pentschev:
Alternatively, could this changed only PRs that aren't marked as "Draft"? I think that would be a reasonable middle ground.

ajschmidt8 (approved):

Ethyling:
Hey @pentschev, I understand your point, but the idea here is to be able to track issues and PRs without activity. Ignoring some PRs/issues would be contrary to the objective. We want to be sure we are not letting any issue or PR open without reason and so by commenting every month, we are sure the blocking condition is evaluated again. WDYT?

pentschev:
Honestly, I think this is going to be a big annoyance, I would prefer if we could have a way to add exceptions to some PRs, even if it's just a label or something that would allow us a longer grace period for those marked in such a way.

Ethyling:
Maybe we can create a specific label to ignore some PRs/issues
@mike-wendt WDYT?

mike-wendt:
@pentschev @madsbk I understand your position and agree it will add a few minutes to update these PRs monthly even to say `still blocked`. That said, this PR is a part of a RAPIDS wide update to ensure we're reviewing issues/PRs in a more timely manner, and enabling the PM team to track and respond to these issues/PRs. The 30 day `stale` window is to ensure all issues/PRs are reviewed _each release_, so we keep the relevant ones around and close those that are not being worked on. The 90 day `rotten` window should only exist at the start of implementing this across all RAPIDS repos as going forward we should be addressing these more quickly than 90 days. The ultimate goal is so the repos and the PM team can quickly target the oldest issues/PRs first and then keep the others updated to minimize the number of `stale` labels. 

Feedback from @aravenel and @fondaing is that an issue/PR with several `still blocked` comments has much more context and value versus no updates in months on an issue/PR with a label `0 - Blocked`. Given their feedback we should move forward with the current settings in this PR. Alex said he will follow up with the PICs next month after we have this implemented and are tracking the number of `stale` & `rotten` issues/PRs to see if adjustments need to be made. For now, let's start with these settings in use by other repos and meet with the PM team next month about adjustments.

Ethyling:
Using this PR to resolve auto-merger issue: https://github.com/rapidsai/dask-cuda/pull/528

PR #527, opened 2021-02-16
ProxifyHostFile: tracking of external objects

madsbk:
This PR implements tracking of external objects. The idea is that tasks can register objects that should be exposed to spilling when `ProxifyHostFile`  is running out of memory.

Also, this PR makes the long running explicit-comms tasks [`local_shuffle()`](https://github.com/rapidsai/dask-cuda/blob/a574ccb53aac6d2be961e90db8a4f61f7fc80b20/dask_cuda/explicit_comms/dataframe/shuffle.py#L106) use this new feature, which means it shouldn't require more GPU memory than the regular shuffle implementation in Dask.

### DIFF
--- a/dask_cuda/proxify_device_objects.py
+++ b/dask_cuda/proxify_device_objects.py
@@ -41,6 +41,10 @@ def proxify(obj, proxied_id_to_proxy, found_proxies, subclass=None):
     _id = id(obj)
     if _id in proxied_id_to_proxy:
         ret = proxied_id_to_proxy[_id]
+        finalize = ret._obj_pxy.get("external_finalize", None)
+        if finalize:
+            finalize()
+            proxied_id_to_proxy[_id] = ret = asproxy(obj, subclass=subclass)
     else:
         proxied_id_to_proxy[_id] = ret = asproxy(obj, subclass=subclass)
     found_proxies.append(ret)
@@ -67,6 +71,14 @@ def proxify_device_object_proxy_object(obj, proxied_id_to_proxy, found_proxies):
         if _id in proxied_id_to_proxy:
             obj = proxied_id_to_proxy[_id]
         else:
+            proxied_id_to_proxy[_id] = obj
+
+    finalize = obj._obj_pxy.get("external_finalize", None)
+    if finalize:
+        finalize()
+        obj = obj._obj_pxy_copy()
+        if not obj._obj_pxy_is_serialized():
+            _id = id(obj._obj_pxy["obj"])
             proxied_id_to_proxy[_id] = obj
 
     found_proxies.append(obj)
--- a/dask_cuda/proxify_device_objects.py
+++ b/dask_cuda/proxify_device_objects.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List
+from typing import Any, List, MutableMapping
 
 from dask.utils import Dispatch
 
@@ -9,7 +9,7 @@ dispatch = Dispatch(name="proxify_device_objects")
 
 def proxify_device_objects(
     obj: Any,
-    proxied_id_to_proxy: Dict[int, ProxyObject],
+    proxied_id_to_proxy: MutableMapping[int, ProxyObject],
     found_proxies: List[ProxyObject],
 ):
     """ Wrap device objects in ProxyObject
@@ -22,7 +22,7 @@ def proxify_device_objects(
     ----------
     obj: Any
         Object to search through or wrap in a ProxyObject.
-    proxied_id_to_proxy: Dict[int, ProxyObject]
+    proxied_id_to_proxy: MutableMapping[int, ProxyObject]
         Dict mapping the id() of proxied objects (CUDA device objects) to
         their proxy and is updated with all new proxied objects found in `obj`.
     found_proxies: List[ProxyObject]
--- a/dask_cuda/proxify_device_objects.py
+++ b/dask_cuda/proxify_device_objects.py
@@ -56,7 +56,6 @@ def proxify_device_object_default(obj, proxied_id_to_proxy, found_proxies):
 
 @dispatch.register(ProxyObject)
 def proxify_device_object_proxy_object(obj, proxied_id_to_proxy, found_proxies):
-
     # We deserialize CUDA-serialized objects since it is very cheap and
     # makes it easy to administrate device memory usage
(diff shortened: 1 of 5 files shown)


madsbk:
@beckernick @VibhuJawa, it would be great if one of you have time try this PR. You need to set `DASK_JIT_UNSPILL=True` and `DASK_EXPLICIT_COMMS=True`.

dask_cuda/proxify_device_objects.py
```diff
@@ -22,41 +23,50 @@ def proxify_device_objects(
...
+    return dispatch(obj, proxied_id_to_proxy, found_proxies, excl_proxies)
 
 
 def proxify(obj, proxied_id_to_proxy, found_proxies, subclass=None):
     _id = id(obj)
     if _id in proxied_id_to_proxy:
         ret = proxied_id_to_proxy[_id]
+        finalize = ret._obj_pxy.get("external_finalize", None)
+        if finalize:
+            finalize()
```
pentschev, line 49: Any chance `finalize` can be non-callable? Should the condition be `if callable(finalize):` perhaps and emit an error when is `not callable(finalize) and finalize is not None`?

dask_cuda/proxify_device_objects.py
```diff
@@ -70,21 +80,39 @@ def proxify_device_object_proxy_object(obj, proxied_id_to_proxy, found_proxies):
         else:
             proxied_id_to_proxy[_id] = obj
 
-    found_proxies.append(obj)
+    finalize = obj._obj_pxy.get("external_finalize", None)
+    if finalize:
+        finalize()
```
pentschev, line 85: Same question as above to callable/non-callable `finalize` here.

dask_cuda/proxify_host_file.py
```diff
@@ -181,13 +181,38 @@ def get_access_info(self) -> Tuple[int, List[Tuple[int, int, List[ProxyObject]]]
...
 
+    def add_external(self, obj):
+        # Notice, since `self.store` isn't modified, no lock is needed
+        found_proxies: List[ProxyObject] = []
+        proxied_id_to_proxy = {}
+        # Notice, we are excluding found objects that are already proxies
+        ret = proxify_device_objects(
+            obj, proxied_id_to_proxy, found_proxies, excl_proxies=True
+        )
+        last_access = time.time()
```
pentschev, line 192: I meant to ask about this more broadly on the proxy object effort, is the value of `time.time()` local to the current worker or also used by remote workers? If only local to the current worker, could we resort to `time.monotonic()` instead? `time.monotonic()` guarantees not going backwards, whereas `time.time()` does not, and that can be problematic in some systems.

If `time.time()` is used by remote workers, we may have a bigger problem, what if one of the workers' system time is completely messed up, what would happen then?

dask_cuda/proxy_object.py
```diff
@@ -199,6 +199,12 @@ def __init__(
         self._obj_pxy_lock = threading.RLock()
         self._obj_pxy_cache = {}
 
+    def __del__(self):
+        """In order to call `external_finalize()` ASAP, we call it here"""
+        external_finalize = self._obj_pxy.get("external_finalize", None)
+        if external_finalize is not None:
+            external_finalize()
```
pentschev, line 206: Same here, could it happen to be non-callable?

dask_cuda/tests/test_proxify_host_file.py
```diff
@@ -129,3 +129,59 @@ def test_cudf_get_device_memory_objects():
     ]
     res = get_device_memory_objects(objects)
     assert len(res) == 4, "We expect four buffer objects"
+
+
+def test_externals():
```
pentschev, line 134: The loose use of `itemsize` in the tests here seems error-prone. Unless of course I'm misunderstanding the real meaning of `itemsize`, wouldn't it be safer to do something analogous to the snippet below?

```python
dtype = np.float64
itemsize = dtype(1).itemsize
dhf = ProxifyHostFile(device_memory_limit=itemsize)
dhf["k1"] = cupy.arange(1, dtype=dtype) + 1
```

pentschev (commented):
I have some questions here, but overall LGTM.

dask_cuda/proxify_device_objects.py
madsbk: The `finalize` [is always callable](https://docs.python.org/3/library/weakref.html#weakref.finalize) if it isn't `None`.

dask_cuda/proxify_host_file.py
madsbk: Good point, we are using `last_access` to spill the _oldest_ object first so this could be an issue in principle.
Fixed.

dask_cuda/tests/test_proxify_host_file.py
madsbk: To make it more clear, I am now defining:
```python 
cupy = pytest.importorskip("cupy")
cupy.cuda.set_allocator(None)
one_item_array = lambda: cupy.arange(1)
one_item_nbytes = one_item_array().nbytes
```

madsbk:
Thanks for the review @pentschev, I think I have address all your concerns.

dask_cuda/proxify_device_objects.py
pentschev: It should always be callable, I'm asking if there's any chance it `external_finalize` could eventually be something other than a callable and whether it's worth checking that. This is probably me just being overly concerned, so I thought I would check with you.

madsbk: I think that is too paranoid :) 
`_obj_pxy["external_finalize"]` is only assigned one place: https://github.com/rapidsai/dask-cuda/blob/d8f371572dd38997d909acef2d51a406a23f4824/dask_cuda/proxify_host_file.py#L219

pentschev: Fair enough.

pentschev (approved):
LGTM, thanks @madsbk !

pentschev:
If there's no rush, I would wait another day to see if @beckernick or @VibhuJawa have a chance to try out the PR, otherwise it can be merged any time IMO.

madsbk:
Since explicit-comms's shuffle is disabled by default, I think it is safe to merge.

@gpucibot merge

madsbk:
@gpucibot merge

pentschev:
Thanks @madsbk !

Issue #536, opened 2021-02-18
[PTDS] Different memory allocation/free behavior with increased threads per worker


charlesbluca:
I am profiling FFT operations using 2 and 4 threads per worker with PTDS enabled (#517), and am interested in the behavior surrounding CUDA API calls to `cudaFree`. My code:

```python
import cupy
import dask.array as da
import nvtx
import rmm
from dask.distributed import Client, wait
from dask_cuda import LocalCUDACluster

def main():

    cluster = LocalCUDACluster(n_workers=2, threads_per_worker=2, rmm_pool_size="15GB") # threads_per_worker=4
    client = Client(cluster)
    client.run(cupy.cuda.set_allocator, rmm.rmm_cupy_allocator)
    
    rs = da.random.RandomState(RandomState=cupy.random.RandomState)
    x = rs.random(size=(10000, 10000), chunks=(1000, 10000))

    with nvtx.annotate("fft op", color="purple"):
        da.fft.fft(x).compute()

if __name__ == "__main__":
    main()
```

`nsys` profile (focusing on streams and CUDA API calls) for 2 threads per worker:

![image](https://user-images.githubusercontent.com/20627856/108415238-48f9ea00-71fb-11eb-9885-2aefd4f96f50.png)

4 threads per worker:

![image](https://user-images.githubusercontent.com/20627856/108415314-616a0480-71fb-11eb-921b-a3a1c6fa48e6.png)

Some questions:

- Why is `cuModuleLoadData` (gold) called by N-1 of the threads' default streams during the big `cudaFree` call, and why does this take much longer with 4 threads than 2?
- Regardless of thread count, there are some additional small memory allocations/frees that are called before the script terminates - does this have something to do with clean up? (some screenshots of this happening with 4 threads)

![image](https://user-images.githubusercontent.com/20627856/108416258-72ffdc00-71fc-11eb-9050-8fc47988cfbe.png)

![image](https://user-images.githubusercontent.com/20627856/108416291-7eeb9e00-71fc-11eb-9a81-a337c31eb219.png)

cc @quasiben

pentschev:
`cudaFree(0)` is a common operation to be run at the beginning of an application to ensure the CUDA context is initialized, and that is a normal but time consuming operation. You can ignore that beginning because it's going to be part of every CUDA application.

[...]

PR #537, opened 2021-02-19
Replace compute() calls on CuPy benchmarks by persist()

pentschev:
Based on the discussion from https://github.com/rapidsai/dask-cuda/issues/536, the CuPy benchmarks were using `compute()` instead of `persist()`, thus causing the data to be sent back to the client, and our intent is not to benchmark that step, this PR fixes that behavior.

I've targeted branch-0.19 for now because of the code freeze, it isn't 100% critical IMO to have this in 0.18, but certainly I'd prefer that. WDYT @quasiben @jakirkham ?

### DIFF
--- a/dask_cuda/benchmarks/local_cupy.py
+++ b/dask_cuda/benchmarks/local_cupy.py
@@ -87,12 +87,11 @@ async def _run(client, args):
     if args.profile is not None:
         async with performance_report(filename=args.profile):
             t1 = clock()
-            await client.compute(func(*func_args))
+            await wait(client.persist(func(*func_args)))
             took = clock() - t1
     else:
         t1 = clock()
-        res = client.compute(func(*func_args))
-        await client.gather(res)
+        await wait(client.persist(func(*func_args)))
         if args.type == "gpu":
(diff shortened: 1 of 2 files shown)


jakirkham (approved):

jakirkham:
Sounds good. Thanks Peter! 😄

Looks like there is a conflict that we might need to fix first

pentschev:
> Sounds good. Thanks Peter! 😄

What sounds good, the fix or targeting branch-0.19?

> Looks like there is a conflict that we might need to fix first

Indeed, thanks for the heads up, should be fixed now! 🙂

pentschev:
I'm merging this, since all tests pass after resolving conflicts. Thanks @jakirkham for the review!

pentschev:
@gpucibot merge

jakirkham:
> > Sounds good. Thanks Peter! 😄
> 
> What sounds good, the fix or targeting branch-0.19?

Both 😄 Sorry for being unclear

> > Looks like there is a conflict that we might need to fix first
> 
> Indeed, thanks for the heads up, should be fixed now! 🙂

Thanks Peter!

PR #322, opened 2020-06-23
WIP: Add option to enable RMM logging

jakirkham:
Includes an option to enable RMM logging for Dask-CUDA workers.

cc @rlratzel @dillon-cullinan @shwina

PR #542, opened 2021-02-26
Add option to enable RMM logging

charlesbluca:
Continuation of #322; this PR adds:

- Keyword argument `rmm_log_directory` to `LocalCUDACluster` and `CUDAWorker` to enable per-worker RMM logging to a specified directory
- Option `--rmm-log-directory` to the `dask-cuda-worker` CLI to achieve the same as above
- Option `--rmm-log-directory` to the benchmarks to enable worker *and* scheduler RMM logging
- Tests to `test_local_cuda_cluster.py` and `test_dask_cuda_worker.py` to check that logging is happening

The log files use the following naming convention:

- `rmm_log_<N>.dev0.txt` for workers spawned through the construction of a cluster
- `rmm_log_<IP>:<PORT>.dev0.txt` for workers spawned using the CLI
- `rmm_log_scheduler.dev0.txt` for schedulers

Some questions:

- Should we leave `dev0` in the log file names? It seems redundant, but I didn't really put much time into seeing if we could stop RMM from appending that to the file names.
- Are the tests sufficient? Right now they are only checking that the memory resources are `rmm.mr.LoggingResourceAdaptors` when logging is enabled but a check that the files were actually generated could be nice.
- Should we consider adding logging to the scheduler/client?

cc @jakirkham @pentschev

### DIFF
--- a/dask_cuda/benchmarks/utils.py
+++ b/dask_cuda/benchmarks/utils.py
@@ -48,7 +48,7 @@ def parse_benchmark_args(description="Generic dask-cuda Benchmark", args_list=[]
         default=None,
         type=str,
         help="Directory to write worker and scheduler RMM log files to. "
-        "Has no effect if RMM memory pool is disabled.",
+        "Logging is only enabled if RMM memory pool is enabled.",
     )
     parser.add_argument(
         "--all-to-all", action="store_true", help="Run all-to-all before computation",
--- a/dask_cuda/benchmarks/utils.py
+++ b/dask_cuda/benchmarks/utils.py
@@ -44,6 +44,14 @@ def parse_benchmark_args(description="Generic dask-cuda Benchmark", args_list=[]
         "--disable-rmm-pool", action="store_true", help="Disable the RMM memory pool"
     )
     parser.add_argument(
+        "--rmm-log-directory",
+        default=None,
+        type=str,
+        help="Directory to write per-worker RMM log files to. If users "
+        "want logging from the client or scheduler, that must be set "
+        "separately. Has no effect if RMM memory pool is disabled.",
+    )
+    parser.add_argument(
         "--all-to-all", action="store_true", help="Run all-to-all before computation",
     )
     parser.add_argument(
@@ -222,14 +230,36 @@ def get_scheduler_workers(dask_scheduler=None):
     return dask_scheduler.workers
 
 
-def setup_memory_pool(pool_size=None, disable_pool=False):
+def setup_memory_pool(
+    dask_worker=None, pool_size=None, disable_pool=False, log_directory=None,
+):
     import cupy
 
     import rmm
 
+    logging = log_directory is not None
+
     if not disable_pool:
         rmm.reinitialize(
-            pool_allocator=True, devices=0, initial_pool_size=pool_size,
+            pool_allocator=True,
+            devices=0,
+            initial_pool_size=pool_size,
+            logging=logging,
+            log_file_name=os.path.join(
+                log_directory,
+                "rmm_log_%s.txt"
+                % (
+                    (
+                        dask_worker.name.split("/")[-1]
+                        if isinstance(dask_worker.name, str)
+                        else dask_worker.name
+                    )
+                    if hasattr(dask_worker, "name")
+                    else "scheduler"
+                ),
+            )
+            if logging
+            else None,
         )
         cupy.cuda.set_allocator(rmm.rmm_cupy_allocator)
 
--- a/dask_cuda/benchmarks/utils.py
+++ b/dask_cuda/benchmarks/utils.py
@@ -47,9 +47,8 @@ def parse_benchmark_args(description="Generic dask-cuda Benchmark", args_list=[]
         "--rmm-log-directory",
         default=None,
         type=str,
-        help="Directory to write per-worker RMM log files to. If users "
-        "want logging from the client or scheduler, that must be set "
-        "separately. Has no effect if RMM memory pool is disabled.",
+        help="Directory to write worker and scheduler RMM log files to. "
+        "Has no effect if RMM memory pool is disabled.",
     )
     parser.add_argument(
         "--all-to-all", action="store_true", help="Run all-to-all before computation",
--- a/dask_cuda/benchmarks/utils.py
+++ b/dask_cuda/benchmarks/utils.py
@@ -236,6 +236,8 @@ def setup_memory_pool(
 
     import rmm
 
+    from dask_cuda.utils import get_rmm_log_file_name
+
     logging = log_directory is not None
 
     if not disable_pool:
@@ -244,21 +246,7 @@ def setup_memory_pool(
             devices=0,
             initial_pool_size=pool_size,
             logging=logging,
-            log_file_name=os.path.join(
-                log_directory,
-                "rmm_log_%s.txt"
-                % (
-                    (
-                        dask_worker.name.split("/")[-1]
-                        if isinstance(dask_worker.name, str)
-                        else dask_worker.name
-                    )
-                    if hasattr(dask_worker, "name")
-                    else "scheduler"
-                ),
-            )
-            if logging
-            else None,
+            log_file_name=get_rmm_log_file_name(dask_worker, logging, log_directory),
         )
         cupy.cuda.set_allocator(rmm.rmm_cupy_allocator)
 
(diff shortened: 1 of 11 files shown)


dask_cuda/benchmarks/utils.py
```diff
@@ -222,14 +229,36 @@ def get_scheduler_workers(dask_scheduler=None):
...
+                        dask_worker.name.split("/")[-1]
+                        if isinstance(dask_worker.name, str)
+                        else dask_worker.name
+                    )
+                    if hasattr(dask_worker, "name")
+                    else "scheduler"
+                ),
+            )
+            if logging
+            else None,
```
pentschev, line 261: Could we move this into a function in https://github.com/rapidsai/dask-cuda/blob/branch-0.19/dask_cuda/utils.py ? I see the same is used again below, so better to centralize that.

dask_cuda/cli/dask_cuda_worker.py
```diff
@@ -112,6 +112,15 @@
...
     "trying to enable both will result in an exception.",
 )
+@click.option(
+    "--rmm-log-directory",
+    default=None,
+    help="Directory to write per-worker RMM log files to. If users "
+    "want logging from the client or scheduler, that must be set "
+    "separately."
+    "NOTE: Has no effect if --rmm-pool-size is not specified and "
+    "--rmm-managed-memory is not enabled.",
```
pentschev, line 122: I don't think that's true, RMM logs are valid independent of the underlying memory management resource underneath.

dask_cuda/local_cuda_cluster.py
```diff
@@ -90,11 +90,15 @@ class LocalCUDACluster(LocalCluster):
...
-        If True, initialize each worker with RMM and set it to use managed
-        memory. If False, RMM may still be used if `rmm_pool_size` is specified,
+        If ``True``, initialize each worker with RMM and set it to use managed
+        memory. If ``False``, RMM may still be used if `rmm_pool_size` is specified,
         but in that case with default (non-managed) memory type.
         WARNING: managed memory is currently incompatible with NVLink, trying
         to enable both will result in an exception.
+    rmm_log_directory: str
+        Directory to write per-worker RMM log files to. If users want logging
+        from the client or scheduler, that must be set separately. Has no effect
```
pentschev, line 100: This doesn't apply here, users can't set client or scheduler logs unless we specify a way to do so here. I think they're anyway not so important, so I would just rephrase to:

pentschev, line 100:
```suggestion
        Directory to write per-worker RMM log files to, the client and
        scheduler are not logged here. Has no effect
```

pentschev: And again, I don't think the `rmm_pool_size`/`rmm_managed_memory` statement is true here either.

pentschev (changes requested):
Overall looks good @charlesbluca , I've made a few requests and wrote a couple of questions, please take a look when you have a chance.

dask_cuda/cli/dask_cuda_worker.py
```diff
@@ -112,6 +112,15 @@
     "WARNING: managed memory is currently incompatible with NVLink, "
     "trying to enable both will result in an exception.",
 )
+@click.option(
+    "--rmm-log-directory",
+    default=None,
```
quasiben, line 117: When using a Pool and RMM is set to None will there be an error or will RMM not log ?

pentschev: Right now if `--disable-rmm-pool` is explicitly passed, there will be no logging. Do you have a preference on what should happen in that case @quasiben ?

dask_cuda/cli/dask_cuda_worker.py
charlesbluca: Should we allow RMM logging regardless of those decisions then? Right now, logging is enabled through `RMMSetup`, and only if `self.nbytes is not None or self.managed_memory is True`.

dask_cuda/benchmarks/utils.py
charlesbluca: Sure!

dask_cuda/cli/dask_cuda_worker.py
pentschev: Ah ok, I see what you're referring to. The sentence then may be a bit misleading to dumber people like me (unclear if "no" or "and" has precedence). I would rephrase it to:
"NOTE: Logging will only be enabled if --rmm-pool-size or --rmm-managed-memory are specified."

charlesbluca: Okay - out of interest, is there any utility to having logging on if we don't use a pool or managed memory?

pentschev: Theoretically, yes. Libraries such as cuDF use RMM by default, even if it doesn't use a pool or managed memory. However, we generally want people to use a pool, so I'm not sure whether logging for all cases is something we want to incentivize. I'm ok either way, but my personal opinion is we don't need it now.

dask_cuda/cli/dask_cuda_worker.py
charlesbluca: Is this referring to the `dask-cuda-worker` CLI or the benchmark options? In both cases, if `--rmm-log-directory` is set to None, then `logging` will be set to False in their respective RMM setups and it will be disabled.

Like @pentschev said, if `--disable-rmm-pool` is passed in the benchmarks the worker/scheduler's calls to `rmm.reintialize` will be skipped altogether and logging will be disabled. This relates somewhat to the conversation we had above - is there any utility to enabling RMM logging even if we aren't using a memory pool or managed memory?

jakirkham: Is it possible to pass `--rmm-log-directory` without a directory? If so, what happens?

charlesbluca: Nice catch! If it's the last argument provided, it gives an error:

```
>>> dask-cuda-worker localhost:8786 --rmm-log-directory
Error: --rmm-log-directory option requires an argument
```

However, if it isn't the last argument, the following argument is interpreted as the directory to write to:

```
>>> dask-cuda-worker localhost:8786 --rmm-log-directory --no-dashboard
...
>>> ls -- --no-dashboard/
rmm_log_127.0.0.1:34325.dev0.txt  rmm_log_127.0.0.1:35957.dev0.txt 
```

Is there anything we can do to check for this?

jakirkham: Yeah it would be good if it could error consistently. I'm sure there is a way for click to do this, but I don't know offhand what is

charlesbluca: The closest I could find is the [handling of option-like arguments](https://click.palletsprojects.com/en/7.x/arguments/#option-like-arguments), but I think that's limited to `click.arguments`; setting `ignore_unknown_options` to False in the command context didn't change anything here.

If we do find a general solution for this, it would probably be good to open up a separate PR applying it throughout the CLI, as it looks like most of the options behave in a similar fashion; they just usually lead to errors later on unrelated to click.

pentschev: > Like @pentschev said, if --disable-rmm-pool is passed in the benchmarks the worker/scheduler's calls to rmm.reintialize will be skipped altogether and logging will be disabled. This relates somewhat to the conversation we had above - is there any utility to enabling RMM logging even if we aren't using a memory pool or managed memory?

Somehow I thought this comment was for the benchmarks only. Regardless, I think @charlesbluca has already clarified.

> The closest I could find is the [handling of option-like arguments](https://click.palletsprojects.com/en/7.x/arguments/#option-like-arguments), but I think that's limited to `click.arguments`; setting `ignore_unknown_options` to False in the command context didn't change anything here.
> 
> If we do find a general solution for this, it would probably be good to open up a separate PR applying it throughout the CLI, as it looks like most of the options behave in a similar fashion; they just usually lead to errors later on unrelated to click.

Even though @jakirkham raises a good point, I wouldn't worry too much about it in this PR, as this is also an issue with all other arguments that require a value and somehow we didn't have too many issues with it. I think we could simply open an issue to eventually address that but not worry about it in this PR.

pentschev (approved):
Thanks for addressing the questions I raised @charlesbluca , LGTM now.

quasiben (approved):

quasiben:
@gpucibot merge

Issue #544, opened 2021-03-04
Improve UCX documentation and samples


pentschev:
We currently have some documentation on using UCX with Dask-CUDA available in https://dask-cuda.readthedocs.io/en/latest/ucx.html . However, I feel that documentation is a bit convoluted, so reviewing and improving its text is a good idea. Furthermore, it could be easier for readers if they have a small set of samples they can refer to inside the dask-cuda repository, perhaps under a new directory `samples/ucx`, where we could have a few simple scripts to run a single-node cluster with `LocalCUDACluster`, as well as one or more scripts to run `dask-scheduler` + `dask-cuda-worker` (possibly more than just one, for multi-node clusters) + client code.

@charlesbluca is this something you would be interested in doing? There's no rush on it, I just thought someone like you would be perfect for it, as you still don't have the bias on writing all the options by heart. 🙂

charlesbluca:
Happy to take this on! I don't have as much experience setting up UCX through the CLI, but we can use this issue to discuss that when the time comes.

charlesbluca:
Just now looking into this! A question on the UCX configuation - are the listed environment variables being processed through `dask.config`? If so, maybe the "Configuration" section could be cleaned up a bit by adding these options to the [Dask configuration reference](https://docs.dask.org/en/latest/configuration-reference.html), which we could then link to for more complicated UCX options. Then maybe for the mandatory settings, we could make a `ucx.yaml` configuration that includes them all?

pentschev:
> Just now looking into this! A question on the UCX configuation - are the listed environment variables being processed through `dask.config`? If so, maybe the "Configuration" section could be cleaned up a bit by adding these options to the [Dask configuration reference](https://docs.dask.org/en/latest/configuration-reference.html), which we could then link to for more complicated UCX options.

I agree that the configuration reference page could be improved with descriptions for the sections that say "No Comment". However, I think we still should list them in our docs because it's _very easy_ to introduce typos, for example note that there is a double-underscore after `DASK_UCX`.

> Then maybe for the mandatory settings, we could make a `ucx.yaml` configuration that includes them all?

[...]

PR #545, opened 2021-03-05
Improve UCX documentation and examples

charlesbluca:
Addressing #544, this PR aims to clarify the requirements, configuration, and usage of UCX with Dask-CUDA.

Still a lot to be done:

- [x] Flesh out hardware/software requirements
- [x] Rework CLI/Python usage examples
- [x] Clarify some uncertainties in the Configuration section
- [ ] Add standalone examples of UCX usage

### DIFF
--- a/dask_cuda/initialize.py
+++ b/dask_cuda/initialize.py
@@ -64,12 +64,18 @@ def initialize(
         Set environment variables to enable UCX RDMA connection manager support,
         implies ``enable_infiniband=True``.
         Default is ``False``.
-    net_devices: str
-        Explicit interface name, such as ``"ib0"`` for InfiniBand or ``"eth0"``
-        if InfiniBand is disabled.
+    net_devices: callable or str
+        If callable, the function must take exactly one argument (the index of
+        current GPU) that will be used to get the interface name, such as
+        ``lambda dev: "mlx5_%d:1" % (dev // 2)``, which would return
+        ``"mlx5_1:1"`` for GPU 3.
+        If a string, must be an explicit interface name, such as ``"ib0"``
+        for InfiniBand or ``"eth0"`` if InfiniBand is disabled.
         Default is ``""``, which will result in all available devices being used.
     cuda_device_index: None or int
-        What does this argument control?
+        Index of the current GPU, which will be supplied to ``net_devices`` if
+        it is callable.
+        Default is ``None``.
     """
 
     if create_cuda_context:
--- a/dask_cuda/initialize.py
+++ b/dask_cuda/initialize.py
@@ -1,28 +1,3 @@
-"""
-This initialization scripts will create CUDA context and initialize UCX-Py,
-depending on user parameters.
-
-It is sometimes convenient to initialize the CUDA context, particularly before
-starting up Dask workers which create a variety of threads.
-
-To ensure UCX works correctly, it is important to ensure it is initialized with
-the correct options. This is important for scheduler, workers and client. This
-initialization script will ensure that based on the flags and options passed by
-the user.
-
-This module is intended to be used within a Worker preload script.
-https://docs.dask.org/en/latest/setup/custom-startup.html
-
-You can add it to your global config with the following yaml
-
-    distributed:
-      worker:
-        preload:
-          - dask_cuda.initialize_ucx
-
-See https://docs.dask.org/en/latest/configuration.html for more information
-about Dask configuration.
-"""
 import logging
 
 import click
@@ -44,6 +19,59 @@ def initialize(
     net_devices="",
     cuda_device_index=None,
 ):
+    """Create CUDA context and initialize UCX-Py, depending on user parameters.
+
+    Sometimes it is convenient to initialize the CUDA context, particularly before
+    starting up Dask workers which create a variety of threads.
+
+    To ensure UCX works correctly, it is important to ensure it is initialized with
+    the correct options. This is important for scheduler, workers, and client. This
+    function will ensure that based on the flags and options passed by the user.
+
+    This function can also be used within a Worker preload script.
+    https://docs.dask.org/en/latest/setup/custom-startup.html
+
+    You can add it to your global config with the following YAML:
+
+    .. code-block:: yaml
+
+        distributed:
+          worker:
+            preload:
+              - dask_cuda.initialize
+
+    See https://docs.dask.org/en/latest/configuration.html for more information
+    about Dask configuration.
+
+    Parameters
+    ----------
+    create_cuda_context: bool
+        Create CUDA context on initialization.
+        Default is ``True``.
+    enable_tcp_over_ucx: bool
+        Set environment variables to enable TCP over UCX, even if InfiniBand
+        and NVLink are not supported or disabled.
+        Default is ``False``.
+    enable_infiniband: bool
+        Set environment variables to enable UCX InfiniBand support, implies
+        ``enable_tcp_over_ucx=True``.
+        Default is ``False``.
+    enable_nvlink: bool
+        Set environment variables to enable UCX NVLink support, implies
+        ``enable_tcp_over_ucx=True``.
+        Default is ``False``.
+    enable_rdmacm: bool
+        Set environment variables to enable UCX RDMA connection manager support,
+        implies ``enable_infiniband=True``.
+        Default is ``False``.
+    net_devices: str
+        Explicit interface name, such as ``"ib0"`` for InfiniBand or ``"eth0"``
+        if InfiniBand is disabled.
+        Default is ``""``, which will result in all available devices being used.
+    cuda_device_index: None or int
+        What does this argument control?
+    """
+
     if create_cuda_context:
         try:
             numba.cuda.current_context()
--- a/dask_cuda/initialize.py
+++ b/dask_cuda/initialize.py
@@ -25,9 +25,9 @@ def initialize(
     starting up Dask workers which create a variety of threads.
 
     To ensure UCX works correctly, it is important to ensure it is initialized with
-    the correct options. This is especially important for the scheduler and client,
-    which cannot be configured to use UCX with arguments like ``LocalCUDACluster``
-    and ``dask-cuda-worker``. This function will ensure that they are provided a UCX
+    the correct options. This is especially important for the client, which cannot be
+    configured to use UCX with arguments like ``LocalCUDACluster`` and 
+    ``dask-cuda-worker``. This function will ensure that they are provided a UCX
     configuration based on the flags and options passed by the user.
 
     This function can also be used within a worker preload script for UCX configuration
--- a/dask_cuda/initialize.py
+++ b/dask_cuda/initialize.py
@@ -25,10 +25,13 @@ def initialize(
     starting up Dask workers which create a variety of threads.
 
     To ensure UCX works correctly, it is important to ensure it is initialized with
-    the correct options. This is important for scheduler, workers, and client. This
-    function will ensure that based on the flags and options passed by the user.
+    the correct options. This is especially important for the scheduler and client,
+    which cannot be configured to use UCX with arguments like ``LocalCUDACluster``
+    and ``dask-cuda-worker``. This function will ensure that they are provided a UCX
+    configuration based on the flags and options passed by the user.
 
-    This function can also be used within a Worker preload script.
+    This function can also be used within a worker preload script for UCX configuration
+    of mainline Dask/Distributed.
     https://docs.dask.org/en/latest/setup/custom-startup.html
 
     You can add it to your global config with the following YAML:
--- a/dask_cuda/initialize.py
+++ b/dask_cuda/initialize.py
@@ -26,7 +26,7 @@ def initialize(
 
     To ensure UCX works correctly, it is important to ensure it is initialized with
     the correct options. This is especially important for the client, which cannot be
-    configured to use UCX with arguments like ``LocalCUDACluster`` and 
+    configured to use UCX with arguments like ``LocalCUDACluster`` and
     ``dask-cuda-worker``. This function will ensure that they are provided a UCX
     configuration based on the flags and options passed by the user.
 
--- a/docs/source/ucx.rst
+++ b/docs/source/ucx.rst
@@ -2,12 +2,7 @@ UCX Integration
 ===============
 
 Communication can be a major bottleneck in distributed systems.
-Dask-CUDA addresses this by supporting integration with `UCX <https://www.openucx.org/>`_, an optimized communication framework that provides high-performance networking and supports a variety of transport methods, including:
-
-- NVLink for *something ubiquitous to NVLink systems*
-- InfiniBand for *something ubiquitous to InfiniBand systems*
-- TCP for systems that do not have specialized hardware
-
+Dask-CUDA addresses this by supporting integration with `UCX <https://www.openucx.org/>`_, an optimized communication framework that provides high-performance networking and supports a variety of transport methods, including `NVLink <https://www.nvidia.com/en-us/data-center/nvlink/>`_ and `Infiniband <https://www.mellanox.com/pdf/whitepapers/IB_Intro_WP_190.pdf>`_ for systems with specialized hardware, and TCP for systems without it.
 This integration is enabled through `UCX-Py <https://ucx-py.readthedocs.io/>`_, an interface that provides Python bindings for UCX.
 
 
@@ -17,15 +12,17 @@ Requirements
 Hardware
 ^^^^^^^^
 
-*Go into more detail on what hardware is required for NVLink, InfiniBand support*
+To use UCX with NVLink or InfiniBand, relevant GPUs must be connected with NVLink bridges or NVIDIA Mellanox InfiniBand Adapters, respectively.
+NVIDIA provides comparison charts for both `NVLink bridges <https://www.nvidia.com/en-us/design-visualization/nvlink-bridges/>`_ and `InfiniBand adapters <https://www.nvidia.com/en-us/networking/infiniband-adapters/>`_.
 
 Software
 ^^^^^^^^
 
-*Does UCX integration require anything else other than UCX-Py?*
+UCX integration requires an environment with both UCX and UCX-Py installed; see `UCX-Py Installation <https://ucx-py.readthedocs.io/en/latest/install.html>`_ for detailed instructions on this process.
 
-When using UCX, each NVLink and InfiniBand memory buffer must create a mapping between each unique pair of processes they are transferred across; this can be quite costly, taking up to 100 ms per mapping.
+When using UCX, each NVLink and InfiniBand memory buffer must create a mapping between each unique pair of processes they are transferred across; this can be quite costly, potentially in the range of hundreds of milliseconds per mapping.
 For this reason, it is strongly recommended to use `RAPIDS Memory Manager (RMM) <https://github.com/rapidsai/rmm>`_ to allocate a memory pool that is only prone to a single mapping operation, which all subsequent transfers may rely upon.
+A memory pool also prevents the Dask scheduler from deserializing CUDA data, which will cause a crash.
 
 Configuration
 ^^^^^^^^^^^^^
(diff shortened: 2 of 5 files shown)


charlesbluca:
Thinking that the [UCX usage sections](https://dask-cuda.readthedocs.io/en/latest/ucx.html#launching-scheduler-workers-and-clients-separately) could probably be simplified a bit if we better fleshed out `dask-cuda-worker` and `dask_cuda.initialize(...)` in their own pages.

docs/source/ucx.rst
```diff
@@ -1,35 +1,87 @@
...
+
+    Replaces ``sockcm`` with ``rdmacm`` in ``UCX_TLS`` and ``UCX_SOCKADDR_TLS_PRIORITY``, *enabling remote direct memory access (RDMA) for connection management.*
+    This is recommended by UCX for use with InfiniBand, and will have no effect if InfiniBand tranfers are disabled.
+
+- ``ucx.net-devices: <str>`` -- recommended.
+
+    Explicitly sets ``UCX_NET_DEVICES`` instead of defaulting to ``"all"``, which can result in suboptimal performance.
+    If using InfiniBand, set to the desired IB device, e.g. ``"mlx5_0:1"``.
+    If InfiniBand is disabled, set to the ethernet device, e.g. ``"enp1s0f0"`` on a DGX-1.
+    All available UCX-compatible devices can be listed by running ``ucx_info -d`` or ``ifconfig`` *(are all the options under ifconfig compatible)*.
```
pentschev, line 66: I don't think all options under `ifconfig` are valid, for example `veth*` (often used by Docker) aren't. `ucx_info -d` only should be more reliable.

Also, interestingly the most important use case for us is missing in this section (it was missing before your proposed changes here too), which is the `"auto"` option, as documented in https://github.com/rapidsai/dask-cuda/blob/e72d776645ee3ee7e1672f73c9c043731cf401d5/dask_cuda/cli/dask_cuda_worker.py#L187-L203 and https://github.com/rapidsai/dask-cuda/blob/e72d776645ee3ee7e1672f73c9c043731cf401d5/dask_cuda/local_cuda_cluster.py#L72-L86 . Could you add that here as well? It's documented in "Important Notes" in this file below, but it seems sensible to have it here too.

docs/source/ucx.rst
```diff
@@ -1,35 +1,87 @@
...
+- ``rmm.pool-size: <str | int>`` -- recommended.
+
+    Allocates an RMM pool of the specified size for the process; size can be provided with an integer number of bytes or in human readable format, e.g. ``"4GB"``.
+    In addition to reducing the cost of mapping incurred by memory transfers, a pool can prevent the Dask scheduler from deserializing CUDA data and causing a crash.
+    It is recommended to set the pool size to at least the minimum amount of memory used by the process; if possible, one can map all GPU memory to a single pool, to be utilized for the lifetime of the process.
+
+
+.. note::
+    These options can also be used with mainline Dask/Distributed.
+    However, this may disable a variety of features, such as *put some features exclusive to Dask-CUDA here*.
```
pentschev, line 77: `net-devices="auto"` is an exclusive feature of Dask-CUDA, I don't think there are others that apply to this section. The line below this should also be clarifying in this context.

pentschev (commented):
I added some comments here @charlesbluca , but it's looking good so far, thanks for working on this!

docs/source/ucx.rst
```diff
@@ -1,35 +1,87 @@
 UCX Integration
 ===============
 
-Communication can be a major bottleneck in distributed systems, and that's no different in Dask and Dask-CUDA. To address that, Dask-CUDA provides integration with `UCX-Py <https://ucx-py.readthedocs.io/>`_, a Python interface for the `UCX <https://www.openucx.org/>`_ communication framework.  UCX is a low-level library that provides high-performance networking and supports several transports, including NVLink and InfiniBand for systems that have such specialized hardware, as well as TCP for those that do not.
+Communication can be a major bottleneck in distributed systems.
+Dask-CUDA addresses this by supporting integration with `UCX <https://www.openucx.org/>`_, an optimized communication framework that provides high-performance networking and supports a variety of transport methods, including:
 
+- NVLink for *something ubiquitous to NVLink systems*
+- InfiniBand for *something ubiquitous to InfiniBand systems*
```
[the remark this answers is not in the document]
charlesbluca: Sure! More details would probably be better suited for a hardware requirements section.

docs/source/ucx.rst
```diff
@@ -1,35 +1,87 @@
...
-    These options can also be used with mainline Dask/Distributed outside of Dask-CUDA, however, we recommend using Dask-CUDA when possible. See the :doc:`Specializations for GPU Usage <specializations>` page for details on the benefits of using Dask-CUDA.
+Hardware
+^^^^^^^^
+
+*Go into more detail on what hardware is required for NVLink, InfiniBand support*
+
+Software
+^^^^^^^^
+
+*Does UCX integration require anything else other than UCX-Py?*
```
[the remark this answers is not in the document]
charlesbluca: Thanks for the link! I will also go into some detail about when you would want to use the Conda package vs building/installing from source (essentially whether or not you plan to use IB/RDMA).

docs/source/ucx.rst
```diff
@@ -1,35 +1,87 @@
...
+
+- ``ucx.cuda_copy: true`` -- **required.**
+    
+    Adds ``cuda_copy`` to ``UCX_TLS``, enabling *all* transfers over UCX.
+    *Is this accurate to say?*
+
+- ``ucx.tcp: true`` -- **required.**
+
+    Adds ``tcp`` to ``UCX_TLS``, enabling TCP transfers over UCX if NVLink or InfiniBand are unavailable or disabled.
+    *Why is this required if they are available and enabled?*
```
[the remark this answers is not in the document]
charlesbluca: Would it maybe be better to use that for the reasoning to enable it in this case?

docs/source/ucx.rst
```diff
@@ -1,35 +1,87 @@
...
-- ``DASK_UCX__NET_DEVICES=mlx5_0:1`` (default: ``None``, causes UCX to decide what device to use, possibly being suboptimal, implies ``UCX_NET_DEVICES=all``): this is very important when ``DASK_UCX__INFINIBAND=True`` to ensure the scheduler is connected over the InfiniBand interface. When ``DASK_UCX__INFINIBAND=False`` it's recommended to use the ethernet device instead, e.g., ``DASK_UCX__NET_DEVICES=enp1s0f0`` on a DGX-1.
-- ``DASK_RMM__POOL_SIZE=1GB``: allocates an RMM pool for the process. In some circumstances, the Dask scheduler will deserialize CUDA data and cause a crash if there's no pool.
+    Adds ``rc`` to ``UCX_TLS``, enabling InfiniBand transfers over UCX; affects inter-node communication only.
+
+
+- ``ucx.rdmacm: true`` -- recommended for InfiniBand.
+
+    Replaces ``sockcm`` with ``rdmacm`` in ``UCX_TLS`` and ``UCX_SOCKADDR_TLS_PRIORITY``, *enabling remote direct memory access (RDMA) for connection management.*
+    This is recommended by UCX for use with InfiniBand, and will have no effect if InfiniBand tranfers are disabled.
```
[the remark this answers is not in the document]
charlesbluca: Is that to say that unexpected behavior will occur if `ucx.rdmacm` is enabled but `ucx.infiniband` is not?

docs/source/ucx.rst
charlesbluca: It does make sense to include information about `"auto"` here. Are there any resources on installing UCX-Py with hwloc support? Didn't see anything explicit about it in the installation guide.

docs/source/ucx.rst
charlesbluca: Yeah, so far from the config section alone I have:

- Use of CUDA transfers for task on GPU
- Setting `net-devices="auto"`

Anything I'm missing here?

docs/source/ucx.rst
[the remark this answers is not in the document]
pentschev: I think that information is better suited to the UCX-Py docs link I mentioned above. Let's try to avoid duplicating information which will eventually become inconsistent between the two docs.

docs/source/ucx.rst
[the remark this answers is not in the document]
pentschev: You can write a brief comment, but we probably don't want to rant about it here for too long as it's hard to draw a line of how much detail we want to go in. Here again a long rant would probably suit better UCX-Py (not suggesting we need to do it).

docs/source/ucx.rst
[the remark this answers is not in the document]
pentschev: Not unexpected behavior, the behavior is very much expected, resulting in an error on initialization.

docs/source/ucx.rst
pentschev: It's in the build dependencies: https://ucx-py.readthedocs.io/en/latest/install.html#build-dependencies

docs/source/ucx.rst
pentschev: > Use of CUDA transfers for task on GPU

This is again not true. A user can very much use mainline Dask with GPUs, they will only not benefit from the specializations Dask-CUDA provides -- thus my comment on the line that is immediately below this clarifying that difference. Therefore, the only thing I can think of here is indeed `net-devices="auto"`.

docs/source/ucx.rst
charlesbluca: Question about specifying an interface when InfiniBand is disabled:

I notice later on in the documentation, we say when setting up a scheduler for UCX that setting `DASK_UCX__NET_DEVICES` can generally be skipped when using a non-InfiniBand interface is used. Is this universally true, or does it only apply to the scheduler?

pentschev: This can apply to all processes. `UCX_NET_DEVICES` is merely a way to limit the visibility of network devices to UCX, meaning it won't "see" any others. We can still choose the device by specifying `--interface` to scheduler and workers, which is generally preferred if no IB is used. To be even more explicit, we're likely going to drop `DASK_UCX__NET_DEVICES` entirely starting from UCX 1.10, because UCX is now capable of ensuring it uses the IB interface topologically closest to the GPU (which is what `"auto"` does now).

charlesbluca:
Thanks for the review @pentschev! Addressed them as best I could.

I reorganized the use cases into a general "Usage" section and consolidated most of the information on client setup to an individual "Client" section that showcases `dask_cuda.initialize`; feel free to let me know how that looks. My first thought is that it may be nested too deep when it gets to sections on setting up the scheduler + workers using the CLI.

charlesbluca:
Is there anything we can do with the API reference to make `dask_cuda.initialize` show up with its arguments? It seems like it's meant for general use, but right now it seems like the only way to see what arguments it takes is to check the source code on GitHub.

charlesbluca:
Made some changes to the API source so that `dask_cuda.initialize.initialize` should show up properly now; also moved the docstring for the `dask_cuda.initialize` module to the function itself and added some parameters.

dask_cuda/initialize.py
```diff
@@ -44,6 +19,65 @@ def initialize(
...
+
+    .. code-block:: yaml
+
+        distributed:
+          worker:
+            preload:
+              - dask_cuda.initialize
+
+    See https://docs.dask.org/en/latest/configuration.html for more information
+    about Dask configuration.
```
pentschev, line 44: We need to add a note here now to let users know that this is already done for `LocalCUDACluster` (including the client, if it's instantiated on the same process as the cluster) and `dask-cuda-worker`, so users don't need to do that again. The common place where users would want to use this is when they instantiate their standalone client connecting to a cluster that has already been started with a combination of `dask-scheduler`/`dask-cuda-worker`s.

docs/source/ucx.rst
```diff
@@ -1,135 +1,199 @@
...
 
-.. code-block:: bash
+  Adds ``rc`` to ``UCX_TLS``, enabling InfiniBand transfers over UCX.
 
-    DASK_RMM__POOL_SIZE=1GB DASK_UCX__CUDA_COPY=True DASK_UCX__TCP=True DASK_UCX__NVLINK=True DASK_UCX__INFINIBAND=True DASK_UCX__RDMACM=True DASK_UCX__NET_DEVICES=mlx5_0:1 dask-scheduler --protocol ucx --interface ib0
+- ``ucx.rdmacm: true`` -- **recommended for InfiniBand.**
 
-Note above how we use ``DASK_UCX__NET_DEVICES=mlx5_0:1`` (the Mellanox name for ``ib0``) and the same interface with ``--interface ib0``. If the system doesn't have an InfiniBand interface available, you would normally use the main network interface, such as ``eth0``, as seen below:
+  Replaces ``sockcm`` with ``rdmacm`` in ``UCX_TLS`` and ``UCX_SOCKADDR_TLS_PRIORITY``, enabling remote direct memory access (RDMA) for InfiniBand transfers.
+  This is recommended by UCX for use with InfiniBand, and will not work it is disabled.
```
pentschev, line 53:
```suggestion
  This is recommended by UCX for use with InfiniBand, and will not work if InfiniBand is disabled.
```

docs/source/ucx.rst
```diff
@@ -1,135 +1,199 @@
...
+  
+
+- ``rmm.pool-size: <str|int>`` -- **recommended.**
+
+  Allocates an RMM pool of the specified size for the process; size can be provided with an integer number of bytes or in human readable format, e.g. ``"4GB"``.
+  It is recommended to set the pool size to at least the minimum amount of memory used by the process; if possible, one can map all GPU memory to a single pool, to be utilized for the lifetime of the process.
+
+.. note::
+    These options can be used with mainline Dask/Distributed.
+    However, this will disable a variety of features, such as the automatic detection of InfiniBand interfaces.
```
pentschev, line 73:
```suggestion
    However, some features are exclusive to Dask-CUDA, such as the automatic detection of InfiniBand interfaces. 
```

docs/source/ucx.rst
```diff
@@ -1,135 +1,199 @@
...
+Dask-CUDA workers using UCX communication can be started manually with the ``dask-cuda-worker`` CLI tool or automatically with ``LocalCUDACluster``.
+In either case, a ``distributed.Client`` must be made for the worker cluster using the same UCX configuration.
 
 dask-cuda-worker
 ^^^^^^^^^^^^^^^^
 
-All ``DASK_*`` configurations described above have analogous parameters in ``dask-cuda-worker`` which are preferred over the regular configurations used for ``dask-scheduler`` due to some specializations, such as ``--net-devices="auto"`` which will correctly assign the topologically closest IB interface to the GPU of each worker, something that's not possible with ``DASK_UCX__NET_DEVICES``.
+A Dask cluster with UCX support can be started using the ``dask-cuda-worker`` CLI tool with a Dask scheduler which has been configured for UCX.
+
+*Do we want to include any reasons you would do this instead of using LocalCUDACluster?*
```
pentschev, line 88: `LocalCUDACluster` is single-node only, anything else requires `dask-cuda-worker`.

docs/source/ucx.rst
```diff
@@ -1,135 +1,199 @@
...
+    DASK_UCX__CUDA_COPY=True \
+    DASK_UCX__TCP=True \
+    DASK_UCX__NVLINK=True \
+    DASK_UCX__INFINIBAND=True \
+    DASK_UCX__RDMACM=True \
+    DASK_UCX__NET_DEVICES=mlx5_0:1 \
+    DASK_RMM__POOL_SIZE=1GB \
+    dask-scheduler --protocol ucx --interface ib0
 
+Note the specification of ``mlx5_0:1`` as our UCX net device; because the scheduler does not rely upon Dask-CUDA, it cannot automatically detect InfiniBand interfaces, so we must specify one explicitly *(is this true?)*.
```
pentschev, line 108: Yes, it is true.

docs/source/ucx.rst
```diff
@@ -1,135 +1,199 @@
...
-    os.environ["DASK_RMM__POOL_SIZE"] = "1GB"
-    os.environ["DASK_UCX__CUDA_COPY"] = "True"  # os.environ needs using strings, not Python True/False
-    os.environ["DASK_UCX__TCP"] = "True"
-    os.environ["DASK_UCX__NVLINK"] = "True"
-    os.environ["DASK_UCX__INFINIBAND"] = "True"
-    os.environ["DASK_UCX__NET_DEVICES"] = "mlx5_0:1"
+All the relevant Dask configuration options for UCX have analogous parameters in ``dask-cuda-worker``; see :doc:`Worker <worker>` for a complete list of these options.
 
-    from distributed import Client
+To start workers with all supported transports and a 1 gigabyte RMM pool *(do we need to specify an interface here?)*:
```
pentschev, line 128: In most cases, no, UCX will be capable of identifying the optimal network interface to use.

docs/source/ucx.rst
```diff
@@ -1,135 +1,199 @@
...
+        enable_tcp_over_ucx=True,
+        enable_nvlink=True,
+        enable_infiniband=True,
+        ucx_net_devices="auto",
+        rmm_pool_size="1GB",
     )
     client = Client(cluster)
 
-    # Client code goes here
+Note the specification of ``"mlx5_0:1"`` as our net device in ``initialize()``; because the scheduler and client do not rely upon Dask-CUDA, they cannot automatically detect InfiniBand interfaces, so we must specify one explicitly (is this true?).
```
pentschev, line 194: Yes, as in the previous comment in the scheduler section.

pentschev (commented):
Thanks @charlesbluca , this is looking good so far!

I added a few minor suggestions and replied most of your questions (if I missed any, please let me know).

Since this is getting larger than I initially predicted, what do you say we merge this after my comments are addressed and continue with samples in a follow-up PR? I generally think this is better to avoid getting PRs too large and increasingly more difficult to track changes.

docs/source/ucx.rst
charlesbluca: In this case, should I remove the recommendation to explicitly set `ucx.net-devices` to an ethernet device when InfiniBand is disabled? Seems like an inconsistency, unless I'm missing something.

dask_cuda/initialize.py
charlesbluca: Okay, in that case I will emphasize in the docstring that this script is meant for use with the client to mirror the pre-configured CUDA cluster/workers, and that the worker preload script would only need to be used with mainline Dask/Distributed.

charlesbluca:
Addressed your comments - let me know if I should change anything. If not, I'm happy to merge and handle the examples in a separate PR! Thanks for the help 😄

docs/source/ucx.rst
pentschev: The way I read is we recommend users to set to the "optimal" device, so I would still leave it in place. I know in the past we had issues with suboptimal devices being chosen and it seems like UCX 1.10+ is better in that regard, so what I meant to say was that starting from UCX 1.10 we'll then prefer not to set `UCX_NET_DEVICES`, but I don't think in our current state (UCX 1.9), we're ready for that.

dask_cuda/initialize.py
```diff
@@ -25,10 +25,13 @@ def initialize(
     starting up Dask workers which create a variety of threads.
 
     To ensure UCX works correctly, it is important to ensure it is initialized with
-    the correct options. This is important for scheduler, workers, and client. This
-    function will ensure that based on the flags and options passed by the user.
+    the correct options. This is especially important for the scheduler and client,
+    which cannot be configured to use UCX with arguments like ``LocalCUDACluster``
+    and ``dask-cuda-worker``. This function will ensure that they are provided a UCX
+    configuration based on the flags and options passed by the user.
```
pentschev, line 31: But this function can't be used with `dask-scheduler`, so I don't think it really applies to that, only to the client. The scheduler will still have to rely on environment variables, so I would remove it from here.

pentschev (changes requested):
One more small change, then I think we can merge it, but we're still blocked by https://github.com/rapidsai/dask-cuda/pull/546.

dask_cuda/initialize.py
charlesbluca: If the scheduler still relies upon environment variables, does that mean that they must be set explicitly when running Python code?

I had initially removed the `os.environ[...]` calls from the `LocalCUDACluster` example as I thought all non-cluster/worker UCX config was handled through `initialize(...)` - I can add those back in if that's not the case.

pentschev: We have 4 different types:

1. `LocalCUDACluster`: internally sets everything correctly to scheduler and workers, and also does that if `Client` is in the same process being instantiated after the cluster;
2. `dask-scheduler` (CLI): must set UCX options via environment variables (or `distributed.yaml`);
3. `dask-cuda-worker` (CLI): must  set UCX options via command-line arguments;
4. `Client` (standalone process): must set UCX options via environment variables (I think `distributed.yaml` also works here) or may use `dask_cuda.initialize`.

So what I meant was that for case 2 we can't use `dask_cuda.initialize` because it's a standalone process started via CLI. On the other hand, for `Client` we can either choose to use `dask_cuda.initialize` or environment variables, thus I think it only makes sense documenting for use with the client, so I would change the sentence above to:

"This is especially important for the ~scheduler and~ client, ..."

jakirkham:
rerun tests

pentschev (approved):
LGTM, thanks @charlesbluca for the work on improving docs here!

pentschev:
@gpucibot merge

PR #551, opened 2021-03-10
Add standalone examples of UCX usage

charlesbluca:
Following up on #545, this PR adds standalone examples of UCX usage to a new `examples/ucx/` directory.

Right now, these are pretty simple - just showing cluster, worker, and client setups for NVLink, InfiniBand, both, or neither. I imagine there are problem some more complicated/niche set up we could showcase here.

cc @pentschev


pentschev (commented):
I'm not exactly sure what's the best approach here, I've been thinking about this but I don't really like having so many files that are almost identical, any small updates will be a major hassle in the future and the options may not be clear to a viewer without looking into all files. Maybe we should think how to exemplify all options but within a single file. For the bash scripts we can simply add some arguments such as:

```shell
CLUSTER_TYPE=$1

if [ $CLUSTER_TYPE == "IB" ]; then
    DASK_UCX__CUDA_COPY=True
    DASK_UCX__TCP=True
    DASK_UCX__INFINIBAND=True
    DASK_UCX__RDMACM=True
    DASK_UCX__NET_DEVICES=mlx5_0:1
    DASK_RMM__POOL_SIZE=1GB
elif [ $CLUSTER_TYPE == "NVLINK_IB" ]; then
    DASK_UCX__CUDA_COPY=True
    DASK_UCX__TCP=True
    DASK_UCX__NVLINK=True
    DASK_UCX__INFINIBAND=True
    DASK_UCX__RDMACM=True
    DASK_UCX__NET_DEVICES=mlx5_0:1
    DASK_RMM__POOL_SIZE=1GB
fi
```

Maybe we could do something similar with `click` for Python. I'm not sure if those are the best solutions, but just trying to think of some options.

charlesbluca:
Agreed - the current examples are certainly a hassle to keep track of. I think a helper utility like you're describing would work better for simpler cases like what I currently have listed. Happy to consolidate these into something like that.

In addition to that utility, would there be any use in having a collection of intermediate (maybe user-submitted) UCX workflows? Not really sure if intermediate usage differs dramatically from what we currently have showcased in the docs.

pentschev:
> In addition to that utility, would there be any use in having a collection of intermediate (maybe user-submitted) UCX workflows? Not really sure if intermediate usage differs dramatically from what we currently have showcased in the docs.

Are you specifically asking about the Dask code (arrays, dataframes, etc.)? If so, I'd prefer that we don't mix the scripts for setting up a cluster showcasing the differences in UCX usage with workflow code, just to keep it as slim as possible and avoid confusion with their intended purpose. We could still have workflows somewhere in the dask-cuda repository, but I think these would be somewhat redundant with benchmarks.

charlesbluca:
No, mostly just thinking about the possibility of UCX set ups that might require additional configuration - for example, setting up a cluster where different workers have different sized RMM pools (not sure if this has any utility).

If most users are going to end up using the setups we show in the docs with little to no additional configuration, then I'm happy to just consolidate the examples into a Python + Bash utility and leave it at that.

pentschev:
> No, mostly just thinking about the possibility of UCX set ups that might require additional configuration - for example, setting up a cluster where different workers have different sized RMM pools (not sure if this has any utility).

The only use for this that I can think of is with a cluster where the GPUs have different amounts of memory, but that is unrelated to UCX, so I think it's not important to cover that, at least not these samples.

> If most users are going to end up using the setups we show in the docs with little to no additional configuration, then I'm happy to just consolidate the examples into a Python + Bash utility and leave it at that.

I think that would be enough, we mostly want samples that we can refer users to for UCX because that's currently a bit cumbersome.

charlesbluca:
First pass at a UCX utility for CUDA workers; some thoughts:

- I would like a more robust flag system to set other variables (RMM pool size, interface, etc.) instead of the current single argument that only decides the transports; this could result in significant overlap with the actual CLI though.
- Is there a way to get the host address from the CLI scheduler instance to pass to the workers? As it is, the host that the scheduler is running on would need to be added to the script manually, and we can't specify a host address explicitly when we also specify an interface.

Now working on a Python utility.

pentschev:
> * I would like a more robust flag system to set other variables (RMM pool size, interface, etc.) instead of the current single argument that only decides the transports; this could result in significant overlap with the actual CLI though.

Again I don't think we need to do that for RMM, the point here is demo-ing UCX options. Interface might make sense, but I'd like to keep it UCX-oriented.

> * Is there a way to get the host address from the CLI scheduler instance to pass to the workers? As it is, the host that the scheduler is running on would need to be added to the script manually, and we can't specify a host address explicitly when we also specify an interface.

There's two things you could do:

1. Provide an argument to the script: `SCHEDULER_ADDRESS=$2`;
2. Write a [scheduler file](https://github.com/dask/distributed/blob/7c6709fd8d2ee6356ba43ceff64cd04e9803b2a1/distributed/cli/dask_scheduler.py#L96-L103) to a path accessible by all nodes, later [using the same file on workers](https://github.com/rapidsai/dask-cuda/blob/09196cb5c92effca6da660231e58a5bf4ac72c76/dask_cuda/cli/dask_cuda_worker.py#L140-L146).

charlesbluca:
> Provide an argument to the script: SCHEDULER_ADDRESS=$2;

Would this argument then be passed to `dask-scheduler`'s [`--host`](https://github.com/dask/distributed/blob/7c6709fd8d2ee6356ba43ceff64cd04e9803b2a1/distributed/cli/dask_scheduler.py#L30) or a different flag?

I think the scheduler file sounds like a good option that I'll try out.

> Again I don't think we need to do that for RMM, the point here is demo-ing UCX options.

Sure, that makes sense. If I add additional arguments, I'll keep it limited to those you would need to change in order to get a functioning cluster.

pentschev:
> Would this argument then be passed to `dask-scheduler`'s [`--host`](https://github.com/dask/distributed/blob/7c6709fd8d2ee6356ba43ceff64cd04e9803b2a1/distributed/cli/dask_scheduler.py#L30) or a different flag?

No, it would be the first argument to `dask-cuda-worker`, something such as `ucx://SCHEDULER_IP:8786`. That of course implies that the user knows where the scheduler is listening.

examples/ucx/client_initialize.py
```diff
@@ -0,0 +1,45 @@
...
+def main(
+    address, enable_nvlink, enable_infiniband,
+):
+
+    enable_rdmacm = False
+    ucx_net_devices = None
+
+    if enable_infiniband:
+        enable_rdmacm = True
+        ucx_net_devices = "mlx5_0:1"
```
charlesbluca, line 31: Is it reliable to set the UCX net device for IB explicitly like this, or should we leave this open for user input?

examples/ucx/dask_cuda_worker.sh
```diff
@@ -0,0 +1,55 @@
...
+fi
+if [[ $transport == *"nvlink"* ]]; then
+    DASK_UCX__NVLINK=True
+
+    worker_flags+=" --enable-nvlink"
+fi
+if [[ $transport == *"ib"* ]]; then
+    DASK_UCX__INFINIBAND=True
+    DASK_UCX__RDMACM=True
+    DASK_UCX__NET_DEVICES=mlx5_0:1
```
charlesbluca, line 46: Same question as before with setting net devices explicitly.

examples/ucx/local_cuda_cluster.py
```diff
@@ -0,0 +1,59 @@
...
+    ucx_net_devices = None
+
+    if enable_infiniband:
+        enable_rdmacm = True
+        ucx_net_devices = "auto"
+
+    if (enable_infiniband or enable_nvlink) and not interface:
+        raise ValueError("Interface must be specified if using NVLink or Infiniband")
+
+    cluster = LocalCUDACluster(  # noqa F842
```
charlesbluca, line 47: How can we keep this cluster running until the user manually terminates the script? I would like it so a user can use this script to start up a cluster that can later be connected to by the example client initialization script.

charlesbluca (commented):
Added basic Python CLIs for both cluster and client initialization; also changed around the dask-cuda-worker script so it behaves closer to a CLI.

I have a couple questions regarding these additions:

examples/ucx/local_cuda_cluster.py
pentschev: `LocalCUDACluster` is generally intended to be used in the same process as `Client`, and in that case the `Client` doesn't need to set additional variables as `LocalCUDACluster` already took care of that. With that in mind, I would simply suggest we add:

```python
from distributed import Client

...

client = Client(cluster)
```

Doing the above would make this sample complete IMO.

pentschev (commented):
This PR looks really great overall, @charlesbluca ! I responded to your questions, let me know what you think.

charlesbluca:
Thanks @pentschev! Addressed those comments with these last commits.

examples/ucx/local_cuda_cluster.py
```diff
@@ -35,25 +38,41 @@ def main(
...
+        )
+
+    # set up environment
+    initialize(
+        enable_tcp_over_ucx=True,
+        enable_nvlink=enable_nvlink,
+        enable_infiniband=enable_infiniband,
+        enable_rdmacm=enable_rdmacm,
+        net_devices=client_devices,
+    )
```
pentschev, line 60: Sorry if my previous comment was unclear, but this is what I meant, `initialize` isn't necessary here because `LocalCUDACluster` does that internally: https://github.com/rapidsai/dask-cuda/blob/250260e54e0b3c4065bf7bba77cb13f7b032f030/dask_cuda/local_cuda_cluster.py#L257-L264 . Because the `Client` will then be using device 0, leaving `ucx_net_devices="auto"` will also correctly select the client's device.

In short, we only need https://github.com/rapidsai/dask-cuda/pull/551/commits/79b7d5d03574bd68c821cd45edf51e06af9abe52#diff-d9b0b02a0c749f46db506de5db50c2944b96f42895a621ac051539cc5f0da484R3 and https://github.com/rapidsai/dask-cuda/pull/551/commits/79b7d5d03574bd68c821cd45edf51e06af9abe52#diff-d9b0b02a0c749f46db506de5db50c2944b96f42895a621ac051539cc5f0da484R73-R75 from this commit, the rest can be reverted to how it was previously.

pentschev (changes requested):

examples/ucx/local_cuda_cluster.py
charlesbluca: Ah okay I understand - in that case I should probably alter the [Client](https://github.com/rapidsai/dask-cuda/blob/branch-0.19/docs/source/ucx.rst#client) section of the UCX docs, which also makes this mistake.

pentschev: Oops, that was my fault for not noticing it when you updated the docs, sorry for the extra work. 🙁

charlesbluca: No worries! Was a quick fix 😃

BTW, do GH review comments generally not give out notifications? I've found myself missing several comments (yours included) because I had to refresh a page to get them to appear.

pentschev: I usually get notifications, but I rely on email notifications. I'm not sure about the actual GH notifications, I stopped using them because they force you to actually load the issue/PR to see what's going on, whereas on email I can just navigate through most of those unimportant ones much more quickly.

pentschev (approved):

charlesbluca:
Thanks for the help on this @pentschev!

Am I able to merge this with gpuCI or can only codeowners?

pentschev:
> Am I able to merge this with gpuCI or do only codeowners have the permissions?

I don't know, try it. 😄

charlesbluca:
@gpucibot merge

charlesbluca:
Looks like I can't 😅

pentschev:
Thanks @charlesbluca for the work here!

PR #556, opened 2021-03-25
Fix UCX examples for InfiniBand

charlesbluca:
Some small changes to the UCX examples to fix some IB-related issues I discussed with @pentschev:

- Disables RDMACM for all examples; right now this isn't working, but when it is we can uncomment out the original lines
- Use `--scheduler-file` to link `dask-scheduler` to `dask-cuda-workers`, as `localhost` cannot be used when IB is enabled
- Add a small workload for the client and shut it down in the cluster and client setup examples

### DIFF
--- a/examples/ucx/local_cuda_cluster.py
+++ b/examples/ucx/local_cuda_cluster.py
@@ -1,5 +1,7 @@
 import click
+import cupy
 
+from dask import array as da
 from dask.distributed import Client
 from dask.utils import parse_bytes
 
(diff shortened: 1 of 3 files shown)


examples/ucx/local_cuda_cluster.py
```diff
@@ -60,7 +62,15 @@ def main(
...
     # initialize client
-    client = Client(cluster)  # noqa F841
+    client = Client(cluster)
+
+    # client code here
+    rs = da.random.RandomState(RandomState=cupy.random.RandomState)
+    x = rs.random((10000, 10000), chunks=1000)
+    x.sum().compute()
+
+    # shutdown client
```
pentschev, line 72:
```suggestion
    # shutdown cluster
```

pentschev (changes requested):

charlesbluca:
Thanks for the review @pentschev!

pentschev (approved):
LGTM, thanks @charlesbluca !

pentschev:
@gpucibot merge

Issue #560, opened 2021-04-05
[DOCS] missing CLI params


lmeyerov:
Looks like a bunch of the CLI params @ https://github.com/rapidsai/dask-cuda/blob/branch-0.19/dask_cuda/cli/dask_cuda_worker.py (ex: `--nthreads`) are not in https://github.com/rapidsai/dask-cuda/blob/branch-0.19/docs/source/worker.rst  . I'm fine looking at code, but maybe a useful heads up :)

charlesbluca:
I'm actually reworking the docs now, so I've been looking for a way to auto-generate the CLI parameter docs - thanks for pointing this out!

charlesbluca:
Just to follow up on this, I think [click-contrib/sphinx-click](https://github.com/click-contrib/sphinx-click) should be perfect for this.

charlesbluca:
This *should* be resolved with the next release, as the CLI param docs are now autogenerated:

- https://github.com/rapidsai/dask-cuda/blob/branch-0.20/docs/source/api.rst#worker

Thanks for bringing this up!

Issue #554, opened 2021-03-22
Add minimal example of dask-cuda command line cluster setup


beckernick:
In the docs, we have a clear [example](https://dask-cuda.readthedocs.io/en/latest/worker.html) of setting up a LocalCUDACluster. We also list several additional arguments to the "Command Line Tool", which corresponds to the [Dask CLI](https://docs.dask.org/en/latest/setup/cli.html). However, it's not obvious from the dask-cuda documentation what needs to change relative to standard usage of the Dask CLI. We have an example of using `dask-cuda-worker` in the [UCX integration setup](https://dask-cuda.readthedocs.io/en/latest/ucx.html#usage), but it's quite possible a user wouldn't look at the UCX integration if they aren't already planning to use UCX.

It would be valuable if we included a minimal example of setting up a multi-node cluster (or equivalent single-node setup) with `dask-cuda-worker` from the command line, like in the Dask docs. This could live in either dask-cuda or dask. This documentation would be a valuable asset to share with users interested in testing multi-node or from the command line.

pentschev:
Just recently @charlesbluca added some code to try and address specifically that, which now lives in https://github.com/rapidsai/dask-cuda/tree/branch-0.19/examples/ucx . Does that cover your question, if not, what else could we do to improve it?

charlesbluca:
Maybe it would be better to duplicate some of the general usage info in the UCX section in the Worker section, so that it's more visible? We can always slim down the UCX section later on if we find that parts of it are redundant.

I'm also happy to expand upon the standalone utilities - if we made the use of UCX an option, then they could function as a minimal example of a standard multi/single-node set up.

beckernick:
@charlesbluca I think that approach would be a great idea

charlesbluca:
Currently drafting some changes to address the docs aspect of this in #561 - in general, I am leaning towards putting basic usage examples in the first few pages, and moving more advanced configuration to standalone examples in the docs.

jakirkham:
Is this resolved with PR ( https://github.com/rapidsai/dask-cuda/pull/561 ) or was there more work needed here?

charlesbluca:
I think that's sufficient - we could always potentially add an example script like [those made for UCX](https://github.com/rapidsai/dask-cuda/tree/branch-0.20/examples/ucx).

@beckernick did you get a chance to check out the changes to the docs?

pentschev:
[...]

PR #70, opened 2019-06-06
Add a hard-coded DGX configuration (#46)

mrocklin:
Trying this again.  The last PR was aimed at branch-0.7

PR #561, opened 2021-04-06
Update API reference and examples in docs

charlesbluca:
Addresses  #554; I tried to tidy up the installation/specializations/quickstart pages so that some of the more technical stuff could go into some published examples, and probably a new page discussing configuration of LocalCUDACluster / `dask-cuda-worker`.

Also added `sphinx-click` to the dependencies so we can more easily document any changes to `dask-cuda-worker` (#560) - I would also like to get the docstrings for the CLI, LocalCUDACluster, and `initiailize()` up to date and more concise if possible.


dask_cuda/cli/dask_cuda_worker.py
```diff
@@ -19,154 +19,160 @@
...
 )
-@click.option("--nthreads", type=int, default=1, help="Number of threads per process.")
 @click.option(
     "--name",
     type=str,
     default=None,
-    help="A unique name for this worker like 'worker-1'. "
-    "If used with --nprocs then the process number "
-    "will be appended like name-0, name-1, name-2, ...",
+    help="""A unique name for this worker like ``worker-1``; if used with ``--nprocs``,
```
pentschev, line 43: This is just nipticking, but in here, worker-1 is actually a string, so using back quote doesn't make sense to me as it's normally used elsewhere for keywords, and I would just use single or double quotes instead.

For keywords such as --nprocs, maybe we should just use a single back quote, as there's no text parsing and instead only CLI prints? Or is Sphinx parsing dask-cuda-worker CLI arguments now?

Same comments apply to other cases below, if we decide to change.

dask_cuda/cli/dask_cuda_worker.py
```diff
@@ -19,154 +19,160 @@
...
     default=True,
-    help="Reconnect to scheduler if disconnected",
+    show_default=True,
+    required=False,
+    help="Launch the dashboard",
+)
+@click.option(
+    "--dashboard-address",
+    type=str,
+    default=":0",
```
pentschev, line 115: Does ":0" mean a random port will be assigned? If so, is that what we want or do we want to keep Distributed's default (":8787")? Also, are you sure the colon is required/supported here?

docs/source/install.rst
```diff
@@ -0,0 +1,35 @@
+Installation
+============
+
+To use Dask-CUDA on your system, you will need the following:
+
+- At least one CUDA-compatible GPU
+- NVIDIA CUDA Toolkit
+- *Anything else?*
```
pentschev, line 8: Technically, Dask-CUDA won't interface directly with the CUDA Toolkit, so it's not exactly a dependency. However, it will be a dependency for other libraries that will be used for CUDA computation, like CuPy or Numba. IOW, the compute libraries will need to satisfy the CUDA Toolkit dependency, and not Dask-CUDA.

docs/source/install.rst
```diff
@@ -0,0 +1,35 @@
...
+    conda install -c rapidsai -c nvidia -c conda-forge dask-cuda cudatoolkit=11.0
+
+Pip
+---
+
+To do the same with ``pip``::
+
+    python -m pip install dask-cuda
+
+*Is there any way to specify toolkit version here? Or is pip install not recommended*
```
pentschev, line 26: As I mentioned above, the CUDA Toolkit isn't a dependency to Dask-CUDA, so whatever library users want will need to satisfy that instead.

dask_cuda/cli/dask_cuda_worker.py
```diff
@@ -175,38 +181,36 @@
...
-    "InfiniBand only and will still cause unpredictable errors if not _ALL_ "
-    "interfaces are connected and properly configured.",
+    help="""If specified, workers will use this interface for UCX communication; can be
+    a string (like ``eth0``), or ``auto`` to pick the optimal interface based on
+    the system's topology; typically only used with ``--enable-infiniband``
+
+    .. warning::
+        ``auto`` requires UCX-Py to be installed and compiled with hwloc support;
+        unexpected errors can occur when using ``auto`` if any interfaces are
+        disconnected or improperly configured""",
```
pentschev, line 206: This is removing a lot of important information, such as the IB interface names that can be used, as well as "auto" being limited to IB only. Can we add that back?

pentschev (commented):
Thanks @charlesbluca for working on this, I added some comments.

dask_cuda/cli/dask_cuda_worker.py
charlesbluca: Not sure, this was already set when I edited the file - looking at the blame, this was done in #70 for a hard-coded example, so maybe it's good to change this to match with Distributed's default now.

docs/source/install.rst
charlesbluca: Good to know - is there anything in particular that is required to use Dask-CUDA? Right now I'd imagine:

- CUDA-compliant GPU
- Some installation of CUDA

On that topic, I think to address the fact that the install instructions used to direct users to install cuDF + the toolkit, maybe I'll add a section at the bottom of this page directing to the general RAPIDS quickstart guide.

dask_cuda/cli/dask_cuda_worker.py
charlesbluca: Yeah I chose the backticks here as it's being formatted by sphinx-click now, the same way a docstring would be:

image

I agree that adding quotes to strings would help distinguish them from options, although I'm not sure the best way to address the backticks in the CLI output.

dask_cuda/cli/dask_cuda_worker.py
```diff
@@ -19,154 +19,160 @@
...
+    .. note::
+        Logging will only be enabled if ``--rmm-pool-size`` or ``--rmm-managed-memory``
+        are specified.""",
+)
+@click.option("--pid-file", type=str, default="", help="File to write the process PID")
+@click.option(
+    "--resources",
+    type=str,
+    default="",
+    help="""Resources for task constraints like ``GPU=2 MEM=10e9``; resources are
```
charlesbluca, line 100: Didn't really understand this input when I passed over it; is this meant to be one string i.e. `"GPU=2 MEM=10e9"`?

dask_cuda/cli/dask_cuda_worker.py
pentschev: Maybe then it makes sense to write ``"worker-1"`` ?

Double backticks for CLI output are ok then, given we now have Sphinx autogenerating formatted output.

dask_cuda/cli/dask_cuda_worker.py
pentschev: Hmm, ok, I thought that was being added just now, I initially missed the removal lines, so let's keep it as is for now.

docs/source/install.rst
pentschev: It's difficult to say what is particularly required, we intentionally try to make it as unrestricted as possible. It's definitely true that CUDA GPUs and a CUDA installation are required, but this is actually necessary by the downstream libraries that will make use of the GPU itself, as Dask-CUDA will not interact directly with the GPU. However, Numba is a dependency and [Numba CUDA support depends on CUDA Toolkit](https://numba.pydata.org/numba-doc/latest/cuda/overview.html#software). Perhaps we should then indeed say CUDA Toolkit is a requirement, but that isn't available in PIP so the user needs to get it elsewhere. We're fine with providing PIP packages, but we shouldn't encourage it, so maybe we should intentionally limit documentation on it.

dask_cuda/cli/dask_cuda_worker.py
pentschev: AFAIU, yes. This is to provide annotations so that the application may later constrain certain parts of its execution to specific workers, see [Worker Resources](https://distributed.dask.org/en/latest/resources.html).

dask_cuda/local_cuda_cluster.py
```diff
@@ -39,96 +39,95 @@ def __init__(self, *args, **kwargs):
...
-    threads_per_worker: int
-        Number of threads to be used for each CUDA worker process.
-    protocol: str
-        Protocol to use for communication, e.g., ``"tcp"`` or ``"ucx"``.
-    enable_tcp_over_ucx: bool
-        Set environment variables to enable TCP over UCX, even if InfiniBand
-        and NVLink are not supported or disabled.
-    enable_infiniband: bool
-        Set environment variables to enable UCX InfiniBand support, requires
+    CUDA_VISIBLE_DEVICES : str or list of int, optional
```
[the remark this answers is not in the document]
charlesbluca: I was working off of the [style guide from numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#sections), but I realize that it would probably be good for a user to know if the default value for a kwarg is `None`. This was mostly done to create a consistent location for the default values to be displayed, as before some kwargs didn't have their defaults displayed in the docstring.

I'm happy to change this so that `optional` kwargs now have a `default None`, but in that case I think for sake of completeness it would be good to specify that `None` is a valid argument for these kwargs, and say specifically what happens. Using `jit_unspill` as an example:

```
    jit_unspill : bool, optional
        Enable just-in-time unspilling. This is experimental and doesn't support memory
        spilling to disk. See ``proxy_object.ProxyObject`` and
        ``proxify_host_file.ProxifyHostFile`` for more info.
``` 

should become:

```
    jit_unspill : bool or None, default None
        Enable just-in-time unspilling. Can be a boolean, or None to fall back on the
        value of ``dask.jit-unspill`` in the local Dask configuration (disabling if this
        is not set). This is experimental and doesn't support memory spilling to disk.
        See ``proxy_object.ProxyObject`` and
        ``proxify_host_file.ProxifyHostFile`` for more info.
```

Does that sound good?

dask_cuda/local_cuda_cluster.py
```diff
@@ -39,96 +39,95 @@ def __init__(self, *args, **kwargs):
...
+        String (like ``"0,1,2,3"``) or  list (like ``[0, 1, 2, 3]``) to restrict
+        activity to different GPUs.
+    n_workers : int, optional
+        Number of workers. Will override the value of ``CUDA_VISIBLE_DEVICES`` if
+        specified.
+    threads_per_worker : int, default 1
+        Number of threads to be used for each Dask worker process.
+    processes : bool, default True
+        TODO
```
charlesbluca, line 64: Is this parameter required to extend a `LocalCluster`? Looking through the code, all it really does is raise a `ValueError` if it's set false, and it isn't used when calling the super constructor:

https://github.com/rapidsai/dask-cuda/blob/7a6be90e1a72d1caec7f77ce1221d8d6c31134c4/dask_cuda/local_cuda_cluster.py#L286-L290

dask_cuda/local_cuda_cluster.py
```diff
@@ -39,96 +39,95 @@ def __init__(self, *args, **kwargs):
...
+        ``"auto"``, or 0 for no memory management.
+    device_memory_limit : int, float or str, default 0.8
+        Size of the CUDA device LRU cache, which is used to determine when the worker
+        starts spilling to host memory. Can be an integer (bytes), float (fraction of
+        total device memory), string (like ``"5GB"`` or ``"5000M"``), ``"auto"``, or 0
+        to disable spilling to host (i.e. allow full device memory usage).
+    data :
+        TODO
+    local_directory : str, optional
+        TODO
```
charlesbluca, line 77: What do these parameters control?

charlesbluca (commented):
A few more questions:

dask_cuda/local_cuda_cluster.py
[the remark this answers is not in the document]
pentschev: IMO, I would just not document defaults and let users instead look at the function signature, although this is perhaps not the most popular of opinions. My concern mostly comes to: either we should document _everything_ or _nothing_, and over time it's generally the case that documentation tends to drift apart from implementation quite easily. In some cases we may we elaborate in the description what is the default for more complex kwargs (e.g., `device_memory_limit` and `ucx_net_devices`), and I would be fine with that.

Summarizing, I don't mind adding "optional" or "default ...", but then it should be consistently done to _all_ `LocalCUDACluster` kwargs in this PR, and hopefully extend that to the rest of dask-cuda in the future.

dask_cuda/local_cuda_cluster.py
pentschev: I don't think we need it, it's probably something that there was an idea in mind at the early days of dask-cuda, but it's not necessary at the moment so I'd just remove it.

dask_cuda/local_cuda_cluster.py
pentschev: `data` is a mapping object where Dask workers store chunks, depending on its internal mechanism it will control device-to-host and host-to-disk spilling, etc. Although it's possible to extend it, it's quite complex to do so and generally users are better off leaving it alone, so I would not document it, and perhaps we could remove it and let it be passed to `**kwargs`, which we could then pop its value so it's not directly exposed in  `LocalCUDACluster`'s API.

`local_directory` is the same as in [Distributed Worker](https://distributed.dask.org/en/latest/worker.html): "Path on local machine to store temporary files".

dask_cuda/local_cuda_cluster.py
charlesbluca: Opened up a separate PR addressing that - #562.

dask_cuda/cli/dask_cuda_worker.py
charlesbluca: Would this be good to open a separate issue about? Not sure if this results in weird default behavior for the dashboard.

dask_cuda/local_cuda_cluster.py
charlesbluca: Thanks! I ended up expanding #562 to encompass this as well.

dask_cuda/cli/dask_cuda_worker.py
```diff
@@ -19,194 +19,202 @@
...
+    help="""External interface used to connect to the scheduler. Usually an ethernet
+    interface is used for connection, and not an InfiniBand interface (if one is
+    available).""",
 )
 @click.option(
-    "--dashboard-prefix", type=str, default=None, help="Prefix for the Dashboard"
+    "--death-timeout",
+    type=str,
+    default=None,
+    help="Seconds to wait for a scheduler before closing.",
```
charlesbluca, line 140: It looks like this argument isn't used in `CUDAWorker` - I think we would want to pass this on to the worker's nanny, since that's where I believe the timeout is implemented.

dask_cuda/local_cuda_cluster.py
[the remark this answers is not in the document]
pentschev: These are definitely the most important/higher-level ones and IMO suffices for now. If there's more, I'd prefer that in a follow-up PR as this is getting quite extensive already. 🙂

dask_cuda/cli/dask_cuda_worker.py
pentschev: If it's not even being passed and nobody complained for so long it probably means it's not important, let's drop it instead.

charlesbluca: It probably would've been easier to make a list of all the kwargs to drop before opening a PR 😅 I'll address this in a separate PR since #562 is nearly ready to merge.

pentschev: Sounds good to me.

docs/source/install.rst
charlesbluca: Thanks for the explanation! In that case, CUDA toolkit should probably go in the list of main requirements, and then I think I'll put CUDA and the compliant GPUs as requirements for the optional RAPIDS libraries.

One question - when Conda installing Dask-CUDA, is the CUDA toolkit version specified (like `cudatoolkit=11.2`) to install the toolkit alongside Dask-CUDA, or to make sure that its dependencies are using versions that support the system installation of the toolkit? If it's the latter, do we know how Pip handles the CUDA toolkit version? If not, then that makes a pretty good case to recommend Conda install and removing the Pip instructions altogether.

docs/source/install.rst
```diff
@@ -1,20 +1,15 @@
...
 ============
 
-To use Dask-CUDA on your system, you will need the following:
-
-- At least one CUDA-compatible GPU
-- NVIDIA CUDA Toolkit
-- *Anything else?*
-
-Once these requirements are satisfied, Dask-CUDA can be install using ``conda``, ``pip``, or from source.
+To use Dask-CUDA on your system, you will need the `NVIDIA CUDA Toolkit <https://developer.nvidia.com/cuda-toolkit>`_.
```
pentschev, line 4: This is not exactly correct either. If you have a driver version compatible with that of `cudatoolkit` package installed via conda, you don't need to install the toolkit from the NVIDIA website. IOW, this should only be needed for pip.

charlesbluca:
I think this PR is at a good closing point - the API reference is updated, the installation + quickstart is fleshed out a bit more, and there are clear minimal examples of using LocalCUDACluster and `dask-cuda-worker`. Some follow up stuff that would be nice:

- Some examples of usage with cuDF/CuPy
- Migrating the UCX Dask configuration options to Dask's docs when dask/distributed#4683 gets merged
- Generalizing the UCX utilities to work with TCP as well (or just make another standalone utility for TCP)

dask_cuda/local_cuda_cluster.py
```diff
@@ -155,11 +162,12 @@ class LocalCUDACluster(LocalCluster):
 
     def __init__(
         self,
+        CUDA_VISIBLE_DEVICES=None,
         n_workers=None,
```
pentschev, line 166: Failures in `test_spill` are because `n_workers` is being passed by position. Could you fix them?

But also in the future we should prevent reordering kwargs for the same reason.

charlesbluca: Sure! Good point, I recognize that the reordering probably causes more harm than good here.

docs/source/install.rst
pentschev: Could we address this as well? I think after that we're good to merge.

charlesbluca: Sure! Sorry, missed this one.

charlesbluca: Actually, for clarity, is the driver version in question that of CUDA?

dask_cuda/cli/dask_cuda_worker.py
```diff
@@ -19,188 +19,222 @@
...
+    help="""Module that should be loaded by each worker process like ``"foo.bar"`` or
+    ``"/path/to/foo.py"``.""",
 )
-@click.argument(
-    "preload_argv", nargs=-1, type=click.UNPROCESSED, callback=validate_preload_argv
+@click.option(
+    "--dashboard-prefix",
+    type=str,
+    default=None,
+    help="""Prefix for the dashboard. Can be a string (like ...) or ``None`` for no
```
charlesbluca, line 160: Do we know what dashboard prefix input should look like?

charlesbluca (commented):
One last question I missed:

docs/source/install.rst
pentschev: Driver and CUDA toolkit versions are two different things, the CUDA toolkit version must always be compatible with that of the driver version. Take a look at [Table 1](https://docs.nvidia.com/deploy/cuda-compatibility/index.html#binary-compatibility), it lists what's the minimum driver version you need for a specific CUDA toolkit's version.

dask_cuda/cli/dask_cuda_worker.py
pentschev: I'm not sure, maybe @jacobtomlinson can help with this one.

docs/source/install.rst
charlesbluca: What I'm trying to understand is how the driver is installed (it seems like it's included with a website download of CUDA toolkit, but I'm not finding much info on downloading the driver itself), and then maybe how to list the currently installed version of the driver. So far, I'm imagining the install section should look something like:

**Conda**
- Install NVIDIA driver or identify the currently installed version
- Conda install Dask-CUDA and a compatible version of CUDA toolkit (maybe linking the chart)

**Pip**
- Install CUDA toolkit from the website (this should come with the NVIDIA driver)
- Pip install Dask-CUDA

pentschev: The driver has to be installed by a user with admin privileges (i.e., root), therefore drivers can't be distributed via conda, so that means the driver is a system requirement. The driver can be obtained from a package of the user's Linux distro, or from NVIDIA's website, and in the latter case it's distributed both as part of the [CUDA toolkit](https://developer.nvidia.com/cuda-toolkit) as a [standalone package](https://www.nvidia.com/Download/index.aspx).

I think the idea above for Conda and Pip looks fine to me.

charlesbluca: Thanks! I actually found [this installation guide](https://docs.nvidia.com/datacenter/tesla/tesla-installation-notes/index.html), which reiterates what you said - I think I'll include as the primary resource for handling driver support (in the Conda case).

dask_cuda/cli/dask_cuda_worker.py
charlesbluca: Actually, I see that this is also the default for Distributed, so I'm going to assume this is a safe default.

pentschev (approved):
LGTM, thanks @charlesbluca !

pentschev:
@gpucibot merge

PR #562, opened 2021-04-07
Remove kwargs from `LocalCUDACluster`

charlesbluca:
Based on some discussions in #561, this PR removes some kwargs from `LocalCUDACluster` that may not be needed:

- `processes` doesn't have any impact on its construction (other than throwing a `ValueError` if it's set false); this should be safe to remove altogether
- `data` is pretty complex and it isn't commonly specified, so it can be removed from the named arguments and accessed through the `kwargs` object if it is specified

### DIFF
--- a/dask_cuda/local_cuda_cluster.py
+++ b/dask_cuda/local_cuda_cluster.py
@@ -228,7 +228,7 @@ class LocalCUDACluster(LocalCluster):
         else:
             self.jit_unspill = jit_unspill
 
-        data = kwargs.get("data")
+        data = kwargs.pop("data", None)
         if data is None:
             if self.jit_unspill:
                 data = (
--- a/dask_cuda/local_cuda_cluster.py
+++ b/dask_cuda/local_cuda_cluster.py
@@ -160,7 +160,6 @@ class LocalCUDACluster(LocalCluster):
         memory_limit="auto",
         device_memory_limit=0.8,
         CUDA_VISIBLE_DEVICES=None,
-        data=None,
         local_directory=None,
         protocol=None,
         enable_tcp_over_ucx=False,
@@ -229,6 +228,7 @@ class LocalCUDACluster(LocalCluster):
         else:
             self.jit_unspill = jit_unspill
 
+        data = kwargs.get("data")
         if data is None:
             if self.jit_unspill:
                 data = (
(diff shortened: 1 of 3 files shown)


jakirkham:
Should `data` be dropped from the `super` call below since it is now in `kwargs`?

charlesbluca:
I don't think so, as `data` is still assigned a value later on:

https://github.com/rapidsai/dask-cuda/blob/7a6be90e1a72d1caec7f77ce1221d8d6c31134c4/dask_cuda/local_cuda_cluster.py#L238-L255

jakirkham:
In that case should we pop it from `kwargs`? Otherwise it is being passed twice to `super`. Once explicitly and once implicitly

charlesbluca:
Yes! Thanks, I missed that.

pentschev (approved):
LGTM, thanks @charlesbluca !

jakirkham:
Should add we will need to do the same thing for `processes`

pentschev:
> Should add we will need to do the same thing for `processes`

Actually it just seems like we have a bunch of tests that explicitly pass `processes=True`, we are better off removing those instead. Could you do that @charlesbluca ?

charlesbluca:
Done! Should we check if `processes` is provided as a kwarg and make sure to pop it so it doesn't end up in the super call?

pentschev:
> Done! Should we check if `processes` is provided as a kwarg and make sure to pop it so it doesn't end up in the super call?

I don't mind leaving it as is as it isn't really supported, but if others prefer, we can pop from kwargs and reinstate the `ValueError` exception.

jakirkham:
Continuing to raise an error makes sense as this should make it clearer that it is unsupported (as opposed to erroring when it is passed again to `super`). Maybe we can just check `if not kwargs.pop("processes", True)`?

pentschev:
Failures are happening across all PRs, I've opened https://github.com/rapidsai/dask-cuda/issues/564 to track this.

pentschev:
@charlesbluca I retargeted this to branch-0.20 since we're in code freeze and this isn't really a very high priority change for 0.19, better keep it safe.

pentschev:
> That's fine! Should I have permissions to retarget PRs? #561 #563 should also be retargetted since those are low priority.

Yes, you should be able to do that by clicking "Edit" next to the PR title, then changing to branch-0.20 from the drop-down list where branch-0.19 is currently selected.

jakirkham:
Went ahead and just retargeted those as well. Hope that is ok 🙂

charlesbluca:
Thanks @jakirkham!

pentschev:
@gpucibot merge

PR #563, opened 2021-04-07
Remove `--death-timeout` from option from `dask-cuda-worker`

charlesbluca:
This removes `death_timeout` from the `dask-cuda-worker` CLI and `CUDAWorker`, as it previously was not being used for anything. I think we could get it working by passing its value on to the nanny processes being spawned by `CUDAWorker`, but @pentschev suggested it might be better to remove it since its lack of functionality hasn't really been a problem up until now.


pentschev:
Failures are happening across all PRs, I've opened https://github.com/rapidsai/dask-cuda/issues/564 to track this.

pentschev:
rerun tests

pentschev:
@gpucibot merge

pentschev (approved):

Issue #443, opened 2020-11-17
Passing `rmm_pool_size` larger than memory will not create a pool


pentschev:
During some conversations offline, I noticed passing `rmm_pool_size` that's larger than the total GPU memory will succeed, although no pool gets created or warnings get thrown. A simple reproducer is below:

```python
from dask_cuda import LocalCUDACluster
cluster = LocalCUDACluster(rmm_pool_size="100GB", silence_logs=False)
```

By checking `nvidia-smi`, it's clear that the cluster gets created but without a memory pool.

pentschev:
This should be fixed by https://github.com/dask/distributed/issues/4297 .

quasiben:
This is still an issue

pentschev:
With https://github.com/dask/distributed/issues/4297 merged in, this is now fixed:

```
$ dask-cuda-worker tcp://localhost:8786 --rmm-pool-size 100GB
...
distributed.utils - ERROR - std::bad_alloc: out_of_memory: RMM failure at:~/miniconda3/envs/test_dev/include/rmm/mr/device/pool_memory_resource.hpp:183: Maximum pool size exceeded
Traceback (most recent call last):
  File "~/miniconda3/envs/test_dev/lib/python3.8/site-packages/distributed/utils.py", line 681, in log_errors
    yield
  File "~/miniconda3/envs/test_dev/lib/python3.8/site-packages/distributed/worker.py", line 3283, in plugin_add
    result = plugin.setup(worker=self)
  File "~/miniconda3/envs/test_dev/lib/python3.8/site-packages/dask_cuda/utils.py", line 85, in setup
    rmm.reinitialize(
  File "~/miniconda3/envs/test_dev/lib/python3.8/site-packages/rmm/rmm.py", line 84, in reinitialize
    rmm.mr._initialize(
  File "rmm/_lib/memory_resource.pyx", line 672, in rmm._lib.memory_resource._initialize
  File "rmm/_lib/memory_resource.pyx", line 732, in rmm._lib.memory_resource._initialize
  File "rmm/_lib/memory_resource.pyx", line 270, in rmm._lib.memory_resource.PoolMemoryResource.__cinit__
MemoryError: std::bad_alloc: out_of_memory: RMM failure at:~/miniconda3/envs/test_dev/include/rmm/mr/device/pool_memory_resource.hpp:183: Maximum pool size exceeded
```

PR #566, opened 2021-04-09
Add support for `CudaAsyncMemoryResource`

### LINKED ISSUE #565: Extend CLI and LocalCluster to support CUDA Async Memory Resource
With the new functionality in CUDA 11.2, RMM now supports a `CudaAsyncMemoryResource` in addition to the existing resources. It would be nice to extend the dask-cuda CLI to allow starting a cluster with an RMM pool using either of the memory resources.

EDIT: Turns out we don't want to use this with an explicit RMM pool, but we might still want to expose this as it would allow a caching pool allocator that could play more nicely with other GPU libraries that don't use RMM

charlesbluca:
Closes #565

Adds the `--rmm-pool-async`/`rmm_pool_async` option to the CLI and cluster to enable the use of `rmm.mr.CudaAsyncMemoryResource` in the RMM initialization.

### DIFF
--- a/dask_cuda/cli/dask_cuda_worker.py
+++ b/dask_cuda/cli/dask_cuda_worker.py
@@ -121,8 +121,8 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
 
     .. note::
         The asynchronous allocator requires CUDA Toolkit 11.2 or newer. It is also
-        incompatible with RMM pools and managed memory, and will be preferred over them
-        if both are enabled.""",
+        incompatible with RMM pools and managed memory, trying to enable both will
+        result in an exception.""",
 )
 @click.option(
     "--rmm-log-directory",
--- a/dask_cuda/cli/dask_cuda_worker.py
+++ b/dask_cuda/cli/dask_cuda_worker.py
@@ -129,8 +129,8 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     default=None,
     help="Directory to write per-worker RMM log files to; the client "
     "and scheduler are not logged here."
-    "NOTE: Logging will only be enabled if --rmm-pool-size or "
-    "--rmm-managed-memory are specified.",
+    "NOTE: Logging will only be enabled if --rmm-pool-size, "
+    "--rmm-managed-memory, or --rmm-async are specified.",
 )
 @click.option(
     "--reconnect/--no-reconnect",
--- a/dask_cuda/cli/dask_cuda_worker.py
+++ b/dask_cuda/cli/dask_cuda_worker.py
@@ -122,7 +122,7 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     .. note::
         The asynchronous allocator requires CUDA Toolkit 11.2 or newer. It is also
         incompatible with RMM pools and managed memory, trying to enable both will
-        result in an exception.""",
+        result in failure.""",
 )
 @click.option(
     "--rmm-log-directory",
--- a/dask_cuda/cli/dask_cuda_worker.py
+++ b/dask_cuda/cli/dask_cuda_worker.py
@@ -103,6 +103,12 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     "and not cluster-wide!",
 )
 @click.option(
+    "--rmm-pool-async/--no-rmm-pool-async",
+    default=False,
+    show_default=True,
+    help="Use an ``rmm.mr.CudaAsyncMemoryResource`` when initializing an RMM pool.",
+)
+@click.option(
     "--rmm-managed-memory/--no-rmm-managed-memory",
     default=False,
     help="If enabled, initialize each worker with RMM and set it to "
@@ -210,6 +216,7 @@ def main(
     memory_limit,
     device_memory_limit,
     rmm_pool_size,
+    rmm_pool_async,
     rmm_managed_memory,
     rmm_log_directory,
     pid_file,
@@ -254,6 +261,7 @@ def main(
         memory_limit,
         device_memory_limit,
         rmm_pool_size,
+        rmm_pool_async,
         rmm_managed_memory,
         rmm_log_directory,
         pid_file,
--- a/dask_cuda/cli/dask_cuda_worker.py
+++ b/dask_cuda/cli/dask_cuda_worker.py
@@ -103,12 +103,6 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     "and not cluster-wide!",
 )
 @click.option(
-    "--rmm-pool-async/--no-rmm-pool-async",
-    default=False,
-    show_default=True,
-    help="Use an ``rmm.mr.CudaAsyncMemoryResource`` when initializing an RMM pool.",
-)
-@click.option(
     "--rmm-managed-memory/--no-rmm-managed-memory",
     default=False,
     help="If enabled, initialize each worker with RMM and set it to "
@@ -117,6 +111,18 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     "(non-managed) memory type."
     "WARNING: managed memory is currently incompatible with NVLink, "
     "trying to enable both will result in an exception.",
+)
+@click.option(
+    "--rmm-async/--no-rmm-async",
+    default=False,
+    show_default=True,
+    help="""Initialize each worker withh RMM and set it to use RMM's asynchronous
+    allocator. See ``rmm.mr.CudaAsyncMemoryResource`` for more info.
+
+    .. note::
+        The asynchronous allocator requires CUDA Toolkit 11.2 or newer. It is also
+        incompatible with RMM pools and managed memory, and will be preferred over them
+        if both are enabled.""",
 )
 @click.option(
     "--rmm-log-directory",
@@ -216,8 +222,8 @@ def main(
     memory_limit,
     device_memory_limit,
     rmm_pool_size,
-    rmm_pool_async,
     rmm_managed_memory,
+    rmm_async,
     rmm_log_directory,
     pid_file,
     resources,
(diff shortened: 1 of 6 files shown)


dask_cuda/cli/dask_cuda_worker.py
```diff
@@ -102,6 +102,12 @@
     "NOTE: This size is a per worker (i.e., per GPU) configuration, "
     "and not cluster-wide!",
 )
+@click.option(
+    "--rmm-pool-async/--no-rmm-pool-async",
```
charlesbluca, line 106: I chose to go with a flag for this option - if we think there might be other resources down the line we also want to support, a string input may be a better option.

dask_cuda/utils.py
```diff
@@ -42,6 +43,11 @@ def setup(self, worker=None):
         if self.nbytes is not None or self.managed_memory is True:
             import rmm
 
+            if self.pool_async is True:
+                rmm.mr.set_current_device_resource(
+                    rmm.mr.PoolMemoryResource(rmm.mr.CudaAsyncMemoryResource())
+                )
```
charlesbluca, line 49: Does this break down if RMM managed memory is enabled? If so, I can check to make sure it's disabled.

charlesbluca (commented):
Some small questions:

charlesbluca:
It looks like CUDA 11.2 has different behavior, in that the memory resource remains an `rmm.mr.PoolMemoryResource` regardless of if `rmm_pool_async` is enabled. Is there a way we can check for CUDA version in the tests to handle this?

pentschev:
`cudaMallocAsync` was only introduced in CUDA 11.2, https://github.com/rapidsai/rmm/blob/80bfeb2816845da5921551ae5c158e92427bec86/python/rmm/tests/test_rmm.py#L511-L539 will give you some ideas on how to test that async is both active and how you can skip tests for older CUDA versions.

dask_cuda/cli/dask_cuda_worker.py
pentschev: Yeah, this is why I suggested the string idea. For now we can keep it like this and extend it in the future.

dask_cuda/utils.py
pentschev: These are two different malloc calls, `cudaMallocAsync` and `cudaMallocManaged`, so I believe we can't have an async managed malloc.

@shwina could you confirm if my comment above is correct? Are we only able to use one or the other?

charlesbluca:
Thanks @pentschev! In that case, I'm a little interested in what's happening for the 11.0 tests, as I would expect a `RuntimeError` upon trying to set up async there.

pentschev:
> I'm assuming that the `CudaMemoryResource` that the workers return for `rmm.mr.get_current_device_resource_type` in the 11.0 cases is because `RMMSetup` is silently failing to allocate a pool, similar to what's happening in #443. With that knowledge, should we try to add some form of check that the user has 11.2 if they enable async, or wait on [dask/distributed#4297](https://github.com/dask/distributed/issues/4297) to begin handling worker plugin exceptions?

It's possible, but some discussion seems to point that we don't want an async pool, but rather just switching to the async allocator, as commented in https://github.com/rapidsai/dask-cuda/issues/565#issuecomment-816979483 . Could you update the PR here to reflect that? I think that will also make the assertion easier, you can probably then just do the same as https://github.com/rapidsai/rmm/blob/80bfeb2816845da5921551ae5c158e92427bec86/python/rmm/tests/test_rmm.py#L519-L521 .

charlesbluca:
Sure! Does this mean that async can be enabled for RMM pools and managed memory?

pentschev:
No, that means we're changing the allocator to the async allocator (which is neither the default nor the managed memory allocators), and without a pool. In summary, enabling the async allocator (something like `--rmm-async-malloc`) is mutually exclusive with both `--rmm-pool-size` and `--rmm-managed-memory` and would raise an error if that's attempted.

dask_cuda/utils.py
```diff
@@ -32,14 +32,26 @@ def setup(self, worker=None):
...
+    def __init__(self, nbytes, managed_memory, async_alloc, log_directory):
         self.nbytes = nbytes
         self.managed_memory = managed_memory
+        self.async_alloc = async_alloc
         self.logging = log_directory is not None
         self.log_directory = log_directory
 
     def setup(self, worker=None):
-        if self.nbytes is not None or self.managed_memory is True:
+        if self.async_alloc is True:
```
charlesbluca, line 43: I added a separate conditional checking for the async allocator to make sure it is mutually exclusive with pools or managed memory - is there a cleaner way of doing this or should I not bother?

dask_cuda/utils.py
```diff
@@ -32,14 +32,26 @@ def setup(self, worker=None):
...
+            import rmm
+
+            rmm.mr.set_current_device_resource(rmm.mr.CudaAsyncMemoryResource())
+
+            rmm.reinitialize(
+                logging=self.logging,
+                log_file_name=get_rmm_log_file_name(
+                    worker, self.logging, self.log_directory
+                ),
```
charlesbluca, line 52: Should I still be calling `rmm.reinitialize()` in the async case? If so, should `pool_allocator` be enabled?

charlesbluca (commented):
Updated to reflect this; some questions:

dask_cuda/utils.py
pentschev: Quoting @shwina : 

> You don’t need the reinitialize call here. reinitialize is a higher-level API to `set_current_device_resource`, but it doesn’t expose the async memory resource. `set_current_device_resource()` should already have “reinitialized” RMM with the pool async MR.

dask_cuda/utils.py
pentschev: I think we should either raise an error if the user specifies both (otherwise, given the assumption they're mutually exclusive, it's impossible for a user to know which one is actually being used), or at least emit a warning. Warnings can be annoying and easy to ignore, thus I'd prefer raising an error.

dask_cuda/utils.py
charlesbluca: Okay, in that case I'll remove that - I guess my larger concern surrounding this is how we would allow for RMM logging when using the async allocator.

shwina: You could use `rmm.enable_logging()` after calling `set_current_device_resource()`.

charlesbluca: Great! Thanks for the help 🙂

dask_cuda/utils.py
charlesbluca: Sounds good - I opted for a `ValueError` if both are enabled, so I ended up leaving the separate conditional since now it shouldn't be ambiguous which RMM setup is used.

dask_cuda/utils.py
shwina: Correct, there's no way to use both.

dask_cuda/cli/dask_cuda_worker.py
```diff
@@ -121,8 +121,8 @@
 
     .. note::
         The asynchronous allocator requires CUDA Toolkit 11.2 or newer. It is also
-        incompatible with RMM pools and managed memory, and will be preferred over them
-        if both are enabled.""",
+        incompatible with RMM pools and managed memory, trying to enable both will
+        result in an exception.""",
```
pentschev, line 125: Since this is the CLI, is this really going to raise an exception or fail with an error instead? I'm not really sure what happens when an exception is raise in the `CUDAWorker` class when running the CLI tool.

dask_cuda/local_cuda_cluster.py
```diff
@@ -143,10 +151,12 @@ class LocalCUDACluster(LocalCluster):
...
     ValueError
-        If ``ucx_net_devices`` is an empty string, or if it is ``"auto"`` and UCX-Py is
-        not installed, or if it is ``"auto"`` and ``enable_infiniband=False``, or UCX-Py
-        wasn't compiled with hwloc support, or both RMM managed memory and
-        NVLink are enabled.
+       If ``ucx_net_devices=""``, if NVLink and RMM managed memory are
+       both enabled, if RMM pools / managed memory and asynchronous allocator are both
+       enabled, or if ``ucx_net_devices="auto"`` and:
+
+        - UCX-Py is not installed or wasn't compiled with hwloc support or
```
pentschev, line 158:
```suggestion
        - UCX-Py is not installed or wasn't compiled with hwloc support; or
```

dask_cuda/local_cuda_cluster.py
```diff
@@ -143,10 +151,12 @@ class LocalCUDACluster(LocalCluster):
...
+        - ``enable_infiniband=False``
```
pentschev, line 159: Seems like you accidentally removed a white space from the beginning of each line in this block, which probably will result in a complaint when parsing the docstrings.

dask_cuda/utils.py
charlesbluca: Good to know! As resolved below, we now raise a `ValueError` if both are enabled.

dask_cuda/cli/dask_cuda_worker.py
charlesbluca: It looks like this fails with an error - I mostly just copied the phrasing used to describe the behavior of using NVLink + managed memory:

https://github.com/rapidsai/dask-cuda/blob/1526017749ae0a3137c40605bb992d426d9fee3b/dask_cuda/cli/dask_cuda_worker.py#L112-L114

I can change this here and in #561 if it would make more sense to say the CLI will fail outright in these cases.

dask_cuda/local_cuda_cluster.py
charlesbluca: Good catch!

pentschev (approved):
LGTM, thanks @charlesbluca !

PR #568, opened 2021-04-12
Proxify ignore some types

madsbk:
This PR makes it possible to _not_ proxify some types. 
For now, this disable proxification of `cupy.ndarray` until we find a good solution to handle them.

### DIFF
--- a/dask_cuda/proxify_device_objects.py
+++ b/dask_cuda/proxify_device_objects.py
@@ -1,3 +1,4 @@
+import pydoc
 from typing import Any, List, MutableMapping
 
 from dask.utils import Dispatch
@@ -5,6 +6,24 @@ from dask.utils import Dispatch
 from .proxy_object import ProxyObject, asproxy
 
 dispatch = Dispatch(name="proxify_device_objects")
+ignore_types = ()
+
+
+def _register_ignore_types():
+    """Lazy register types that shouldn't be proxified"""
+    global ignore_types
+
+    # TODO: make configable
+    for (module, types) in [("cupy", ("cupy.ndarray",))]:
+
+        def f():
+            global ignore_types
+            ignore_types = tuple(pydoc.locate(t) for t in types)
+
+        dispatch.register_lazy(module, f)
+
+
+_register_ignore_types()
 
 
 def proxify_device_objects(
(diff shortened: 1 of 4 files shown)


dask_cuda/proxify_device_objects.py
```diff
@@ -1,10 +1,50 @@
...
+            toplevel = path.split(".", maxsplit=1)[0].strip()
+            toplevels[toplevel].add(path.strip())
+
+    for toplevel, ignores in toplevels.items():
+
+        def f(paths):
+            global ignore_types
+            ignore_types = ignore_types + tuple(pydoc.locate(p) for p in paths)
+
+        dispatch.register_lazy(toplevel, partial(f, ignores))
```
pentschev, line 47: This is more of a nitpick question, but why do you need to separate objects into different toplevels? I can't seem to understand why wouldn't just assigning `path` to a `set` suffice. In that case, this code here would be much simpler.

pentschev (approved):
> For now, this disable proxification of `cupy.ndarray` until we find a good solution to handle them.

Out of curiosity, why can't `cupy.ndarray` be proxified just like other objects? I must have missed the explanation for that elsewhere.

Overall looks good to me, just a minor nitpick. Thanks @madsbk !

dask_cuda/proxify_device_objects.py
madsbk: It is because `register_lazy` only looks at the top level module. That is, when it detect that `cupy` is imported it calls `f()`. Also notice, each call to `register_lazy` _overwrites_ the previous registered function

madsbk:
> Out of curiosity, why can't `cupy.ndarray` be proxified just like other objects? I must have missed the explanation for that elsewhere.

It is because some functions such as `cupy.dot()` are written in Cython where they _cast_ the input to a `cupy.ndarray`. I am working on a solution to overcome this issue.

dask_cuda/proxify_device_objects.py
pentschev: Makes sense, thanks for clarifying.

pentschev:
Thanks @madsbk !

pentschev:
@gpucibot merge

Issue #569, opened 2021-04-13
Add Conda environment files for development


charlesbluca:
As I'm remaking some of my dev environments, I noticed that dask-cuda doesn't seem to have an environment file for development - there is the [requirements.txt](https://github.com/rapidsai/dask-cuda/blob/branch-0.20/requirements.txt) and some dependencies listed implicitly in the [Conda recipe](https://github.com/rapidsai/dask-cuda/blob/branch-0.20/conda/recipes/dask-cuda/meta.yaml), but those don't seem to include some of the other requirements for testing or build docs.

Is there a dev environment file I'm missing? If not, would it be feasible/useful to add one? Not sure if we would need to add multiple for different versions of CUDA toolkit like [cuDF's case](https://github.com/rapidsai/cudf/tree/branch-0.20/conda/environments).

pentschev:
Generally this isn't a requirement because we're not really building anything from source (unlike cuDF), so the instructions from the [RAPIDS Getting Started](https://rapids.ai/start.html) page are also sufficient for a dev environment.

Are you suggesting a specialized use case for which the instructions from the link above doesn't cover? If so, which one would that be?

charlesbluca:
I suppose cuDF isn't really the best example since the environment there is necessity rather than a convenience. I agree that the quickstart page does proper instructions on how to install the nightly build of dask-cuda, thought it may not be immediately obvious to non-RAPIDS contributors since the package isn't listed there.

The main use case I'm thinking of is if a user wants to do documentation work / full testing on dask-cuda in a minimal environment. In this case, if they followed the quickstart or dask-cuda's documentation to install it, they would still need to do some trial and error to sort out the requirements for running the full test suite, linting the codebase, or building the documentation.

pentschev:
I would very much prefer that we document this rather than having conda environment files. The natural follow-up to that would be "How about developers that use Pip?", and this ends up with us having to maintain several files that are very seldom used.

IMO, the instructions we would need are simply:
* How to run pytests;
* How to build docs;
* Linting using pre-commit.

I don't necessarily oppose to having conda environment files for that, but I just think it's great if we can keep maintenance efforts low.

jakirkham:
Maybe this is useful for [reusing `requirements.txt` with Conda]( https://stackoverflow.com/a/57845418 )?

jakirkham:
[...]

PR #573, opened 2021-04-16
Make package versions consistent between Conda recipe and `requirements.txt`

charlesbluca:
Keeps the versions of Numba and Dask consistent across the Conda recipe and requirements file - see [#569 (comment)](https://github.com/rapidsai/dask-cuda/issues/569#issuecomment-819028783)

### DIFF
--- a/conda/recipes/dask-cuda/meta.yaml
+++ b/conda/recipes/dask-cuda/meta.yaml
@@ -23,7 +23,8 @@ requirements:
     - python
     - setuptools
   run:
-  {% for package in data.get('packages') %}
+    - python
+  {% for package in data.get('install_requires', []) %}
     - {{ package }}
   {% endfor %}
 
--- a/conda/recipes/dask-cuda/meta.yaml
+++ b/conda/recipes/dask-cuda/meta.yaml
@@ -2,12 +2,12 @@
 
 # Usage:
 #   conda build -c conda-forge -c defaults .
-{% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev').lstrip('v') + environ.get('VERSION_SUFFIX', '') %}
+{% set data=load_setup_py_data() %}
 {% set git_revision_count=environ.get('GIT_DESCRIBE_NUMBER', 0) %}
 {% set py_version=environ.get('CONDA_PY', 36) %}
 package:
   name: dask-cuda
-  version: {{ version }}
+  version: {{ data.get('version') }}
 
 source:
   path: ../../..
@@ -23,12 +23,9 @@ requirements:
     - python
     - setuptools
   run:
-    - python
-    - dask >=2.4.0
-    - distributed >=2.18.0
-    - pynvml >=8.0.3
-    - numpy >=1.16.0
-    - numba >=0.50.0,!=0.51.0
+  {% for package in data.get('packages') %}
+    - {{ package }}
+  {% endfor %}
 
 test:
   imports:
--- a/conda/recipes/dask-cuda/meta.yaml
+++ b/conda/recipes/dask-cuda/meta.yaml
@@ -2,12 +2,12 @@
 
 # Usage:
 #   conda build -c conda-forge -c defaults .
-{% set data=load_setup_py_data() %}
+{% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev').lstrip('v') + environ.get('VERSION_SUFFIX', '') %}
 {% set git_revision_count=environ.get('GIT_DESCRIBE_NUMBER', 0) %}
 {% set py_version=environ.get('CONDA_PY', 36) %}
 package:
   name: dask-cuda
-  version: {{ data.get('version') }}
+  version: {{ version }}
 
 source:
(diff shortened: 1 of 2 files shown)


conda/recipes/dask-cuda/meta.yaml
```diff
@@ -23,12 +23,9 @@ requirements:
     - python
     - setuptools
   run:
-    - python
-    - dask >=2.4.0
-    - distributed >=2.18.0
-    - pynvml >=8.0.3
-    - numpy >=1.16.0
-    - numba >=0.50.0,!=0.51.0
+  {% for package in data.get('packages') %}
```
charlesbluca, line 26: Should we also be using [`resolved_packages('host')`](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html?highlight=load_setup_py_data#conda-build-specific-jinja2-functions) here to include the host packages?

conda/recipes/dask-cuda/meta.yaml
```diff
@@ -2,12 +2,12 @@
 
 # Usage:
 #   conda build -c conda-forge -c defaults .
-{% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev').lstrip('v') + environ.get('VERSION_SUFFIX', '') %}
```
jakirkham, line 5: I think we still need this as OPS uses it to create nightly packages

conda/recipes/dask-cuda/meta.yaml
jakirkham: We probably want to use something like [`data.get('install_requires', [])`]( https://github.com/conda/conda-build/blob/3b99b2222a067e113a2282926871cd1e5406ee2b/conda_build/templates/setuptools.yaml#L28-L30 )

`resolved_packages` pins down to the exact package used at build time, which isn't likely what we want

jakirkham:
Thanks Charles! 😄 

Added a couple suggestions above

conda/recipes/dask-cuda/meta.yaml
charlesbluca: Do you know if the `version` set in `setup.py`

https://github.com/rapidsai/dask-cuda/blob/3c6e0a9d18615376c05074622b0cec2fc1189aeb/setup.py#L12-L16

Matches with the `version` set in this line? Then we could just do something like

```
{% set version = data.get('version', '0.0.0.dev') %}
```

conda/recipes/dask-cuda/meta.yaml
charlesbluca: That makes sense, but in that case should we update `requirements.txt`? It looks like it has different version restrictions from the current recipe - for example:

https://github.com/rapidsai/dask-cuda/blob/3c6e0a9d18615376c05074622b0cec2fc1189aeb/requirements.txt#L5

https://github.com/rapidsai/dask-cuda/blob/3c6e0a9d18615376c05074622b0cec2fc1189aeb/conda/recipes/dask-cuda/meta.yaml#L31

jakirkham: Yeah I think so. Thanks for catching that 🙂

conda/recipes/dask-cuda/meta.yaml
raydouglass: I'm not sure that they match. In `setup.py` it will be prefixed with `v` if I'm not mistaken, but conda certainly disallows that.

I personally don't see a need to change the conda recipe since it matches the rest of RAPIDS.

charlesbluca: The main idea here was to reduce duplication between `setup.py` and the Conda recipe, which is more a convenience than anything else.

If you think this is an unnecessary change, I'm happy to keep the original Conda recipe, although I do think that the [`requirements.txt`](https://github.com/rapidsai/dask-cuda/blob/branch-0.20/requirements.txt) in this repo should be updated to align with it.

jakirkham:
Looks like we are seeing a CI failure, but this appears unrelated to this change. Filed as issue ( https://github.com/rapidsai/dask-cuda/issues/575 )

conda/recipes/dask-cuda/meta.yaml
```diff
@@ -24,11 +24,9 @@ requirements:
     - setuptools
   run:
     - python
-    - dask >=2.4.0
```
jakirkham, line 27: It's worth noting this is older than the version in `requirements.txt`. Would suggest we go with the newer of the two (otherwise no change need). Just mentioning this for awareness

conda/recipes/dask-cuda/meta.yaml
jakirkham: Maybe let's stick to the `requirements.txt` change for now

pentschev:
rerun tests

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
Needs a review from OPS

raydouglass (approved):

Issue #579, opened 2021-04-19
Moving documentation to docs.rapids.ai


charlesbluca:
After some internal discussion, we have decided to move the documentation of Dask-CUDA from Read the Docs to [RAPIDS Docs](https://docs.rapids.ai/). This would make it relatively easy to set up a docs CI workflow script like [this](https://github.com/rapidsai/cudf/blob/branch-0.20/ci/docs/build.sh), and is motivated by the fact that there are a lot of barriers keeping this repo from migrating to Dask org.

What's a clean way to do this without causing confusion? Should we keep both sites up simultaneously for a period of time with some banner notifying the migration? Or have the RTD docs redirect straight to the RAPIDS docs?

cc @quasiben @pentschev

charlesbluca:
Opened #580 to make a gpuCI docs build script.

pentschev:
Redirecting seems like a simpler way to deal with it.

charlesbluca:
I agree, just wasn't sure if RTD had the ability to easily manage a redirect but looks like it can.

charlesbluca:
Dask-CUDA's docs just got published to RAPIDS docs:

- https://docs.rapids.ai/api/dask-cuda/nightly/

Now we would just want to figure out redirecting the RTD version of the docs.

jakirkham:
In conda-forge we did this like [so]( https://github.com/conda-forge/conda-forge.github.io/tree/df804a45b7a99823f7cf6e330ff7caac07ea24e4/rtd )

charlesbluca:
Great - thanks for that example! Do you know if we would need to keep that `rtd/` directory in our repo if we disable webhooks on RTD's end after the last build?

jakirkham:
Peter was mentioning redirecting, which agree probably makes sense (there may be other links to RTD out there).

So the `rtd` directory is important for supplying a documentation build that creates a webpage on RTD to do the redirecting. This is set in [the RTD config file]( https://github.com/conda-forge/conda-forge.github.io/blob/df804a45b7a99823f7cf6e330ff7caac07ea24e4/.readthedocs.yaml#L3-L4 ). We wouldn't want that in our usual docs hence the separate directory.

PR #580, opened 2021-04-19
Add docs CI build script

charlesbluca:
Adds a docs build script for gpuCI to execute on merges; addresses #579.

Depends on rapidsai/integration#250

cc @quasiben

### DIFF
--- a/ci/docs/build.sh
+++ b/ci/docs/build.sh
@@ -15,7 +15,6 @@ export PATH=/conda/bin:/usr/local/cuda/bin:$PATH
 export HOME=$WORKSPACE
 export PROJECT_WORKSPACE=/rapids/dask-cuda
 export NIGHTLY_VERSION=$(echo $BRANCH_VERSION | awk -F. '{print $2}')
-export PROJECTS=(dask-cuda)
 
 gpuci_logger "Check environment..."
 env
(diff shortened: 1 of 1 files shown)


ci/docs/build.sh
```diff
@@ -0,0 +1,53 @@
...
+
+gpuci_logger "Check environment..."
+env
+
+gpuci_logger "Check GPU usage..."
+nvidia-smi
+
+gpuci_logger "Activate conda env..."
+. /opt/conda/etc/profile.d/conda.sh
+conda activate rapids
```
charlesbluca, line 28: Are there any packages we would need to add to this environment to get docs builds working?

https://github.com/rapidsai/dask-cuda/blob/d39fff9cd6f766dc03cb9f6858b19caf7d3cc438/conda/environments/builddocs_py37.yml#L5-L9

ci/docs/build.sh
```diff
@@ -0,0 +1,53 @@
...
+    echo ">>>> ERROR: Could not detect PROJECT_WORKSPACE in environment"
+    echo ">>>> WARNING: This script contains git commands meant for automated building, do not run locally"
+    exit 1
+fi
+
+export DOCS_WORKSPACE=$WORKSPACE/docs
+export PATH=/conda/bin:/usr/local/cuda/bin:$PATH
+export HOME=$WORKSPACE
+export PROJECT_WORKSPACE=/rapids/dask-cuda
+export NIGHTLY_VERSION=$(echo $BRANCH_VERSION | awk -F. '{print $2}')
```
charlesbluca, line 17: `HOME` and `NIGHTLY_VERSION` are exported but unused (copied this from cuDF's build script); are these required by gpuCI or can I remove these?

charlesbluca (commented):
Some questions:

ci/docs/build.sh
dillon-cullinan: HOME is a variable needed by Jenkins, and NIGHTLY_VERSION is used in the build outside this script.

ci/docs/build.sh
dillon-cullinan: The docs are built in the `devel` images and automatically download some basic stuff such as rapids-doc-env. I don't think `sphinx-click` is included in our `devel` environment so this may need to be added manually.

dillon-cullinan (approved):
This looks fine, but read my comments above to make sure you're not missing anything, as I'm not sure if they're required or not.

ci/docs/build.sh
charlesbluca: Would this need to be added from the script, or could it be added to the `devel` environment permanently?

charlesbluca:
Waiting on rapidsai/integration#250 before we merge this.

charlesbluca:
With rapidsai/integration#250 in, this should be good to go - might need to wait on the `devel` image to be rebuilt with the `rapids-doc-env` but after that things should work smoothly.

charlesbluca:
@gpucibot merge

PR #593, opened 2021-04-27
Explicit comms shuffle: sort workers when assigning output partitions

madsbk:


pentschev (commented):
Looks fine to me, do we need a test here too @madsbk ?

madsbk:
> Looks fine to me, do we need a test here too @madsbk ?

No, I don't think so this is only a minor performance optimization.

pentschev (approved):
LGTM, thanks @madsbk .

pentschev:
@gpucibot merge

PR #603, opened 2021-05-11
JIT-Unspill Compatibility Mode

madsbk:
This PR introduces _Compatibility Mode_:
```python
class ProxifyHostFile(MutableMapping):
    """Host file that proxify stored data
    ...

    compatibility_mode: bool
        Enables compatibility-mode, which means that items are un-proxified before
        retrieval. This makes it possible to get some of the JIT-unspill benefits
        without having to be ProxyObject compatible. In order to still allow specific
        ProxyObjects, set the `mark_as_explicit_proxies=True` when proxifying with
        `proxify_device_objects()`. If None, the "jit-unspill-compatibility-mode"
        config value are used, which defaults to False.
    """
```

### DIFF
--- a/dask_cuda/proxify_device_objects.py
+++ b/dask_cuda/proxify_device_objects.py
@@ -1,3 +1,4 @@
+import functools
 import pydoc
 from collections import defaultdict
 from functools import partial
@@ -52,6 +53,7 @@ def proxify_device_objects(
     proxied_id_to_proxy: MutableMapping[int, ProxyObject] = None,
     found_proxies: List[ProxyObject] = None,
     excl_proxies: bool = False,
+    mark_as_explicit_proxies: bool = False,
 ):
     """ Wrap device objects in ProxyObject
 
@@ -73,6 +75,9 @@ def proxify_device_objects(
         If None, use an empty list.
     excl_proxies: bool
         Don't add found objects that are already ProxyObject to found_proxies.
+    mark_as_explicit_proxies: bool
+        Mark found proxies as "explicit", which means that the user allows them
+        as input arguments to dask tasks even in compatibility-mode.
 
     Returns
     -------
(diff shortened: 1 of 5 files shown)


dask_cuda/proxify_host_file.py
```diff
@@ -142,13 +143,26 @@ class ProxifyHostFile(MutableMapping):
     ----------
     device_memory_limit: int
         Number of bytes of CUDA device memory used before spilling to host.
+    compatibility_mode: bool
```
pentschev, line 146:
```suggestion
    compatibility_mode: bool or None
```

dask_cuda/proxify_host_file.py
```diff
@@ -142,13 +143,26 @@ class ProxifyHostFile(MutableMapping):
...
+        Enables compatibility-mode, which means that items are un-proxified before
+        retrieval. This makes it possible to get some of the JIT-unspill benefits
+        without having to be ProxyObject compatible. In order to still allow specific
+        ProxyObjects, set the `mark_as_explicit_proxies=True` when proxifying with
+        `proxify_device_objects()`. If None, the "jit-unspill-compatibility-mode"
+        config value are used, which defaults to False.
     """
 
-    def __init__(self, device_memory_limit: int):
+    def __init__(self, device_memory_limit: int, compatibility_mode: bool = None):
```
pentschev, line 155: Is this allowed, to define the type as `bool` and assign `None` to it?

dask_cuda/tests/test_proxify_host_file.py
```diff
@@ -216,3 +219,34 @@ def test_proxify_device_objects_of_cupy_array():
...
+    with dask.config.set(jit_unspill_compatibility_mode=True):
+        with dask_cuda.LocalCUDACluster(n_workers=1, jit_unspill=True) as cluster:
+            with Client(cluster):
+                ddf = dask.dataframe.from_pandas(
+                    cudf.DataFrame({"key": np.arange(10)}), npartitions=npartitions
+                )
+                res = ddf.shuffle(on="key", shuffle="tasks").persist()
+                assert "ProxyObject" not in str(type(res.compute()))
+                res = res.map_partitions(is_proxy_object).compute()
+                assert not any(res.to_list())
```
pentschev, line 252: Should we make this two separate tests with `pytest.mark.parametrize`?

pentschev (commented):
@madsbk overall looks good to me. I've added a few questions/comments.

dask_cuda/proxify_device_objects.py
```diff
@@ -66,19 +68,98 @@ def proxify_device_objects(
...
+    skip_explicit_proxies: bool
+        When True, skipping proxy objects marked as explicit proxies.
+
+    Returns
+    -------
+    ret: Any
+        A copy of `obj` where all CUDA device objects are unproxify
+    """
+    typ = type(obj)
+    if typ is dict:
```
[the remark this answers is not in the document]
madsbk: Only because that is Dask traverses the values because of performance. But I think you are right, in this case `isinstance` is better

dask_cuda/proxify_host_file.py
madsbk: I am not totally sure, but this is allowed by mypy by default: https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-optional

dask_cuda/tests/test_proxify_host_file.py
madsbk: Fixed

dask_cuda/proxify_device_objects.py
```diff
@@ -66,19 +68,98 @@ def proxify_device_objects(
...
+        return {
+            k: unproxify_device_objects(v, skip_explicit_proxies)
+            for k, v in obj.items()
+        }
+    if typ in (list, tuple, set, frozenset):
```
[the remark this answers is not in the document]
madsbk: Fixed

pentschev (approved):
LGTM, thanks for the work here @madsbk !

pentschev:
@gpucibot merge

madsbk:
Thanks @pentschev !

PR #604, opened 2021-05-12
Document DASK_JIT_UNSPILL

### LINKED ISSUE #597: Document DASK_JIT_UNSPILL
We should write docs for the `DASK_JIT_UNSPILL` functionality first introduced in:  https://github.com/rapidsai/dask-cuda/pull/451.

madsbk:
Fixes #597

Also moving *Spilling from device* to the *ADDITIONAL FEATURES* section


quasiben:
This is great @madsbk, thanks for working on it!  It's probably worth mentioning the benefits (and potential problems) of using it.  Something like, "when experimenting with JIT-UNSPILL we generally see better performance and lower memory usage". @beckernick do you have some general idea on percentages here ?  Is it 20% better?

beckernick:
> @beckernick do you have some general idea on percentages here ? Is it 20% better?

This will be workload dependent. However, for workloads that require significant spilling (such as large joins) we have often seen greater than 50% improvement (i.e., something taking 300 seconds might take only 110 seconds). For workloads that do not, we would not expect to see much delta.

madsbk:
@quasiben updated the docs to include some more information

quasiben (approved):
Thanks @madsbk !

quasiben:
@gpucibot merge

PR #620, opened 2021-05-21
Disable reuse endpoints with UCX >= 1.11

pentschev:
The UCX-Py endpoint reuse is not anymore necessary, so we also disable that for UCX 1.11+. The primary reason it was introduced was to circumvent an issue with CUDA IPC that was resolved by https://github.com/openucx/ucx/pull/6360. Using the endpoint reuse class has also proven to be very slow, taking a long time to initialize for clusters with just a few dozen workers and pretty much unusable for a cluster in the order of 100 workers.

### DIFF
--- a/dask_cuda/utils.py
+++ b/dask_cuda/utils.py
@@ -27,8 +27,10 @@ try:
     import ucp
 
     _ucx_110 = ucp.get_ucx_version() >= (1, 10, 0)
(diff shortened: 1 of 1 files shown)


quasiben (approved):

quasiben:
@gpucibot merge

pentschev:
Thanks @quasiben !

PR #645, opened 2021-06-07
Disable install Dask master in CI

pentschev:


jakirkham (approved):

ajschmidt8 (approved):

ajschmidt8:
This will require a manual merge once CI is complete.

ajschmidt8:
@pentschev, can you elaborate on the `numba` change in https://github.com/rapidsai/dask-cuda/pull/645/commits/9d7697182f2dd4a74da8f8d2d051d44b7a1eacb1?

pentschev:
I thought the build failure was about the Numba version, because the [rest of RAPIDS now pins Numba>=0.53.1](https://github.com/rapidsai/integration/blob/37f52deb0acc7826759cef61f2e5109192fee3db/conda/recipes/versions.yaml#L109-L110):

```
20:40:36 /tmp/tmp26qug3g1/dask-cuda-21.06.00a-py37_56.tar.bz2: C1114 Found invalid dependency "numba >=0.50.0,!=0.51.0" in info/index.json
```

With the above said, I'm not 100% certain this is the (only) issue.

pentschev:
The Numba error is now gone (or not even reached), but there's this one now:

```
20:52:46 TESTS FAILED: dask-cuda-21.06.00a-py37_57.tar.bz2
20:52:49     [gpuci_conda_retry] conda returned exit code: 1
20:52:49     [gpuci_conda_retry] Exiting, no retryable conda errors detected: 'ChecksumMismatchError:' or 'CondaHTTPError:' or 'JSONDecodeError:' or 'ChunkedEncodingError:' or 'CondaMultiError:'
20:52:49     [gpuci_conda_retry] 
20:52:49 Build step 'Execute shell' marked build as failure
20:52:53 [Set GitHub commit status (universal)] ERROR on repos [GHRepository@30668f7b[nodeId=MDEwOlJlcG9zaXRvcnkxNjU5MjY5Nzc=,description=Utilities for Dask and CUDA interactions,homepage=,name=dask-cuda,fork=false,archived=false,size=1100,milestones={},language=Python,commits={},source=<null>,parent=<null>,isTemplate=<null>,url=https://api.github.com/repos/rapidsai/dask-cuda,id=165926977,nodeId=<null>,createdAt=2019-01-15T21:33:16Z,updatedAt=2021-06-07T16:04:05Z]] (sha:9d76971) with context:gpuCI/dask-cuda/build/python/3.7
20:52:53 Setting commit status on GitHub for https://github.com/rapidsai/dask-cuda/commit/9d7697182f2dd4a74da8f8d2d051d44b7a1eacb1
20:52:53 Finished: FAILURE
```

This looks like a CI issue, could you check @ajschmidt8 ?

ajschmidt8:
@pentschev, I'm seeing this error message hidden in the log text. Can you take a look?

image

jakirkham:
@galipremsagar would you be able to look at the issue AJ shared above? 🙂

pentschev:
> I'm seeing this error message hidden in the log text. Can you take a look?

Wait a second, I'm confused now. What's this build doing exactly? This error is because we're picking Dask 2021.06.0, which is exactly what I'm trying to prevent with this PR. AFAIK, 21.06 should pin to Dask 2021.05.1.

pentschev:
@jakirkham I think Prem is good, the problem is that we're picking Dask 2021.06.0, which is not what RAPIDS should be pinning to.

conda/recipes/dask-cuda/meta.yaml
```diff
@@ -31,7 +31,7 @@ requirements:
     - distributed >=2.22.0
```
jakirkham, line 31: I think we may need to constrain `dask` + `distributed` here...

requirements.txt
```diff
@@ -2,4 +2,4 @@ dask>=2.22.0
 distributed>=2.22.0
```
jakirkham, line 2: ...and also here

jakirkham (commented):
If this is related to getting the newer `dask` + `distributed` versions, we likely need to constrain them

conda/recipes/dask-cuda/meta.yaml
pentschev: But do we need to constrain it here? Shouldn't this be happening RAPIDS-wide?

pentschev: Specifically, I was under the impression that https://github.com/rapidsai/integration/blob/37f52deb0acc7826759cef61f2e5109192fee3db/conda/recipes/versions.yaml#L49-L50 would resolve this. Am I wrong?

jakirkham: Yes I think so if we are still seeing that issue.

We do that also

https://github.com/rapidsai/integration/blob/37f52deb0acc7826759cef61f2e5109192fee3db/conda/recipes/versions.yaml#L49-L50
https://github.com/rapidsai/integration/blob/37f52deb0acc7826759cef61f2e5109192fee3db/conda/recipes/versions.yaml#L53-L54

Though I don't think that is working for us here. So we may need to add it ourselves

pentschev: Ok, I'll add here as well, it must probably be by design, as we also release Dask-CUDA PyPI packages we probably don't necessarily want to limit it to the same narrow versions as RAPIDS, but not sure how much that really helps as we don't test against older Dask versions anyway.

ajschmidt8:
> If this is related to getting the newer `dask` + `distributed` versions, we likely need to constrain them

That sounds like it could be the problem. @pentschev, thoughts? Can you try updating your PR to include this?

conda/recipes/dask-cuda/meta.yaml
jakirkham: Yeah makes sense

Good point. Maybe in the next release we can bump the minimum

pentschev:
For some reason, the [test that failed is supposed to be skipped in CUDA < 11.2](https://github.com/rapidsai/dask-cuda/blob/0d6c3b72bc8fdcfd70153e74eb8f20acac48bead/dask_cuda/tests/test_dask_cuda_worker.py#L109-L112) but it ran anyway on the 11.0 build. I'm rerunning the tests and trying to reproduce locally in the meantime.

pentschev:
rerun tests

jakirkham:
It's encouraging to see the other failures went away though :)

pentschev:
The condition was wrong, I think the latest commit will fix the issue.

pentschev:
Everything passes now! @ajschmidt8 could you merge this?

ajschmidt8:
CI is passing. Versions match the integration repo. merging to `21.06`.

PR #656, opened 2021-06-16
Small tweaks to make compatible with dask-mpi

jacobtomlinson:
While working on GPU support in [dask-mpi](https://github.com/dask/dask-mpi) I ran into a few issues with `CUDAWorker`. This PR makes a few tweaks to ensure things are compatible.

1. A valid option for `local_directory` is an empty string representing the current directory. So added a check before trying to create the directory.

2. Integers are valid options for `name` so added an explicit cast to string when concatenating with GPU index.

3. In `dask-mpi` it uses `async with` for creating `Worker` and `Nanny` objects. This broke with `CUDAWorker` because `__aenter__` and `__aexit__` have not been implemented. Looking at distributed those methods are implemented in the base `Server` class, so I've updated `CUDAWorker` here to also use `Server` as a base. 

   I was surprised this wasn't the case already, but perhaps that was an intentional decision? If so it would be quick to implement the async context manager methods directly here instead.

### DIFF
--- a/dask_cuda/__init__.py
+++ b/dask_cuda/__init__.py
@@ -5,8 +5,8 @@ import dask.dataframe.shuffle
 from ._version import get_versions
 from .cuda_worker import CUDAWorker
 from .explicit_comms.dataframe.shuffle import get_rearrange_by_column_tasks_wrapper
-from .proxify_device_objects import proxify_decorator, unproxify_decorator
 from .local_cuda_cluster import LocalCUDACluster
+from .proxify_device_objects import proxify_decorator, unproxify_decorator
 
 __version__ = get_versions()["version"]
 del get_versions
--- a/dask_cuda/cuda_worker.py
+++ b/dask_cuda/cuda_worker.py
@@ -221,10 +221,7 @@ class CUDAWorker(Server):
                 plugins={
                     CPUAffinity(get_cpu_affinity(i)),
                     RMMSetup(
-                        rmm_pool_size,
-                        rmm_managed_memory,
-                        rmm_async,
-                        rmm_log_directory,
+                        rmm_pool_size, rmm_managed_memory, rmm_async, rmm_log_directory,
                     ),
                 },
                 name=name if nprocs == 1 or not name else str(name) + "-" + str(i),
--- a/dask_cuda/cuda_worker.py
+++ b/dask_cuda/cuda_worker.py
@@ -10,6 +10,7 @@ from tornado.ioloop import IOLoop
 
 import dask
 from distributed import Nanny
+from distributed.core import Server
 from distributed.deploy.cluster import Cluster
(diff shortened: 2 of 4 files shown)


madsbk (approved):
LGTM @jacobtomlinson, thanks.

FYI: I am working on the CI error: https://github.com/rapidsai/dask-cuda/pull/658

madsbk:
>     3. I was surprised this wasn't the case already, but perhaps that was an intentional decision? If so it would be quick to implement the async context manager methods directly here instead.

Yeah, that is strange. I don't think there is any reason for that. @pentschev or @quasiben do you know if it is deliberate that `CUDAWorker` doesn't inherit from `distributed.Server`?

pentschev:
> Yeah, that is strange. I don't think there is any reason for that. @pentschev or @quasiben do you know if it is deliberate that `CUDAWorker` doesn't inherit from `distributed.Server`?

I think there's no actual reason for that besides it not being a class that's trying to reproduce the behavior of `Nanny` or `Worker` currently, but simply instantiates `Nanny`. Originally, I think there was only the `dask-cuda-worker` CLI and @jacobtomlinson wrote `CUDAWorker` based on that and changed `dask-cuda-worker` to call `CUDAWorker` instead, IIRC, the reason for that was to ensure we could use it with `SSHCluster`. With that said, I'm not sure whether there's a reason we can't/shouldn't be doing it.

> 3\. Looking at distributed those methods are implemented in the base `Server` class, so I've updated `CUDAWorker` here to also use `Server` as a base.

I haven't looked at it in detail to figure whether there's a difference, but it seems that the inheritance is from `ServerNode` and not `Server` in both [`Nanny`](https://github.com/dask/distributed/blob/main/distributed/nanny.py#L43) and [`Worker`](https://github.com/dask/distributed/blob/main/distributed/worker.py#L199). Are `Server` and `ServerNode` the same?

jacobtomlinson:
> @jacobtomlinson wrote CUDAWorker based on that and changed dask-cuda-worker to call CUDAWorker instead

Heh it's probably my fault that it doesn't inherit from anything. 

> Are Server and ServerNode the same?

`ServerNode` adds some extra methods for starting up web server based services like the dashboard, but as you say we don't need any of that here. My main requirement is that `CUDAWorker` behaves like a `Worker` from an asyncio point of view, which inheriting from `Server` should cover.

pentschev:
rerun tests

pentschev (approved):
It seems that all tests now pass but the old CUDA 11.0 build is not running anymore, leaving CI in a failed state, we probably will need @rapidsai/ops-codeowners to force merge this.

ajschmidt8:
@gpucibot merge

jacobtomlinson:
Thanks @ajschmidt8

PR #658, opened 2021-06-21
get_device_memory_objects(): dispatch on cudf.core.frame.Frame

madsbk:
`get_device_memory_objects()` now dispatch on `cudf.core.frame.Frame` instead of `cudf.core.dataframe.DataFrame` only.

This includes the CI fix, closes https://github.com/rapidsai/dask-cuda/pull/659

PR #659, opened 2021-06-22
CI: cloning distributed before running distributed tests

madsbk:


madsbk:
@quasiben @jakirkham @charlesbluca, any ideas why pytest would complain about `fixture 'cleanup' not found`, when it is defined in `distributed.utils_test` and works in Distributed's CI? It also works on my machine.

pentschev:
@madsbk https://github.com/dask/distributed/issues/4902 , we can't run anymore as in https://github.com/rapidsai/dask-cuda/blob/f3220a4c7121644ef96dab13741e5d52f6a6d4aa/ci/gpu/build.sh#L115-L120, we have to clone the Distributed package and run from the source, as in https://github.com/rapidsai/ucx-py/blob/6547e13c7e5249843c20e2cf4f86fda369f55510/ci/gpu/build.sh#L101-L113 . Could you work on a fix for that in Dask-CUDA?

madsbk:
Thanks @pentschev! Would have taken me ages to fix!

ci/gpu/build.sh
```diff
@@ -111,7 +111,12 @@ else
...
     # Test downstream packages, which requires Python v3.7
     if [ $(python -c "import sys; print(sys.version_info[1])") -ge "7" ]; then
-        gpuci_logger "TEST OF DASK/UCX"
+        # Clone Distributed to avoid pytest cleanup fixture errors
+        # See https://github.com/dask/distributed/issues/4902
+        gpuci_logger "Clone Distributed"
+        git clone https://github.com/dask/distributed
+
+        gpuci_logger "Run Distributed Tests"
         py.test --cache-clear -vs `python -c "import distributed.protocol.tests.test_cupy as m;print(m.__file__)"`
```
pentschev, line 120: You need also to run from the clone and not anymore with `python -c "import ...`, as in https://github.com/rapidsai/ucx-py/blob/6547e13c7e5249843c20e2cf4f86fda369f55510/ci/gpu/build.sh#L107-L113 .

madsbk: Ahh of course thanks!

pentschev (approved):
Thanks @madsbk !

pentschev:
Seems though that there are some failures in the unspill tests though, are these known?

madsbk:
> Seems though that there are some failures in the unspill tests though, are these known?

Yes, fixing everything in https://github.com/rapidsai/dask-cuda/pull/658

PR #671, opened 2021-07-13
Fix NVML index usage in CUDAWorker/LocalCUDACluster

pentschev:
Fix an issue where device index `0` would be used in NVML functions. For CUDA runtime calls, we expect that the GPU being targeted is always on index `0` as its relative to the `CUDA_VISIBLE_DEVICES` ordering. However, NVML relies on absolute indices, thus we have to always use the actual GPU index being targeted, rather than the first one in `CUDA_VISIBLE_DEVICES`.

This is normally not an issue if no `CUDA_VISIBLE_DEVICES` is set, or is just set as `list(",".join(list(str(i) for i in range(get_n_gpus())))`, but it may be an issue when targeting a different list of GPUs. For example on a DGX-1, CPU affinity for GPUs `0-3` is `0-19,40-59`, and for GPUs `4-7` it is `20-39,60-79`, but when the user would set `CUDA_VISIBLE_DEVICES=4` the CPU affinity for the targeted GPU would be the same as for device index `0`, or `0-19,40-59`. This could result in lower performance, as well as wrong computation of total GPU memory for non-homogenous GPU systems.

### DIFF
--- a/dask_cuda/cuda_worker.py
+++ b/dask_cuda/cuda_worker.py
@@ -30,6 +30,7 @@ from .utils import (
     get_n_gpus,
     get_ucx_config,
     get_ucx_net_devices,
+    nvml_device_index,
     parse_device_memory_limit,
 )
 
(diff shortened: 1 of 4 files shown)


pentschev:
This is similar to issues we had not long ago in Distributed, as @charlesbluca would remember. This came up when discussing MIG support with @akaanirban , thanks Anirban for bringing that to my attention.

pentschev:
@beckernick @VibhuJawa this may have affected past performance on DGX A100s, where we had previously to manually launch `dask-cuda-worker` for each GPU individually.

pentschev:
rerun tests

quasiben (approved):

quasiben:
@gpucibot merge

Issue #583, opened 2021-04-19
[BUG] LocalCUDACluster doesn't work with NVIDIA MIG


drobison00:
(py)nvml does not appear to be compatible with MIG, which prevents various Dask services from working correctly, for example 'LocalCUDACluster'.

While this isn't explicitly Dask-cuda's fault, the end result is the same. Adding this issue for others to reference, and for discussion of potential work arounds.

```python
from dask_cuda import LocalCUDACluster
from dask.distributed import Client

cluster = LocalCUDACluster(device_memory_limit=1.0, rmm_managed_memory=True)
client = Client(cluster)
---------------------------------------------------------------------------
NVMLError_NoPermission                    Traceback (most recent call last)
<ipython-input-1-48e0ebf5a2e9> in <module>
     33 
     34 
---> 35 cluster = LocalCUDACluster(device_memory_limit=1.0,
     36                            rmm_managed_memory=True)
     37 client = Client(cluster)

/opt/conda/envs/rapids/lib/python3.8/site-packages/dask_cuda/local_cuda_cluster.py in __init__(self, n_workers, threads_per_worker, processes, memory_limit, device_memory_limit, CUDA_VISIBLE_DEVICES, data, local_directory, protocol, enable_tcp_over_ucx, enable_infiniband, enable_nvlink, enable_rdmacm, ucx_net_devices, rmm_pool_size, rmm_managed_memory, jit_unspill, **kwargs)
    166             memory_limit, threads_per_worker, n_workers
    167         )
--> 168         self.device_memory_limit = parse_device_memory_limit(
    169             device_memory_limit, device_index=0
    170         )

/opt/conda/envs/rapids/lib/python3.8/site-packages/dask_cuda/utils.py in parse_device_memory_limit(device_memory_limit, device_index)
    478         device_memory_limit = float(device_memory_limit)
    479         if isinstance(device_memory_limit, float) and device_memory_limit <= 1:
--> 480             return int(get_device_total_memory(device_index) * device_memory_limit)
    481 
    482     if isinstance(device_memory_limit, str):

/opt/conda/envs/rapids/lib/python3.8/site-packages/dask_cuda/utils.py in get_device_total_memory(index)
    158     """
    159     pynvml.nvmlInit()
--> 160     return pynvml.nvmlDeviceGetMemoryInfo(
    161         pynvml.nvmlDeviceGetHandleByIndex(index)
    162     ).total

[...]

PR #674, opened 2021-07-16
Support for LocalCUDACluster with MIG

akaanirban:
Adds support to start LocalCUDACluster and cuda workers on MIG instances by passing in uuids of the mig instances. Builds off of existing PR #671
More specifically this PR does the following:
1. Allows starting `LocalCUDACluster` as the following: `cluster = LocalCUDACluster(CUDA_VISIBLE_DEVICES=["MIG-uuid1","MIG-uuid2",...])` or by passing them as `,` separated strings. 

Needs Discussion:
0. Apart from manually testing on a MIG instance on the cloud, how would we test this?
1. What if the user does not pass in any argument to `LocalCUDACluster` while using MIG instances? By default `LocalCUDACluster` will try to use all the parent GPUs and run into error.
2. What if we have a deployment with MIG-enabled and non-MIG-enabled GPUs?
3. `dask.distributed` diagnostics will also fail if we run on MIG enabled GPUs since it uses `pynvml` APIS for non-MIG-enabled GPUs only at the moment.

### DIFF
--- a/dask_cuda/tests/test_dask_cuda_worker.py
+++ b/dask_cuda/tests/test_dask_cuda_worker.py
@@ -5,14 +5,14 @@ import subprocess
 
 import pytest
 
-from distributed import Client
+from distributed import Client, wait
 from distributed.system import MEMORY_LIMIT
 from distributed.utils_test import loop  # noqa: F401
 from distributed.utils_test import popen
 
 import rmm
 
-from dask_cuda.utils import get_n_gpus, wait_workers
+from dask_cuda.utils import get_gpu_count_mig, get_n_gpus, wait_workers
 
 _driver_version = rmm._cuda.gpu.driverGetVersion()
 _runtime_version = rmm._cuda.gpu.runtimeGetVersion()
@@ -186,3 +186,53 @@ def test_unknown_argument():
     ret = subprocess.run(["dask-cuda-worker", "--my-argument"], capture_output=True)
     assert ret.returncode != 0
     assert b"Scheduler address: --my-argument" in ret.stderr
+
+
+def test_cuda_mig_visible_devices_and_memory_limit_and_nthreads(loop):  # noqa: F811
+    init_nvmlstatus = os.environ.get("DASK_DISTRIBUTED__DIAGNOSTICS__NVML")
+    try:
+        os.environ["DASK_DISTRIBUTED__DIAGNOSTICS__NVML"] = "False"
+        uuids = get_gpu_count_mig(return_uuids=True)[1]
+        # test only with some MIG Instances assuming the test bed
+        # does not have a huge number of mig instances
+        if len(uuids) > 0:
+            uuids = [i.decode("utf-8") for i in uuids]
+        else:
+            return True
+        CUDA_VISIBLE_DEVICES = ",".join(uuids)
+        os.environ["CUDA_VISIBLE_DEVICES"] = CUDA_VISIBLE_DEVICES
+        nthreads = len(CUDA_VISIBLE_DEVICES)
+        with popen(["dask-scheduler", "--port", "9359", "--no-dashboard"]):
+            with popen(
+                [
+                    "dask-cuda-worker",
+                    "127.0.0.1:9359",
+                    "--host",
+                    "127.0.0.1",
+                    "--nthreads",
+                    str(nthreads),
+                    "--no-dashboard",
+                    "--worker-class",
+                    "dask_cuda.utils.MockWorker",
+                ]
+            ):
+                with Client("127.0.0.1:9359", loop=loop) as client:
+                    assert wait_workers(client, n_gpus=len(uuids))
+                    # Check to see if all workers are up and
+                    # CUDA_VISIBLE_DEVICES cycles properly
+
+                    def get_visible_devices():
+                        return os.environ["CUDA_VISIBLE_DEVICES"]
+
+                    result = client.run(get_visible_devices)
+                    wait(result)
+                    assert all(len(v.split(",")) == len(uuids) for v in result.values())
+                    for i in range(len(uuids)):
+                        assert set(v.split(",")[i] for v in result.values()) == set(
+                            uuids
+                        )
+    finally:
+        if "CUDA_VISIBLE_DEVICES" in os.environ:
+            del os.environ["CUDA_VISIBLE_DEVICES"]
+        if init_nvmlstatus:
+            os.environ["DASK_DISTRIBUTED__DIAGNOSTICS__NVML"] = init_nvmlstatus
--- a/dask_cuda/tests/test_dask_cuda_worker.py
+++ b/dask_cuda/tests/test_dask_cuda_worker.py
@@ -198,7 +198,7 @@ def test_cuda_mig_visible_devices_and_memory_limit_and_nthreads(loop):  # noqa: 
         if len(uuids) > 0:
             uuids = [i.decode("utf-8") for i in uuids]
         else:
-            return True
+            pytest.skip("No MIG devices found")
         CUDA_VISIBLE_DEVICES = ",".join(uuids)
         os.environ["CUDA_VISIBLE_DEVICES"] = CUDA_VISIBLE_DEVICES
         nthreads = len(CUDA_VISIBLE_DEVICES)
--- a/dask_cuda/tests/test_local_cuda_cluster.py
+++ b/dask_cuda/tests/test_local_cuda_cluster.py
@@ -10,7 +10,7 @@ import rmm
 
 from dask_cuda import CUDAWorker, LocalCUDACluster, utils
 from dask_cuda.initialize import initialize
-from dask_cuda.utils import MockWorker
+from dask_cuda.utils import MockWorker, get_gpu_count_mig
 
 _driver_version = rmm._cuda.gpu.driverGetVersion()
 _runtime_version = rmm._cuda.gpu.runtimeGetVersion()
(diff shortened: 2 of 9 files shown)


pentschev:
Thanks @akaanirban for opening this PR. I should add that this builds off of https://github.com/rapidsai/dask-cuda/pull/671 so we will want that merged before this, and as we discussed offline we'll want to have some tests here even if they don't run in CI.

pentschev:
As for style errors I suggest you you `pip install pre-commit` and then run `pre-commit install` inside the dask-cuda directory, and that will automatically fix the issues on new commits, or at least inform what the issues are for flake8 errors.

pentschev:
Also removed my GH handle from the description as if you have your handle there after merged it will forever plague you with mentions whenever there's certain GH actions such as forks and whatnot by other users.

pentschev:
> 0. Apart from manually testing on a MIG instance on the cloud, how would we test this?

As we discussed offline we'll probably need something as follows:

* Programatically find whether there are MIG instances setup, skip tests if not;
* If there are MIG instances, we’ll use all or some of them (which we’ll have to query for UUIDs also programatically) to setup cluster and ensure they work.
 
> 1. What if the user does not pass in any argument to `LocalCUDACluster` while using MIG instances? By default `LocalCUDACluster` will try to use all the parent GPUs and run into error.

Also as discussed offline:

I don’t know if there’s a proper solution to that. I think in a first pass what we’ll want to do is something similar to what was suggested in https://github.com/rapidsai/dask-cuda/issues/583#issuecomment-837485659 , except that we would raise a more user-friendly error, instead of `Insufficient Permissions` we would write something like `NVML failed with Insufficient Permissions, this may be due to MIG instances on the system, make sure you explicitly specify which GPUs are to be used with 'CUDA_VISIBLE_DEVICES'`, and probably document this error and link to that somewhere in https://docs.rapids.ai/api/dask-cuda/nightly/index.html where we would detail what are accepted ways to configure the cluster. Over time we will probably have more usage with MIG systems and then be able to reevaluate better ways to deal with that automatically.

> 2. What if we have a deployment with MIG-enabled and non-MIG-enabled GPUs?

I think we should leave this case for a follow-up discussion, could you file an issue for that?

> 3. `dask.distributed` diagnostics will also fail if we run on MIG enabled GPUs since it uses `pynvml` APIS for non-MIG-enabled GPUs only at the moment.

Also a separate discussion. Once this is in, we should file an issue summarizing the status of that and discuss how to be better handle it.

jacobtomlinson (commented):
We probably need to bump the minimum `pynvml` version to `11.0.0` as that was when MIG support was added.

dask_cuda/tests/test_local_cuda_cluster.py
```diff
@@ -206,3 +206,40 @@ async def test_cluster_worker():
             await new_worker
             await client.wait_for_workers(2)
             await new_worker.close()
+
+
+@gen_test(timeout=20)
+async def test_available_MIG_workers():
```
pentschev, line 212:
```suggestion
async def test_available_mig_workers():
```

dask_cuda/tests/test_dask_cuda_worker.py
```diff
@@ -186,3 +186,53 @@ def test_unknown_argument():
...
+    init_nvmlstatus = os.environ.get("DASK_DISTRIBUTED__DIAGNOSTICS__NVML")
+    try:
+        os.environ["DASK_DISTRIBUTED__DIAGNOSTICS__NVML"] = "False"
+        uuids = get_gpu_count_mig(return_uuids=True)[1]
+        # test only with some MIG Instances assuming the test bed
+        # does not have a huge number of mig instances
+        if len(uuids) > 0:
+            uuids = [i.decode("utf-8") for i in uuids]
+        else:
+            return True
```
pentschev, line 201:
```suggestion
            pytest.skip("No MIG devices found")
```

dask_cuda/tests/test_local_cuda_cluster.py
```diff
@@ -206,3 +206,40 @@ async def test_cluster_worker():
...
+
+    init_nvmlstatus = os.environ.get("DASK_DISTRIBUTED__DIAGNOSTICS__NVML")
+    try:
+        os.environ["DASK_DISTRIBUTED__DIAGNOSTICS__NVML"] = "False"
+        dask.config.refresh()
+        uuids = get_gpu_count_mig(return_uuids=True)[1]
+        if len(uuids) > 0:
+            uuids = [i.decode("utf-8") for i in uuids]
+        else:
+            return True
```
pentschev, line 223:
```suggestion
            pytest.skip("No MIG devices found")
```

dask_cuda/tests/test_utils.py
```diff
@@ -245,3 +245,30 @@ def test_parse_device_memory_limit():
...
+            miguuids = []
+            for i in range(count):
+                try:
+                    mighandle = pynvml.nvmlDeviceGetMigDeviceHandleByIndex(
+                        device=handle, index=i
+                    )
+                    miguuids.append(mighandle)
+                except pynvml.NVMLError:
+                    pass
+            assert len(miguuids) <= 7
```
pentschev, line 274: Why this magical number `7`? What if there are `8` or more?

pentschev: A more accurate assertion would probably be `assert len(miguuids) == count`.

dask_cuda/utils.py
```diff
@@ -130,13 +130,50 @@ def get_gpu_count():
...
+    return_uuids: bool
+        Returns the uuids of the MIG instances available optionally
+
+    """
+    pynvml.nvmlInit()
+    uuids = []
+    for index in range(get_gpu_count()):
+        handle = pynvml.nvmlDeviceGetHandleByIndex(index)
+        try:
+            mode = pynvml.nvmlDeviceGetMigMode(handle)[0]
```
pentschev, line 148:
```suggestion
            is_mig_mode = pynvml.nvmlDeviceGetMigMode(handle)[0]
```

dask_cuda/utils.py
```diff
@@ -130,13 +130,50 @@ def get_gpu_count():
...
+        except pynvml.NVMLError:
+            # if not a MIG device, i.e. a normal GPU, skip
+            continue
+        if mode:
```
pentschev, line 152:
```suggestion
        if is_mig_mode:
```

dask_cuda/utils.py
```diff
@@ -130,13 +130,50 @@ def get_gpu_count():
...
+
+def get_cpu_affinity(device_index=None):
     """Get a list containing the CPU indices to which a GPU is directly connected.
+    Use either the device index or the specified device identifier UUID.
 
     Parameters
     ----------
-    device_index: int
-        Index of the GPU device
+    device_index: int/str
```
pentschev, line 175:
```suggestion
    device_index: int or str
```

dask_cuda/utils.py
```diff
@@ -181,12 +227,17 @@ def get_n_gpus():
...
+    Return total memory of CUDA device with index or with device identifier UUID
     """
     pynvml.nvmlInit()
-    return pynvml.nvmlDeviceGetMemoryInfo(
-        pynvml.nvmlDeviceGetHandleByIndex(index)
-    ).total
+
+    if index and not str(index).isnumeric():
+        # This means index is UUID. This works for both MIG and non-MIG device UUIDs.
+        handle = pynvml.nvmlDeviceGetHandleByUUID(str.encode(str(index)))
```
pentschev, line 236: @rjzamora we discussed some time ago that NVML doesn't respect `CUDA_VISIBLE_DEVICES`, and while that's true, one of the most user-friendly ways to get a MIG device handle is by its UUID. It seems that users would prefer something simpler such as `MIG-0-1` (meaning GPU 0, MIG instance 1, for example), instead of the UUID that looks like `MIG-84fd49f2-48ad-50e8-9f2e-3bf0dfd47ccb`. Do you think this is worth bringing up with the NVML (and potentially CUDA) teams?

dask_cuda/utils.py
```diff
@@ -511,13 +563,24 @@ def nvml_device_index(i, CUDA_VISIBLE_DEVICES):
     1
     >>> nvml_device_index(1, [1,2,3,0])
     2
+    >>> nvml_device_index(1, ["GPU-9baca7f5-0f2f-01ac-6b05-8da14d6e9005",
+            "GPU-9baca7f5-0f2f-01ac-6b05-8da4d6e9hfghfgh","GPU-9baca7f5-0f2f-01ac-6b05-8da14d6e9dfd"])
+    "GPU-9baca7f5-0f2f-01ac-6b05-8da4d6e9hfghfgh"
```
pentschev, line 568:
```suggestion
    >>> nvml_device_index(1, ["GPU-84fd49f2-48ad-50e8-9f2e-3bf0dfd47ccb",
                              "GPU-d6ac2d46-159b-5895-a854-cb745962ef0f",
                              "GPU-158153b7-51d0-5908-a67c-f406bc86be17"])
    "MIG-d6ac2d46-159b-5895-a854-cb745962ef0f"
```

pentschev: Just replacing by actual UUIDs. :)

pentschev (changes requested):
@akaanirban apologies for the long time to review this. I was able to test this as well and everything seems to work as expected. Thanks for the excellent work in this PR, particularly for the clever solutions for the tests. I've left a few minor suggestions/requests there that are mostly cosmetic with one or two exceptions, but otherwise it looks really great!

dask_cuda/tests/test_utils.py
akaanirban: I added 7 because you can only partition a single MIG enabled GPU into atmost 7 independent MIG instances. But there may be less in reality. `nvmlDeviceGetMaxMigDeviceCount` gives us the maximum number of MIG devices/instances that can exist under a given parent NVML device, not the actual number currently present.

pentschev: This is only true today, you never know if this is still going to be the case in a future GPU, so we have to consider that instead of relying on what holds true today.

pentschev:
Thanks @akaanirban for addressing reviews and for all the work!

pentschev:
@gpucibot merge

pentschev (approved):

jakirkham:
Looks like [the recipe requirement]( https://github.com/rapidsai/dask-cuda/blob/7a426e83c644e4d4d15146c56cb1c3d1c55459df/conda/recipes/dask-cuda/meta.yaml#L32 ) wasn't bumped. @akaanirban, can you please send a PR for that?

jakirkham:
Requirements were updated in PR ( https://github.com/rapidsai/dask-cuda/pull/883 )

PR #681, opened 2021-07-22
Deprecation warning for ucx_net_devices='auto' on UCX 1.11+

pentschev:
Add deprecation warning for `ucx_net_devices='auto'` on UCX 1.11+, fallback to UCX's default when that's requested.

Using `ucx_net_devices='auto'` can cause hangs, this is because it would need to be updated to include both the IPoIB and MLX interfaces in `UCX_NET_DEVICES`. Given UCX can now successfully handle automatic CUDA<->IB mapping, -- including on DGX A100, which this option wasn't capable of handling -- there's no reason for us to support that anymore and deprecating it should be the best approach at this time.

### DIFF
--- a/dask_cuda/cuda_worker.py
+++ b/dask_cuda/cuda_worker.py
@@ -25,6 +25,7 @@ from .proxify_host_file import ProxifyHostFile
 from .utils import (
     CPUAffinity,
     RMMSetup,
+    _ucx_111,
     cuda_visible_devices,
     get_cpu_affinity,
     get_n_gpus,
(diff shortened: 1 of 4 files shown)


pentschev:
cc @charlesbluca this should prevent UCX from hanging even when specifying `--ucx-net-devices auto` on UCX 1.11 and newer.

pentschev:
@ayushdg @beckernick @VibhuJawa FYI

quasiben (approved):

quasiben:
@gpucibot merge

PR #683, opened 2021-07-23
Add device memory limit argument to benchmarks

charlesbluca:
Adds a `--device-memory-limit` argument to the benchmark utils, to set a device memory limit on the instantiated `LocalCUDACluster`. This can be helpful when we want to force spilling to host (for context, I am doing this while benchmarking pack/unpack serialization).

### DIFF
--- a/dask_cuda/benchmarks/utils.py
+++ b/dask_cuda/benchmarks/utils.py
@@ -35,6 +35,16 @@ def parse_benchmark_args(description="Generic dask-cuda Benchmark", args_list=[]
         help="Write dask profile report (E.g. dask-report.html)",
     )
     parser.add_argument(
+        "--device-memory-limit",
+        default=None,
+        type=parse_bytes,
+        help="Size of the CUDA device LRU cache, which is used to determine when the "
+        "worker starts spilling to host memory. Can be an integer (bytes), float "
+        "(fraction of total device memory), string (like ``'5GB'`` or ``'5000M'``), or "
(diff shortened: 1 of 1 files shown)


pentschev:
rerun tests

dask_cuda/benchmarks/utils.py
```diff
@@ -211,6 +221,7 @@ def get_cluster_options(args):
             "scheduler_options": {"protocol": args.protocol},
             "worker_module": "dask_cuda.dask_cuda_worker",
             "worker_options": worker_options,
+            "device_memory_limit": args.device_memory_limit,
```
pentschev, line 224: This needs to be set with `worker_options`.

pentschev (changes requested):

dask_cuda/benchmarks/utils.py
```diff
@@ -202,6 +212,8 @@ def get_cluster_options(args):
             worker_options["enable_infiniband"] = ""
         if args.enable_rdmacm:
             worker_options["enable_rdmacm"] = ""
+        if args.device_memory_limit:
+            worker_options["device_memory_limit"] = ""
```
pentschev, line 216: Now it's missing the value to be assigned, which I believe should be `args.device_memory_limit`.

pentschev (changes requested):

pentschev (approved):
LGTM, thanks @charlesbluca !

pentschev:
@gpucibot merge

PR #700, opened 2021-08-09
Add `--benchmark-json` option to all benchmarks

charlesbluca:
Makes `--benchmark-json` an option for all the benchmarks; also refactors the JSON dumping process so that the output file is a line-delimited JSON file, which can be appended to by subsequent runs of the benchmark script. This makes it a lot easier to process the output in Pandas with something like

```python
pd.read_json("benchmark.json", lines=True)
```

Also added the `device_memory_limit` option to the printed output of the scripts, so that it is visible if specified.

### DIFF
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -1,6 +1,7 @@
 import contextlib
 import math
 from collections import defaultdict
+from json import dumps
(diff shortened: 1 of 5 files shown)


pentschev (approved):
Good idea.

LGTM, thanks @charlesbluca !

pentschev:
@gpucibot merge

PR #702, opened 2021-08-12
Reset UCX-Py after rdmacm tests run

pentschev:
Starting a new cluster on the same pytest process after an rdmacm cluster has been used may cause UCX-Py to complain about being already initialized. By resetting UCX-Py after those tests run, that can be prevented.


quasiben (approved):

quasiben:
@gpucibot merge

PR #709, opened 2021-08-23
Set shuffle config default to "tasks"

pentschev:
By default Dask uses partd ("disk") for shuffling, which is unsupported by cuDF/Dask-cuDF. To avoid forcing users to set it on their own, this changes the default to "tasks", unless the explicitly set by the user.

### DIFF
--- a/dask_cuda/initialize.py
+++ b/dask_cuda/initialize.py
@@ -94,6 +94,13 @@ def initialize(
     )
     dask.config.set({"distributed.comm.ucx": ucx_config})
 
+    # partd is used by default for disk-based shuffling, but unsupported by
+    # cuDF/Dask-cuDF. Therefore, we set task-based shuffling instead.
+    try:
+        dask.config.get("shuffle")
+    except KeyError:
+        dask.config.set({"shuffle": "tasks"})
+
 
 @click.command()
 @click.option(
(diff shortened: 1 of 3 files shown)


dask_cuda/initialize.py
```diff
@@ -94,6 +94,13 @@ def initialize(
     )
     dask.config.set({"distributed.comm.ucx": ucx_config})
 
+    # partd is used by default for disk-based shuffling, but unsupported by
+    # cuDF/Dask-cuDF. Therefore, we set task-based shuffling instead.
+    try:
+        dask.config.get("shuffle")
+    except KeyError:
+        dask.config.set({"shuffle": "tasks"})
```
jakirkham, line 102: Maybe we could use [`update_defaults`]( https://docs.dask.org/en/latest/configuration.html#downstream-libraries ) for this?

pentschev: Good idea. Done in https://github.com/rapidsai/dask-cuda/pull/709/commits/f0b386d31eb209d959177aef1384d4851e8c607f .

pentschev:
Seems like this is set by default in https://github.com/dask/distributed/blob/9c92a61ef69df212e56f8bccb380cb872286ab19/distributed/client.py#L706, this PR is unnecessary, closing.

PR #722, opened 2021-09-10
Check if CUDA context was created in distributed.comm.ucx

### LINKED ISSUE #721: Warnings about existing CUDA contexts on dask-cuda cluster startup
On starting up a dask-cuda cluster on a DGX2, I get this warning for every worker:
```
/home/rgelhausen/conda/envs/test/lib/python3.8/site-packages/dask_cuda/initialize.py:22: UserWarning: A CUDA context for device 14 already exists on process ID 3287969. This is often the result of a CUDA-enabled library calling a CUDA runtime function before Dask-CUDA can spawn worker processes. Please make sure any such function calls don't happen at import time or in the global scope of a program.
  warnings.warn(
```

Checking nvidia-smi, I only see one process per GPU, so nothing looks out of the ordinary despite the warning suggesting I may have multiple workers on a single GPU.

pentschev:
Because communications in `Nanny` are initialized before Dask preload plugins, and UCX creates the context directly within its own initializer in Distributed, Dask-CUDA will always think the CUDA context has already been incorrectly initialized when using UCX, which isn't true, with the globals added here Dask-CUDA can verify the CUDA contexts are indeed valid.

Depends on https://github.com/dask/distributed/pull/5308 .

Fixes #721 .

### DIFF
--- a/dask_cuda/initialize.py
+++ b/dask_cuda/initialize.py
@@ -118,10 +118,6 @@ def initialize(
         it is callable. Can be an integer or ``None`` if ``net_devices`` is not
         callable.
     """
-
-    if create_cuda_context:
-        _create_cuda_context()
-
     ucx_config = get_ucx_config(
         enable_tcp_over_ucx=enable_tcp_over_ucx,
         enable_infiniband=enable_infiniband,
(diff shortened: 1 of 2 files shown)


pentschev:
rerun tests

jakirkham (approved):

pentschev:
rerun tests

pentschev:
@gpucibot merge

pentschev:
Thanks @jakirkham for reviewing!

Issue #730, opened 2021-09-21
Implicit array conversion is happening when numpy array ufuncs are being triggered on `ProxyObject`'s


galipremsagar:
When a `ProxyObject` is pass to any numpy array ufunc, they try to implicitly extract `__array__` out of `ProxyObject`:

```python
>>> import cudf
>>> import numpy as np
>>> from dask_cuda.proxy_object import asproxy
>>> s = cudf.Series([1, 2, 3])
>>> y = np.array(100)
>>> np.less_equal(asproxy(s), y)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/nvme/0/pgali/dask-cuda/dask_cuda/proxy_object.py", line 508, in __array__
    return getattr(self._obj_pxy_deserialize(), "__array__")(*args, **kwargs)
  File "/nvme/0/pgali/envs/cudfdev/lib/python3.8/site-packages/cudf/core/frame.py", line 1638, in __array__
    raise TypeError(
TypeError: Implicit conversion to a host NumPy array via __array__ is not allowed, To explicitly construct a GPU array, consider using cupy.asarray(...)
To explicitly construct a host array, consider using .to_array()
```

PR #731, opened 2021-09-21
add __array_ufunc__ to ProxyObject

galipremsagar:
Fixes: #730 

This PR add `__array_ufunc__` to `ProxyObject`

### DIFF
--- a/dask_cuda/proxy_object.py
+++ b/dask_cuda/proxy_object.py
@@ -436,6 +436,15 @@ class ProxyObject:
             else:
                 object.__setattr__(self._obj_pxy_deserialize(), name, val)
 
+    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
+        inputs = tuple(
+            o._obj_pxy_deserialize() if isinstance(o, ProxyObject) else o
+            for o in inputs
+        )
+        return self._obj_pxy_deserialize().__array_ufunc__(
+            ufunc, method, *inputs, **kwargs
+        )
+
     def __str__(self):
         return str(self._obj_pxy_deserialize())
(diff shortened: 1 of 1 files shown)


dask_cuda/proxy_object.py
```diff
@@ -436,6 +436,15 @@ def __setattr__(self, name, val):
             else:
                 object.__setattr__(self._obj_pxy_deserialize(), name, val)
 
+    def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
+        inputs = tuple(
+            o._obj_pxy_deserialize() if isinstance(o, ProxyObject) else o
```
madsbk, line 441: Would it also make sense to call `o._obj_pxy_deserialize()` on each value in `kwargs`?

madsbk (commented):
Nice work @galipremsagar, thanks for the fix. I only have a minor suggestion.

dask_cuda/proxy_object.py
galipremsagar: Right, handling `kwargs` too now.

dask_cuda/proxy_object.py
```diff
@@ -436,6 +436,23 @@ def __setattr__(self, name, val):
...
+            for o in inputs
+        )
+        kwargs = dict(
```
madsbk, line 444: Just a nitpick, I don't think you need the `dict` here.

galipremsagar: removed it 👍

madsbk (approved):

madsbk:
@gpucibot merge

PR #735, opened 2021-09-23
Pin max `dask` and `distributed` versions to `2021.09.1`

galipremsagar:
Changes to be in-line with: https://github.com/rapidsai/cudf/pull/9286

### DIFF
--- a/conda/recipes/dask-cuda/meta.yaml
+++ b/conda/recipes/dask-cuda/meta.yaml
@@ -27,8 +27,8 @@ requirements:
     - setuptools
   run:
     - python
-    - dask >=2.22.0,<=2021.09.1
-    - distributed >=2.22.0,<=2021.09.1
+    - dask=2021.09.1
+    - distributed=2021.09.1
     - pynvml >=8.0.3
     - numpy >=1.16.0
     - numba >=0.53.1
--- a/conda/recipes/dask-cuda/meta.yaml
+++ b/conda/recipes/dask-cuda/meta.yaml
@@ -27,8 +27,8 @@ requirements:
     - setuptools
   run:
     - python
-    - dask >=2.22.0
-    - distributed >=2.22.0
+    - dask >=2.22.0,<=2021.09.1
+    - distributed >=2.22.0,<=2021.09.1
     - pynvml >=8.0.3
     - numpy >=1.16.0
     - numba >=0.53.1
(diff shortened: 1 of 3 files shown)


conda/recipes/dask-cuda/meta.yaml
```diff
@@ -27,8 +27,8 @@ requirements:
     - setuptools
   run:
     - python
-    - dask >=2.22.0
-    - distributed >=2.22.0
+    - dask >=2.22.0,<=2021.09.1
+    - distributed >=2.22.0,<=2021.09.1
```
pentschev, line 31: I think we should pin to _exactly_ `=2021.09.1`. There have been changes that are breaking in older versions, e.g. https://github.com/rapidsai/dask-cuda/pull/722 .

pentschev (changes requested):

conda/recipes/dask-cuda/meta.yaml
pentschev: Any objections @jakirkham @quasiben to my comment above?

jakirkham: Am happy to defer to you here Peter :)

pentschev: Let's then pin to `=2021.09.1`, I don't see any compelling reason to support half-functional versions.

Could you do that @galipremsagar ? And before I forget, we also need to set https://github.com/rapidsai/dask-cuda/blob/28ce42701fdac4bb1f06ca7705885e8f1f657c0b/ci/gpu/build.sh#L36 to `0`. Same with other repos.

galipremsagar:
```suggestion
    - dask=2021.09.1
    - distributed=2021.09.1
```

requirements.txt
```diff
@@ -1,5 +1,5 @@
-dask>=2.22.0
-distributed>=2.22.0
+dask>=2.22.0,<=2021.09.1
+distributed>=2.22.0,<=2021.09.1
```
galipremsagar, line 2:
```suggestion
dask==2021.09.1
distributed==2021.09.1
```

pentschev (approved):
LGTM, thanks @galipremsagar !

pentschev:
@gpucibot merge

pentschev:
I believe we need someone from ops to merge, could you please do that @ajschmidt8 ?

galipremsagar:
@gpucibot merge

PR #756, opened 2021-10-15
Spilling on demand

### LINKED ISSUE #755: [FEA] On demand memory spilling
Currently, Dask and Dask-CUDA has no way of handling OOM errors other then restarting tasks or workers. Instead they spill preemptively based on some very conservative memory thresholds. For instance, most workflows in Dask-CUDA starts spilling when half the GPU memory is in use.

By using a new RMM resource adaptor like https://github.com/rapidsai/rmm/pull/892, we should be able to implement _on demand memory spilling_.

madsbk:
Use https://github.com/rapidsai/rmm/pull/892 to implement spilling on demand. Requires use of [RMM](https://github.com/rapidsai/rmm) and JIT-unspill enabled.

The `device_memory_limit` still works as usual -- when known allocations gets to `device_memory_limit`, Dask-CUDA starts spilling preemptively. However, with this PR it is should be possible to increase `device_memory_limit` significantly since memory spikes will be handled by spilling on demand.

Closes #755

### DIFF
--- a/dask_cuda/proxify_host_file.py
+++ b/dask_cuda/proxify_host_file.py
@@ -442,7 +442,7 @@ class ProxifyHostFile(MutableMapping):
     spill_on_demand: bool or None, default None
         Enables spilling when the RMM memory pool goes out of memory. If ``None``,
         the "spill-on-demand" config value are used, which defaults to True.
-        Notice, enabling this when RMM isn't availabe or used does nothing.
+        Notice, enabling this does nothing when RMM isn't availabe or not used.
     """
 
     # Notice, we define the following as static variables because they are used by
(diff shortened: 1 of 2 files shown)


pentschev:
The failure is an OOM in the new `test_spill_on_demand`. Perhaps we should look for a way to do some sort of mock test? Maybe we can find a way to rewrite RMM's allocation function or the callback function to trigger in a programatic way?

jakirkham:
@gpucibot merge

jakirkham:
Thanks Mads for the PR! Also Peter and Vibhu for the review and testing 😄

EvenOldridge:
Excited to try this out!

jakirkham:
Nightlies are up. So should be good to go Even 😀

PR #757, opened 2021-10-20
Update memory_pause_fraction in test_spill

pentschev:
Changes from https://github.com/dask/distributed/pull/5438 have slightly
modified the behavior for worker spilling options, thus we need to
update memory_pause_fraction to a value that isn't `None`.


pentschev:
rerun tests

dask_cuda/benchmarks/local_cudf_shuffle.py
```diff
@@ -82,6 +82,7 @@ def main(args):
             "ignore", message=".*NVLink.*rmm_pool_size.*", category=UserWarning
         )
 
+        print(cluster_kwargs)
```
quasiben, line 85: Is this leftover from testing or you do want to print kwargs ?

PR #768, opened 2021-10-29
Simplify JIT-unspill and writing docs

madsbk:
Simplify and clean up of JIT-unspill. Particularly, the evict functions are now sharing common code.
Also adding function docs.

### DIFF
--- a/dask_cuda/proxify_host_file.py
+++ b/dask_cuda/proxify_host_file.py
@@ -368,13 +368,6 @@ class ProxyManager:
                 serializer=lambda p: p._pxy_serialize(serializers=("dask", "pickle")),
             )
 
-    def force_evict_from_device(self, nbytes) -> int:
-        return self.evict(
-            nbytes=nbytes,
-            proxies_access=self.get_dev_access_info,
-            serializer=lambda p: p._pxy_serialize(serializers=("dask", "pickle")),
-        )
-
     def maybe_evict_from_host(self, extra_host_mem=0) -> None:
         mem_over_usage = (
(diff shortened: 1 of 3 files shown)


madsbk:
rerun tests

madsbk:
rerun tests

madsbk:
rerun tests

dask_cuda/proxify_host_file.py
```diff
@@ -532,7 +540,17 @@ def fast(self):
         class EvictDummy:
             @staticmethod
             def evict():
-                return None, None, self.manager.force_evict_from_host()
+                # We don't know how much we need to spill but Dask will call evict()
+                # repeatedly until enough is spilled. We ask for 100MB each time.
```
pentschev, line 544: Maybe instead of 100MB we should make it a factor of the total memory, as we do for `device_memory_limit` default?

pentschev (approved):
LGTM @madsbk . I added one suggestion on the default eviction size if you think that's appropriate, otherwise it's good to merge as is. Thanks for the work here!

dask_cuda/proxify_host_file.py
madsbk: Good point, we now try to spill 1% of `memory_limit`

madsbk:
Thanks @pentschev

madsbk:
@gpucibot merge

dask_cuda/proxify_host_file.py
pentschev: Sounds good, thanks for addressing this!

PR #754, opened 2021-10-11
Revert "Temporarily skipping some tests because of a bug in Dask (#753)"

madsbk:
Revert #753 when #746 has been fixed.

PR #770, opened 2021-11-05
Reenable explicit comms tests

madsbk:
The explicit-comms tests are always skipped `sys.version_info.minor < 80`, fixed.
Also adding some type hints

### DIFF
--- a/dask_cuda/proxify_device_objects.py
+++ b/dask_cuda/proxify_device_objects.py
@@ -2,7 +2,7 @@ import functools
 import pydoc
 from collections import defaultdict
 from functools import partial
-from typing import Any, List, MutableMapping
+from typing import List, MutableMapping, TypeVar
 
 import dask
 from dask.utils import Dispatch
@@ -11,6 +11,8 @@ from .proxy_object import ProxyObject, asproxy
 
 dispatch = Dispatch(name="proxify_device_objects")
 ignore_types = None
+
+T = TypeVar("T")
 
 
 def _register_ignore_types():
@@ -49,12 +51,12 @@ def _register_ignore_types():
 
 
 def proxify_device_objects(
-    obj: Any,
+    obj: T,
     proxied_id_to_proxy: MutableMapping[int, ProxyObject] = None,
     found_proxies: List[ProxyObject] = None,
     excl_proxies: bool = False,
     mark_as_explicit_proxies: bool = False,
-):
+) -> T:
     """ Wrap device objects in ProxyObject
 
     Search through `obj` and wraps all CUDA device objects in ProxyObject.
@@ -97,7 +99,7 @@ def proxify_device_objects(
     return ret
 
 
-def unproxify_device_objects(obj: Any, skip_explicit_proxies: bool = False):
+def unproxify_device_objects(obj: T, skip_explicit_proxies: bool = False) -> T:
     """ Unproxify device objects
 
     Search through `obj` and un-wraps all CUDA device objects.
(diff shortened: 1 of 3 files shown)


dask_cuda/proxify_device_objects.py
```diff
@@ -12,6 +12,8 @@
 dispatch = Dispatch(name="proxify_device_objects")
 ignore_types = None
 
+T = TypeVar("T")
```
pentschev, line 15: Sorry for the naive question, but what does `TypeVar("T")` mean? It looks like a `template<class T>` in C++, but I'm not sure I understand why that makes sense here and why it's better than `Any`.

dask_cuda/tests/test_explicit_comms.py
```diff
@@ -20,7 +20,7 @@
 from dask_cuda.utils import get_ucx_config
 
 pytestmark = pytest.mark.skipif(
-    sys.version_info.minor < 80,
+    sys.version_info.minor < 8,
```
pentschev, line 23: It seems this is still needed because of https://github.com/dask/distributed/pull/5380, when that gets merged will we be able to remove this `skipif`?

pentschev (commented):
Hi @madsbk , I've left a few comments that I don't really understand about this PR, could you clarify?

pentschev:
I don't think the failing tests, let's try rerunning.

pentschev:
rerun tests

dask_cuda/proxify_device_objects.py
madsbk: It is basically to say "the input type equals to output type" in functions like:
```python
def unproxify_device_objects(obj: T, skip_explicit_proxies: bool = False) -> T:
```
This way, type checkers like `mypy` can infer the output type of `unproxify_device_objects()`

dask_cuda/tests/test_explicit_comms.py
madsbk: Yes, it is still needed. It is just fixing the typo -- we want to run the test on Python v3.8+ **not** Python v3.80+ :P

pentschev:
That’s weird, both this and https://github.com/rapidsai/dask-cuda/pull/771#issuecomment-961856565 PRs segfaulted on the centos7 build, so it's clear it's not an issue with the PR.

dask_cuda/proxify_device_objects.py
pentschev: Ah sure, that makes sense. Thanks for the clarification!

dask_cuda/tests/test_explicit_comms.py
pentschev: Yes, that makes sense. Will https://github.com/dask/distributed/pull/5380 resolve that, i.e., will we be able to run on Python 3.7 again after it's merged?

pentschev:
rerun tests

jakirkham:
rerun tests

jakirkham:
Sorry Peter 😅

pentschev:
Ha! You're 3 seconds too slow @jakirkham ! 😄

pentschev:
rerun tests

dask_cuda/tests/test_explicit_comms.py
pentschev: https://github.com/dask/distributed/pull/5380 is in now, are we safe removing the `skipif` here?

madsbk:
@gpucibot merge

dask_cuda/tests/test_explicit_comms.py
madsbk: > [dask/distributed#5380](https://github.com/dask/distributed/pull/5380) is in now, are we safe removing the `skipif` here?

True, but let's merge to get the type hints and then merge https://github.com/rapidsai/dask-cuda/pull/754

pentschev (approved):
LGTM, thanks @madsbk !

PR #775, opened 2021-11-11
Avoid proxy object aliasing 

madsbk:
In order detect already proxied object, `proxify_device_objects()` now gets a mapping from proxied objects to their proxy objects.


pentschev (approved):
LGTM, thanks @madsbk !

pentschev:
@gpucibot merge

PR #779, opened 2021-11-12
Tally IDs instead of device buffers directly

madsbk:
Instead of keeping a reference to device memory objects when tallying the device memory usage, we now use IDs and sizes in order to avoid delaying freeing.

### DIFF
--- a/dask_cuda/get_device_memory_objects.py
+++ b/dask_cuda/get_device_memory_objects.py
@@ -10,7 +10,7 @@ class DeviceMemoryId:
     """ID and size of device memory objects
 
     Instead of keeping a reference to device memory objects this class
-    only save the id and size in order to avoid delayed freeing.
+    only saves the id and size in order to avoid delayed freeing.
     """
 
     def __init__(self, obj: object):
--- a/dask_cuda/get_device_memory_objects.py
+++ b/dask_cuda/get_device_memory_objects.py
@@ -10,7 +10,7 @@ class DeviceMemoryId:
     """ID and size of device memory objects
 
     Instead of keeping a reference to device memory objects this class
-    only save the id and size in order to avoid delaying freeing of object.
+    only save the id and size in order to avoid delayed freeing.
     """
 
     def __init__(self, obj: object):
--- a/dask_cuda/get_device_memory_objects.py
+++ b/dask_cuda/get_device_memory_objects.py
@@ -1,16 +1,34 @@
+from typing import Set
+
 from dask.sizeof import sizeof
 from dask.utils import Dispatch
 
 dispatch = Dispatch(name="get_device_memory_objects")
 
 
-def get_device_memory_objects(obj) -> set:
-    """ Find all CUDA device objects in `obj`
+class DeviceMemoryId:
+    """ID and size of device memory objects
+
+    Instead of keeping a reference to device memory objects this class
+    only save the id and size in order to avoid delaying freeing of object.
+    """
+
+    def __init__(self, obj: object):
+        self.id = id(obj)
+        self.nbytes = sizeof(obj)
+
+    def __hash__(self) -> int:
+        return self.id
+
+    def __eq__(self, o) -> bool:
+        return self.id == hash(o)
+
+
+def get_device_memory_ids(obj) -> Set[DeviceMemoryId]:
+    """Find all CUDA device objects in `obj`
 
     Search through `obj` and find all CUDA device objects, which are objects
     that either are known to `dispatch` or implement `__cuda_array_interface__`.
-
-    Notice, the CUDA device objects must be hashable.
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


dask_cuda/get_device_memory_objects.py
```diff
@@ -1,28 +1,46 @@
...
 dispatch = Dispatch(name="get_device_memory_objects")
 
 
-def get_device_memory_objects(obj) -> set:
-    """ Find all CUDA device objects in `obj`
+class DeviceMemoryId:
+    """ID and size of device memory objects
+
+    Instead of keeping a reference to device memory objects this class
+    only save the id and size in order to avoid delayed freeing.
```
pentschev, line 13:
```suggestion
    only saves the id and size in order to avoid delayed freeing.
```

dask_cuda/proxify_host_file.py
```diff
@@ -105,12 +127,19 @@ class ProxiesOnHost(Proxies):
...
 
-    def mem_usage_remove(self, proxy: ProxyObject):
+    def mem_usage_remove(self, proxy: ProxyObject) -> None:
         self._mem_usage -= sizeof(proxy)
 
+    def buffer_info(self) -> List[Tuple[float, int, List[ProxyObject]]]:
+        ret = []
+        for p in self.get_proxies():
+            size = sizeof(p)
+            ret.append((p._pxy_get().last_access, size, [p]))
```
pentschev, line 140: Why is a `List[ProxyObject]` needed here instead of just `ProxyObject`? It seems like it's always a single-element list.

dask_cuda/proxify_host_file.py
```diff
@@ -123,31 +152,45 @@ class ProxiesOnDevice(Proxies):
...
         proxy_id = id(proxy)
         assert proxy_id not in self.proxy_id_to_dev_mems
         self.proxy_id_to_dev_mems[proxy_id] = set()
-        for dev_mem in proxy._pxy_get_device_memory_objects():
+        for dev_mem in get_device_memory_ids(proxy._pxy_get().obj):
             self.proxy_id_to_dev_mems[proxy_id].add(dev_mem)
             ps = self.dev_mem_to_proxy_ids[dev_mem]
             if len(ps) == 0:
-                self._mem_usage += sizeof(dev_mem)
+                self._mem_usage += dev_mem.nbytes
```
pentschev, line 176: Why is `sizeof` not suitable here anymore?

dask_cuda/proxify_host_file.py
```diff
@@ -123,31 +152,45 @@ class ProxiesOnDevice(Proxies):
...
 
-    def mem_usage_remove(self, proxy: ProxyObject):
+    def mem_usage_remove(self, proxy: ProxyObject) -> None:
         proxy_id = id(proxy)
         for dev_mem in self.proxy_id_to_dev_mems.pop(proxy_id):
             self.dev_mem_to_proxy_ids[dev_mem].remove(proxy_id)
             if len(self.dev_mem_to_proxy_ids[dev_mem]) == 0:
                 del self.dev_mem_to_proxy_ids[dev_mem]
-                self._mem_usage -= sizeof(dev_mem)
+                self._mem_usage -= dev_mem.nbytes
```
pentschev, line 185: Same question as above.

dask_cuda/proxify_host_file.py
```diff
@@ -123,31 +152,45 @@ class ProxiesOnDevice(Proxies):
...
+
+    def buffer_info(self) -> List[Tuple[float, int, List[ProxyObject]]]:
+        ret = []
+        for dev_mem, proxy_ids in self.dev_mem_to_proxy_ids.items():
+            proxies = self.get_proxies_by_ids(proxy_ids)
+            last_access = max(p._pxy_get().last_access for p in proxies)
+            ret.append((last_access, dev_mem.nbytes, proxies))
```
pentschev, line 192: Unlike in the previous `buffer_info`, the list here seems to be made of multiple objects. I'm guessing you simply want to make the previous definition compatible?

pentschev (approved):
Thanks @madsbk , it looks good to me in general. I've added a minor typo suggestion and a few questions only.

dask_cuda/proxify_host_file.py
madsbk: This is just for performance, now that we know that `dev_mem` is a `class DeviceMemoryId` it is faster to query the size directly instead of implementing `DeviceMemoryId.__sizeof__()`.

dask_cuda/proxify_host_file.py
madsbk: Yes, this is to standardize the interface, now we have `buffer_info` instead of `get_dev_access_info` and `get_host_access_info`

pentschev:
LGTM, thanks @madsbk !

pentschev:
@gpucibot merge

PR #780, opened 2021-11-12
Fix parsing of device by UUID

pentschev:
With this change, specifying GPUs via their UUIDs in `CUDA_VISIBLE_DEVICES` should work again. Add tests for `LocalCUDACluster`/`dask-cuda-worker`.

### DIFF
--- a/dask_cuda/tests/test_dask_cuda_worker.py
+++ b/dask_cuda/tests/test_dask_cuda_worker.py
@@ -12,7 +12,12 @@ from distributed.utils_test import popen
 
 import rmm
 
-from dask_cuda.utils import get_gpu_count_mig, get_n_gpus, wait_workers
+from dask_cuda.utils import (
+    get_gpu_count_mig,
+    get_gpu_uuid_from_index,
+    get_n_gpus,
+    wait_workers,
+)
 
 _driver_version = rmm._cuda.gpu.driverGetVersion()
 _runtime_version = rmm._cuda.gpu.runtimeGetVersion()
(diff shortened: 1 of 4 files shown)


dask_cuda/tests/test_local_cuda_cluster.py
```diff
@@ -245,3 +245,15 @@ def get_visible_devices():
...
+@gen_test(timeout=20)
+async def test_gpu_uuid():
+    async with LocalCUDACluster(
+        CUDA_VISIBLE_DEVICES=get_gpu_uuid_from_index(0),
+        scheduler_port=0,
+        asynchronous=True,
+    ) as cluster:
+        assert len(cluster.workers) == 1
+        async with Client(cluster, asynchronous=True) as client:
+            await client.wait_for_workers(1)
```
madsbk, line 259: For completeness, maybe also check that the correct `CUDA_VISIBLE_DEVICES` is set on the worker?

dask_cuda/tests/test_dask_cuda_worker.py
```diff
@@ -236,3 +241,28 @@ def get_visible_devices():
...
+                    "dask_cuda.utils.MockWorker",
+                ]
+            ):
+                with Client("127.0.0.1:9359", loop=loop) as client:
+                    assert wait_workers(client, n_gpus=1)
+
+                    result = client.run(lambda: os.environ["CUDA_VISIBLE_DEVICES"])
+                    assert list(result.values())[0] == gpu_uuid
+    finally:
+        del os.environ["CUDA_VISIBLE_DEVICES"]
```
madsbk, line 268: Use `unittest.mock.patch` to avoid deleting `os.environ["CUDA_VISIBLE_DEVICES"]` if it already exist on the system.
Python [docs](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch.dict) has this example:
```python
import os
with patch.dict('os.environ', {'newkey': 'newvalue'}):
    print(os.environ['newkey'])
assert 'newkey' not in os.environ
```

madsbk (changes requested):
Looks good @pentschev, have some minor suggestions

dask_cuda/tests/test_local_cuda_cluster.py
pentschev: Good idea, done in https://github.com/rapidsai/dask-cuda/pull/780/commits/839d03a0f4c6cb1fc6f18649f74912c87dc79690 .

dask_cuda/tests/test_dask_cuda_worker.py
pentschev: That's really cool, I wasn't aware of that. Done in https://github.com/rapidsai/dask-cuda/pull/780/commits/e1070110fa9924dcb9d923f95e6acb8f28246f55 . We have several other tests that can benefit from this, I'll update those in a follow-up PR.

pentschev:
rerun tests

madsbk (approved):

pentschev:
Thanks @madsbk for the review!

pentschev:
@gpucibot merge

PR #787, opened 2021-11-15
Use unittest.mock.patch for all os.environ tests

pentschev:
Makes tests depending on `os.environ` settings less error-prone by applying changes via `unittest.mock.patch` that are automatically cleaned up.


madsbk (approved):
LGTM

pentschev:
Thanks @madsbk for the review!

pentschev:
@gpucibot merge

PR #790, opened 2021-11-16
Add Richardson-Lucy deconvolution benchmark

pentschev:


dask_cuda/benchmarks/local_cupy.py
```diff
@@ -133,6 +139,41 @@ async def _run(client, args):
...
+                dtype=a.dtype,
+                weights=f,
+                mode=mode,
+            )
+
+        def _richardson_lucy(image, psf, im_deconv, psf_mirror):
+            conv = _convolve(im_deconv, psf, mode="constant")
+            relative_blur = image / conv
+            im_deconv *= _convolve(relative_blur, psf_mirror, mode="constant")
+            return im_deconv
```
jakirkham, line 170: We have a [`richardson_lucy` implementation in cuCIM]( https://github.com/rapidsai/cucim/blob/9e346db3a2c9f3a5be931b614f16d31d0bd88cc6/python/cucim/src/cucim/skimage/restoration/deconvolution.py#L375 ), which we could use here

Something worth noting is Richardson-Lucy is [an iterative algorithm]( https://github.com/rapidsai/cucim/blob/9e346db3a2c9f3a5be931b614f16d31d0bd88cc6/python/cucim/src/cucim/skimage/restoration/deconvolution.py#L422 ) that converges on a solution. This involves entering and leaving Fourier space repeatedly. So there is a fair bit of computation, which may affect profiling.

pentschev: Thanks for pointing that out, John! I was trying to reproduce https://github.com/nv-legate/cunumeric/blob/18792f3e988e3240eb10ff6de6d78de7df57d090/examples/richardson_lucy.py#L28-L41 , but I now see the mistake I've made in not iterating over `im_deconv` but rather overwriting it. I'll also take a closer look at the cuCIM implementation and see what I can make up of both approaches.

jakirkham: Ofc! Yeah that makes sense. Feel free to grab that code from cuCIM if it helps.

Should add the `convolve` call there is using some vendored code, but that preceded [CuPy adding `convolve` in 9.0.0]( https://github.com/cupy/cupy/blob/8de4cf8b0dc19199366c3ed839f3ee96f6d182ae/cupyx/scipy/signal/signaltools.py#L11 ). So it should be possible to use CuPy directly for that call. Everything else is also straight CuPy so that should hopefully make it easier to use.

The other interesting thing about this `convolve` call is it will try to do convolution in Fourier space or real space depending on which is faster (using some heuristic). If you determine one is faster for your needs, it may be worth bypassing that autodetection logic and just calling with the appropriate implementation.

One last thought since it seems in their benchmark they used a warm-up run, we might want to consider doing the same thing. After all CuPy will create the kernels on the first run. So it only seems fair to do the same thing here.

PR #791, opened 2021-11-17
Disk IO interface

madsbk:
This PR standardize the API for spilling to disk and avoid a in-memory copy before writing to file.

This is part of the work to support spilling using GPUDirect Storage (GDS).

### DIFF
--- a/dask_cuda/disk_io.py
+++ b/dask_cuda/disk_io.py
@@ -1,22 +1,52 @@
-from typing import Mapping
+from typing import Iterable, Mapping
 
 from distributed.utils import nbytes
 
 
-def disk_write(path, frames, shared_filesystem: bool) -> dict:
-    frame_lengths = tuple(map(nbytes, frames))
+def disk_write(path: str, frames: Iterable, shared_filesystem: bool) -> dict:
+    """Write frames to disk
+
+    Parameters
+    ----------
+    path: str
+        File path
+    frames: Iterable
+        The frames to write to disk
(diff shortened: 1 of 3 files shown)


pentschev (commented):
LGTM @madsbk . Just one question, do you think it's worth adding a regression test for `disk_read`/`disk_write`?

jakirkham:
Have you thought about adding the more basic components to [zict]( https://github.com/dask/zict )? Maybe extending [`File`]( https://github.com/dask/zict/blob/b760997b98aad9aca8ab37fc4e8b5bb06809283d/zict/file.py#L22 )?

madsbk:
> LGTM @madsbk . Just one question, do you think it's worth adding a regression test for `disk_read`/`disk_write`?

I will add tests in the follow up PR that implements GDS. For now, I think the existing spill-to-disk tests are sufficient.

madsbk:
> Have you thought about adding the more basic components to [zict](https://github.com/dask/zict)? Maybe extending [`File`](https://github.com/dask/zict/blob/b760997b98aad9aca8ab37fc4e8b5bb06809283d/zict/file.py#L22)?

Yes, when the implementation of JIT-unspill stabilize (GDS support, spill-on-demand, multiple disk locations, and async spilling), I think it makes a lot of sense to port some (or all) of it upstream to Dask.

pentschev (approved):
LGTM, thanks @madsbk !

pentschev:
@gpucibot merge

PR #800, opened 2021-11-22
Fix spill-to-disk triggered by Dask explicitly

madsbk:
The evict didn't use the `buffer_info` introduced in #779 and the test didn't capture the failure. This PR fixes both issues.

Notice, I think this is small and important enough to go into the `v21.12` release.


pentschev:
rerun tests

pentschev (approved):
LGTM, thanks for this fix @madsbk !

pentschev:
@gpucibot merge

PR #802, opened 2021-11-22
Fixed `ProxyObject.__del__` to use the new Disk IO API from #791

madsbk:
Anyone that knows how to make a `pytest` that tests exceptions in class destructors?

### DIFF
--- a/dask_cuda/proxy_object.py
+++ b/dask_cuda/proxy_object.py
@@ -378,7 +378,7 @@ class ProxyObject:
         pxy.manager.remove(self)
         if pxy.serializer == "disk":
             header, _ = pxy.obj
-            os.remove(header["path"])
+            os.remove(header["disk-io-header"]["path"])
 
     def _pxy_serialize(
         self, serializers: Iterable[str], proxy_detail: ProxyDetail = None,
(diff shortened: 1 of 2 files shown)


pentschev:
rerun tests

pentschev:
> Anyone that knows how to make a `pytest` that tests exceptions in class destructors?

@madsbk what exactly do you want to test? You want to ensure no exceptions are raised in the destructor, or do you want to verify that the exception is indeed raised? If it's the latter, during `__del__` no exceptions should be raised and thus there's a design issue.

madsbk:
This: `You want to ensure no exceptions are raised in the destructor`.
I have tried with the `capsys` and `capfd` fixtures but it doesn't work.

It is not blocking this PR but it would be nice to have a regression test

pentschev:
@madsbk we can add `error::pytest.PytestUnraisableExceptionWarning` to https://github.com/rapidsai/dask-cuda/blob/branch-21.12/pytest.ini , that will error on uncatched exceptions in `__del__`.

pentschev (approved):
LGTM @madsbk . It's up to you whether you want to add `error::pytest.PytestUnraisableExceptionWarning`, or maybe even `error::pytest.PytestWarning`. I think it wouldn't hurt for us to be more strict on pytest warnings.

madsbk:
> LGTM @madsbk . It's up to you whether you want to add `error::pytest.PytestUnraisableExceptionWarning`, or maybe even `error::pytest.PytestWarning`. I think it wouldn't hurt for us to be more strict on pytest warnings.

Thanks @pentschev, let's merge this PR and then try a more strict warning policy in another PR.

madsbk:
@gpucibot merge

PR #807, opened 2021-11-22
Remove automatic `doc` labeler

pentschev:
If there's another category label in a PR, everytime a new commit is pushed GitHub Actions will add a new `doc` category that will cause the label checker to fail. Therefore, we remove the `doc` automatic labeler.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -5,9 +5,6 @@
 python:
  - 'dask_cuda/**'
 
-doc:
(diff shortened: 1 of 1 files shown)


pentschev:
@ajschmidt8 doing this as per our discussion offline.

ajschmidt8 (approved):

ajschmidt8:
@gpucibot merge

PR #813, opened 2021-11-25
Fix skipping GDS test if cucim is not installed

pentschev:
When `ProxifyHostFile.register_disk_spilling()` is configured with `gds=True` an `ImportError` is raised and the test fails. To fix that we force enabling cucim (if available) after disk spilling is configured, or skip the test if cucim is unavailable.

### DIFF
--- a/dask_cuda/tests/test_gds.py
+++ b/dask_cuda/tests/test_gds.py
@@ -23,8 +23,8 @@ def test_gds(gds_enabled, cuda_lib):
         ProxifyHostFile.register_disk_spilling()
         if gds_enabled and not ProxifyHostFile._gds_enabled:
             pytest.importorskip("cucim.clara.filesystem")
-            # In this case, we know that cucim and for testing we
-            # force enabling cucim even if GDS is unavailable.
+            # In this case, we know that cucim is available and for testing
+            # we enable cucim explicitly even if GDS is unavailable.
             ProxifyHostFile._gds_enabled = True
 
         a = data_create()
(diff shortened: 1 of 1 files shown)


dask_cuda/tests/test_gds.py
```diff
@@ -20,12 +20,11 @@ def test_gds(gds_enabled, cuda_lib):
...
     try:
-        ProxifyHostFile.register_disk_spilling(gds=gds_enabled)
+        ProxifyHostFile.register_disk_spilling()
         if gds_enabled and not ProxifyHostFile._gds_enabled:
             pytest.importorskip("cucim.clara.filesystem")
-            # In this case, we know that cucim is available but GDS
-            # isn't installed on the system. For testing, we enable
-            # the use of the cucim anyways.
+            # In this case, we know that cucim and for testing we
+            # force enabling cucim even if GDS is unavailable.
```
madsbk, line 27:
```suggestion
            # In this case, we know that cucim is available and for testing
            # we enable cucim explicitly even if GDS is unavailable.
```

madsbk (changes requested):
Looks good, only have a minor rewording.

dask_cuda/tests/test_gds.py
pentschev: Thanks for the suggestion @madsbk , done in https://github.com/rapidsai/dask-cuda/pull/813/commits/a54d6d06622017bae57d4d42fe8742a01bc6b821 .

pentschev:
@madsbk could you approve if there are no other changes to be done?

madsbk (approved):

madsbk:
@gpucibot merge

pentschev:
Thanks @madsbk !

PR #818, opened 2021-12-08
Query UCX-Py from gpuCI versioning service

pentschev:


ajschmidt8 (commented):
As discussed in our Slack thread, we should move the `curl` call to the script below (which gets run every time a new release branch is created) in order to minimize the number of network requests in CI. Similarly to `cudf` ([link](https://github.com/rapidsai/cudf/blob/2e95fb19eb2876110617af5b564287c748915087/ci/release/update-version.sh#L28-L64)), we can then use some `sed` commands to find and replace any hardcoded values that need to be updated in the repo.


https://github.com/rapidsai/dask-cuda/blob/branch-22.02/ci/release/update-version.sh

pentschev:
@ajschmidt8 I updated scripts to reflect the change you suggested, could you take a look to ensure I did everything correctly?

ci/release/update-version.sh
```diff
@@ -17,11 +17,13 @@ CURRENT_MAJOR=$(echo $CURRENT_TAG | awk '{split($0, a, "."); print a[1]}')
 CURRENT_MINOR=$(echo $CURRENT_TAG | awk '{split($0, a, "."); print a[2]}')
 CURRENT_PATCH=$(echo $CURRENT_TAG | awk '{split($0, a, "."); print a[3]}')
 CURRENT_SHORT_TAG=${CURRENT_MAJOR}.${CURRENT_MINOR}
+CURRENT_UCXPY_VERSION=$(curl -s https://version.gpuci.io/rapids/${CURRENT_SHORT_TAG}`.*)
```
Ethyling, line 20: Is this needed?

ci/release/update-version.sh
```diff
@@ -17,11 +17,13 @@ CURRENT_MAJOR=$(echo $CURRENT_TAG | awk '{split($0, a, "."); print a[1]}')
...
 
 #Get <major>.<minor> for next version
 NEXT_MAJOR=$(echo $NEXT_FULL_TAG | awk '{split($0, a, "."); print a[1]}')
 NEXT_MINOR=$(echo $NEXT_FULL_TAG | awk '{split($0, a, "."); print a[2]}')
 NEXT_SHORT_TAG=${NEXT_MAJOR}.${NEXT_MINOR}
+NEXT_UCXPY_VERSION=$(curl -s https://version.gpuci.io/rapids/${NEXT_SHORT_TAG}`.*)
```
Ethyling, line 26: I think this should be `NEXT_UCXPY_VERSION="$(curl -s https://version.gpuci.io/rapids/${NEXT_SHORT_TAG}).*" `

ci/release/update-version.sh
```diff
@@ -30,5 +32,5 @@ function sed_runner() {
     sed -i.bak ''"$1"'' $2 && rm -f ${2}.bak
 }
 
-
-# No-op
\ No newline at end of file
+# Update UCX-Py version
+sed_runner 's/export UCXPY_VERSION=.*/export UCXPY_VERSION='${NEXT_UCXPY_VERSION}'/g' ci/gpu/build.sh
```
Ethyling, line 36: I think we should use `"` instead of `'` to expand the variable

ajschmidt8:
> I updated scripts to reflect the change you suggested, could you take a look to ensure I did everything correctly?

Sorry, late to reply here. Looks like Jordan's got you covered!

ci/release/update-version.sh
pentschev: Not currently, I had it for completeness but I removed now since it's unused at this time.

ci/release/update-version.sh
pentschev: Agreed, I now see there were two typos, your suggestion should fix both. Thanks @Ethyling !

ci/release/update-version.sh
pentschev: No objections from me, I had copied the format from cuDF's script, but I now see it mixes both formats. For example https://github.com/rapidsai/cudf/blob/67c925c84aa111ad3a54c0352c18c570d17f5d75/ci/release/update-version.sh#L34 and https://github.com/rapidsai/cudf/blob/67c925c84aa111ad3a54c0352c18c570d17f5d75/ci/release/update-version.sh#L58

pentschev:
Sorry for the late reply here as well, I think all suggestions are now addressed. Would you mind taking another look when you have the chance @Ethyling @ajschmidt8 ?

pentschev:
rerun tests

pentschev:
rerun tests

ajschmidt8 (approved):

pentschev:
Forgot to merge this, thanks @ajschmidt8 and @Ethyling for reviews!

pentschev:
@gpucibot merge

PR #820, opened 2021-12-09
Handle explicitly disabled UCX transports

pentschev:
When UCX transports are explicitly disabled by the user, the other ones must be implicitly enabled. Since Distributed currently considers disabled (`False`) and default (`None`) similarly, we need to workaround that so that the user's choice prevail.


quasiben (approved):

quasiben:
@gpucibot merge

PR #834, opened 2022-01-26
No Nannies

quasiben:
Add `--no-nanny` option to CLI

### DIFF
--- a/dask_cuda/cuda_worker.py
+++ b/dask_cuda/cuda_worker.py
@@ -214,7 +214,7 @@ class CUDAWorker(Server):
                     loop=loop,
                     resources=resources,
                     memory_limit=memory_limit,
-                    interface=_get_interface(interface, host, i, net_devices),
+                    interface=interface,
                     host=host,
                     preload=(list(preload) or []) + ["dask_cuda.initialize"],
                     preload_argv=(list(preload_argv) or []) + ["--create-cuda-context"],
@@ -232,7 +232,9 @@ class CUDAWorker(Server):
                             rmm_managed_memory,
                             rmm_async,
                             rmm_log_directory,
+                            rmm_track_allocations,
                         ),
+                        PreImport(pre_import),
                     },
                     name=name
                     if nprocs == 1 or name is None
@@ -244,8 +246,6 @@ class CUDAWorker(Server):
                             enable_infiniband=enable_infiniband,
                             enable_nvlink=enable_nvlink,
                             enable_rdmacm=enable_rdmacm,
-                            net_devices=net_devices,
-                            cuda_device_index=i,
                         )
                     },
                     data=data(nvml_device_index(i, cuda_visible_devices(i))),
@@ -268,7 +268,7 @@ class CUDAWorker(Server):
                     loop=loop,
                     resources=resources,
                     memory_limit=memory_limit,
-                    interface=_get_interface(interface, host, i, net_devices),
+                    interface=interface,
                     host=host,
                     preload=(list(preload) or []) + ["dask_cuda.initialize"],
                     preload_argv=(list(preload_argv) or []) + ["--create-cuda-context"],
@@ -285,7 +285,9 @@ class CUDAWorker(Server):
                             rmm_managed_memory,
                             rmm_async,
                             rmm_log_directory,
+                            rmm_track_allocations,
                         ),
+                        PreImport(pre_import),
                     },
                     name=name
                     if nprocs == 1 or name is None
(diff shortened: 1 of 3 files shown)


dask_cuda/cuda_worker.py
```diff
@@ -221,53 +221,102 @@ def del_pid_file():
...
+                    **kwargs,
+                )
+                for i in range(nprocs)
+            ]
+        else:
+            t = Worker
+
+            self.nannies = [
+                t(
+                    scheduler,
```
jakirkham, line 283: Are we able to reuse this for the `if` block above as well? If so, can we move this out of the branch and call once?

quasiben:
[lastephey](https://github.com/lastephey) if you have time can you try this PR with the `--no-nanny` option on the worker and see if this works along with the env var `UCX_IB_FORK_INIT=n` for scaled tests ?

dask_cuda/cuda_worker.py
pentschev: +1 to John's comment. I think what you need is:

```python
t = Nanny if nanny else Worker

self.nannies = [...]
```

dask_cuda/cuda_worker.py
```diff
@@ -200,53 +200,104 @@ def del_pid_file():
...
-                },
-                data=data(nvml_device_index(i, cuda_visible_devices(i))),
-                worker_class=worker_class,
-                **kwargs,
-            )
-            for i in range(nprocs)
-        ]
+        if nanny:
+            t = Nanny
+            kwargs = {"worker_port": None, "listen_address": None, **kwargs}
```
pentschev, line 205: It isn't immediately clear to me, why is this line necessary?

dask_cuda/tests/test_dask_cuda_worker.py
```diff
@@ -294,6 +294,28 @@ def test_cuda_visible_devices_uuid(loop):  # noqa: F811
...
+                "--no-dashboard",
+                "--no-nanny",
+            ],
+            flush_output=False,
+        ) as worker:
+            with Client("127.0.0.1:9069", loop=loop) as client:
+                assert wait_workers(client, n_gpus=1)
+                assert any(
+                    b"distributed.nanny" not in worker.stdout.readline()
+                    for i in range(5)
```
pentschev, line 315: What does the magic number `5` mean here?

pentschev (changes requested):
Looks mostly good to me, just have one cleanup request and a couple of questions.

dask_cuda/tests/test_dask_cuda_worker.py
quasiben: This is to capture the first few lines and confirm nanny doesn't show up. For example, with nannies:

```
2022-06-14 05:59:22,053 - distributed.nanny - INFO -         Start Nanny at: 'tcp://10.33.227.162:36881'
2022-06-14 05:59:24,289 - distributed.preloading - INFO - Import preload module: dask_cuda.initialize
2022-06-14 05:59:25,043 - distributed.preloading - INFO - Run preload setup click command: dask_cuda.initialize
2022-06-14 05:59:25,043 - distributed.worker - INFO -       Start worker at:  tcp://10.33.227.162:33455
2022-06-14 05:59:25,044 - distributed.worker - INFO -          Listening to:  tcp://10.33.227.162:33455
```

without nannies:

```
(dask-dev) bzaitlen@dgx12:/datasets/bzaitlen/GitRepos/dask-cuda$ CUDA_VISIBLE_DEVICES=0 dask-cuda-worker tcp://10.33.227.162:8786 --no-nanny
2022-06-14 05:59:47,314 - distributed.preloading - INFO - Import preload module: dask_cuda.initialize
2022-06-14 05:59:47,960 - distributed.preloading - INFO - Run preload setup click command: dask_cuda.initialize
2022-06-14 05:59:47,961 - distributed.worker - INFO -       Start worker at:  tcp://10.33.227.162:42383
2022-06-14 05:59:47,961 - distributed.worker - INFO -          Listening to:  tcp://10.33.227.162:42383
```

I can add a comment to explain this

dask_cuda/cuda_worker.py
quasiben: I think the `worker_class` arg is something not supported in the `Worker` object.  I'll look again since it's been awhile and see if I can refactor as you are both suggesting.

We should probably do something closer to the following:
https://github.com/dask/distributed/blob/344868ace6cd10f7b99e4cd2d3830fba66a2cdcf/distributed/cli/dask_worker.py#L427-L433

As for `worker_port` and `listen_address` these are both supported by dask-workers but currently not fully support in dask-cuda-workers

pentschev: > I think the `worker_class` arg is something not supported in the `Worker` object.

Right, but here it is being passed to `Nanny`, not `Worker`.

> I'll look again since it's been awhile and see if I can refactor as you are both suggesting.
> 
> We should probably do something closer to the following: https://github.com/dask/distributed/blob/344868ace6cd10f7b99e4cd2d3830fba66a2cdcf/distributed/cli/dask_worker.py#L427-L433

Agreed, this is similar to what I suggested in https://github.com/rapidsai/dask-cuda/pull/834/files#r896411094 .

> As for `worker_port` and `listen_address` these are both supported by dask-workers but currently not fully support in dask-cuda-workers

And why do we need to specify them? Is this something we should add as arguments instead?

dask_cuda/cuda_worker.py
pentschev: Actually here we probably want `CUDAWorker`, rather than `Worker`.

dask_cuda/tests/test_dask_cuda_worker.py
pentschev: Why can't we go through all the lines? This seems like something that may change with time and the 5 lines won't anymore capture what we want to see (i.e., test will always pass even if there's

pentschev:
Since @wence- was able to resolve the scalability issues with https://github.com/rapidsai/dask-cuda/pull/933 and https://github.com/dask/distributed/pull/6580 , as discussed in https://github.com/rapidsai/dask-cuda/issues/930, should we close this or do you still see use for this PR @quasiben ?

PR #842, opened 2022-01-31
Add option to track RMM allocations

shwina:
Adds the `rmm_track_allocations` option that enables workers to query the amount of RMM memory allocated at any time via `mr.get_allocated_bytes()`.

This is used in https://github.com/dask/distributed/pull/5740.

### DIFF
--- a/dask_cuda/local_cuda_cluster.py
+++ b/dask_cuda/local_cuda_cluster.py
@@ -210,6 +210,7 @@ class LocalCUDACluster(LocalCluster):
         rmm_managed_memory=False,
         rmm_async=False,
         rmm_log_directory=None,
+        rmm_track_allocations=False,
         jit_unspill=None,
         log_spilling=False,
         worker_class=None,
@@ -272,6 +273,7 @@ class LocalCUDACluster(LocalCluster):
                 )
 
         self.rmm_log_directory = rmm_log_directory
+        self.rmm_track_allocations = rmm_track_allocations
 
         if not kwargs.pop("processes", True):
             raise ValueError(
@@ -415,6 +417,7 @@ class LocalCUDACluster(LocalCluster):
                         self.rmm_managed_memory,
                         self.rmm_async,
                         self.rmm_log_directory,
+                        self.rmm_track_allocations
                     ),
                 },
             }
--- a/dask_cuda/local_cuda_cluster.py
+++ b/dask_cuda/local_cuda_cluster.py
@@ -152,6 +152,16 @@ class LocalCUDACluster(LocalCluster):
         .. note::
             Logging will only be enabled if ``rmm_pool_size`` is specified or
             ``rmm_managed_memory=True``.
+    rmm_track_allocations : bool, default False
+        If True, wraps the memory resource used by each worker with a
+        ``rmm.mr.TrackingResourceAdaptor``, which tracks the amount of
+        memory allocated.
+
+        .. note::
+             This option enables additional diagnostics to be collected and
+             reported by the Dask dashboard. However, there is significant overhead
+             associated with this and it should only be used for debugging and
+             memory profiling.
     jit_unspill : bool or None, default None
         Enable just-in-time unspilling. Can be a boolean or ``None`` to fall back on
         the value of ``dask.jit-unspill`` in the local Dask configuration, disabling
--- a/dask_cuda/tests/test_dask_cuda_worker.py
+++ b/dask_cuda/tests/test_dask_cuda_worker.py
@@ -254,7 +254,7 @@ def test_cuda_visible_devices_uuid(loop):  # noqa: F811
                     result = client.run(lambda: os.environ["CUDA_VISIBLE_DEVICES"])
                     assert list(result.values())[0] == gpu_uuid
 
-                    
+
 def test_rmm_track_allocations(loop):  # noqa: F811
     rmm = pytest.importorskip("rmm")
     with popen(["dask-scheduler", "--port", "9369", "--no-dashboard"]):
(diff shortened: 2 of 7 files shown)


pentschev (commented):
That looks really cool, thanks @shwina . Could we add a test as well? We have some in https://github.com/rapidsai/dask-cuda/blob/b60dec90cf56dbc0937dfdce394200b08e3ebeba/dask_cuda/tests/test_local_cuda_cluster.py#L140-L202 that can be used as reference.

pentschev (commented):
@shwina it seems that there are some style issues, I suggest using `pre-commit` to automatically resolve them at commit time.

pentschev (changes requested):
@shwina thanks for addressing those. I have a couple of minor requests, shouldn't be too much work, but otherwise looks great!

pentschev:
Errors are unrelated, opened https://github.com/rapidsai/dask-cuda/issues/861 to track.

dask_cuda/tests/test_dask_cuda_worker.py
```diff
@@ -253,3 +253,34 @@ def test_cuda_visible_devices_uuid(loop):  # noqa: F811
...
+                    rmm.mr.get_current_device_resource_type
+                )
+                for v in memory_resource_type.values():
+                    assert v is rmm.mr.TrackingResourceAdaptor
+
+                memory_resource_upstream_type = client.run(
+                    lambda: type(rmm.mr.get_current_device_resource().upstream_mr)
+                )
+                for v in memory_resource_upstream_type.values():
+                    assert v is rmm.mr.PoolMemoryResource
```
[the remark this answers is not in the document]
shwina: Thanks! I added that.

dask_cuda/local_cuda_cluster.py
```diff
@@ -210,6 +210,7 @@ def __init__(
         rmm_managed_memory=False,
         rmm_async=False,
         rmm_log_directory=None,
+        rmm_track_allocations=False,
```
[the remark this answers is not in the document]
shwina: Done!

pentschev (approved):
LGTM. There's the failing test as reported in https://github.com/rapidsai/dask-cuda/issues/861 which blocks this PR, if this is not critical to go in immediately, let's wait for that to be fixed, otherwise we may need to xfail that test temporarily.

Thanks @shwina !

pentschev:
Actually there are some style issues too @shwina , could you fix them? I would strongly suggest using `pre-commit` to avoid this hassle for every new commit. 😄

dask_cuda/utils.py
```diff
@@ -56,10 +57,12 @@ def __init__(
         self.async_alloc = async_alloc
         self.logging = log_directory is not None
         self.log_directory = log_directory
+        self.rmm_track_allocations = track_allocations
 
     def setup(self, worker=None):
+        import rmm
```
charlesbluca, line 63: Are we safe to always import RMM here? If this function is always run as part of the `RMMSetup` plugin, wouldn't that make RMM an implicit dependency for starting a CUDA cluster?

pentschev: Great catch @charlesbluca , indeed I failed to realize that. Yes, we should remove it from the "main" context of `setup` and leave it solely within the local contexts that exist today, iff one of those options to enable RMM is set by the user.

dask_cuda/utils.py
```diff
@@ -56,10 +57,12 @@ def __init__(
...
+
         if self.async_alloc:
```
pentschev, line 65:
```suggestion
        if self.async_alloc:
```

dask_cuda/utils.py
```diff
@@ -83,6 +85,9 @@ def setup(self, worker=None):
                     worker, self.logging, self.log_directory
                 ),
             )
+        if self.rmm_track_allocations:
```
pentschev, line 88:
```suggestion
        if self.rmm_track_allocations:
            import rmm

```

pentschev (changes requested):

dask_cuda/utils.py
shwina: Thanks both! I've accepted @pentschev's suggestions here.

dask_cuda/utils.py
```diff
@@ -56,10 +57,10 @@ def __init__(
         self.async_alloc = async_alloc
         self.logging = log_directory is not None
         self.log_directory = log_directory
+        self.rmm_track_allocations = track_allocations
 
     def setup(self, worker=None):
         if self.async_alloc:
```
pentschev, line 63:
```suggestion
        if self.async_alloc:
            import rmm
```

dask_cuda/utils.py
```diff
@@ -69,7 +70,6 @@ def setup(self, worker=None):
                     )
                 )
         elif self.initial_pool_size is not None or self.managed_memory:
```
pentschev, line 72:
```suggestion
        elif self.initial_pool_size is not None or self.managed_memory:
            import rmm
```

pentschev (commented):
Forgot to add back the previously removed RMM imports.

pentschev (approved):

pentschev:
LGTM now, thanks @shwina for the work here, and @charlesbluca for the review and fixes here too!

pentschev:
@gpucibot merge

pentschev:
rerun tests

pentschev:
rerun tests

pentschev:
rerun tests

pentschev:
@gpucibot merge

PR #854, opened 2022-02-04
Add --pre-import/pre_import argument

pentschev:
Add argument to pre-import libraries as a Worker plugin to prevent long import times bleeding through later Dask operations.

### DIFF
--- a/dask_cuda/cli/dask_cuda_worker.py
+++ b/dask_cuda/cli/dask_cuda_worker.py
@@ -277,6 +277,13 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     help="""Use a different class than Distributed's default (``distributed.Worker``)
     to spawn ``distributed.Nanny``.""",
 )
+@click.option(
+    "--pre-import",
+    default=None,
+    help="""Pre-import libraries as a Worker plugin to prevent long import times
+    bleeding through later Dask operations. Should be a list of comma-separated names,
+    such as "cudf,rmm".""",
+)
 def main(
     scheduler,
     host,
@@ -309,6 +316,7 @@ def main(
     net_devices,
     enable_jit_unspill,
     worker_class,
+    pre_import,
     **kwargs,
 ):
     if tls_ca_file and tls_cert and tls_key:
@@ -358,6 +366,7 @@ def main(
         net_devices,
         enable_jit_unspill,
         worker_class,
+        pre_import,
         **kwargs,
     )
 
--- a/dask_cuda/cuda_worker.py
+++ b/dask_cuda/cuda_worker.py
@@ -24,6 +24,7 @@ from .initialize import initialize
 from .proxify_host_file import ProxifyHostFile
 from .utils import (
     CPUAffinity,
+    PreImport,
     RMMSetup,
     _ucx_111,
     cuda_visible_devices,
@@ -80,6 +81,7 @@ class CUDAWorker(Server):
         net_devices=None,
         jit_unspill=None,
         worker_class=None,
+        pre_import=None,
         **kwargs,
     ):
         # Required by RAPIDS libraries (e.g., cuDF) to ensure no context
@@ -249,6 +251,7 @@ class CUDAWorker(Server):
                         rmm_async,
                         rmm_log_directory,
                     ),
+                    PreImport(pre_import),
                 },
                 name=name if nprocs == 1 or name is None else str(name) + "-" + str(i),
                 local_directory=local_directory,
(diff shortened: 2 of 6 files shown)


dask_cuda/cli/dask_cuda_worker.py
```diff
@@ -277,6 +277,13 @@
     help="""Use a different class than Distributed's default (``distributed.Worker``)
     to spawn ``distributed.Nanny``.""",
 )
+@click.option(
+    "--pre-import",
+    default=None,
```
randerzander, line 282: Should we not include `cudf` and `rmm` by default?

If not, because we don't want to assume all dask-cuda users are cudf users, this would be a point in favor of thinking more about the concept of dask-cuda "profiles" which include the defaults most users will want without knowing they want them.

quasiben: Building a [RAPIDS worker plugin](http://distributed.dask.org/en/stable/plugins.html#worker-plugins) would make sense.  But yes, for now dask-cuda is still general purpose and we shouldn't include it by default.

dask_cuda/cli/dask_cuda_worker.py
```diff
@@ -277,6 +277,13 @@
...
+    help="""Pre-import libraries as a Worker plugin to prevent long import times
+    bleeding through later Dask operations. Should be a list of comma-separated names,
+    such as "cudf,rmm".""",
```
shwina, line 285: Apologies for the drive-by comment, but wouldn't a `List[str]` be more appropriate here?

dask_cuda/cli/dask_cuda_worker.py
jakirkham: Yeah we do have people come by periodically who are simply using Dask-CUDA alone. They might have issues if we added other hard requirements like these

That said, maybe we could include an example in the docs?

dask_cuda/cli/dask_cuda_worker.py
jakirkham: There is [`nargs` support]( https://click.palletsprojects.com/en/8.0.x/options/#multi-value-options ) in Click. Is that what you are referring to?

jakirkham:
Not sure how doable this would be, but would it make sense to add a test?

dask_cuda/cli/dask_cuda_worker.py
shwina: Ah hmm that wouldn't work either as it's not a fixed number of args. I'm just going to close this suggestion as I don't think there's a better way to do it, honestly.

jakirkham: Is it? I think if [`nargs=-1`]( https://click.palletsprojects.com/en/8.0.x/arguments/#variadic-arguments ), we'd get the intended behavior

pentschev: `nargs=-1` seems to only apply to variadic arguments though (i.e., "unnamed" arguments), but [not support for options](https://github.com/pallets/click/blob/main/src/click/core.py#L2492) as is the case for `--pre-import`.

jakirkham: Makes sense. Then let's leave as-is. There doesn't seem to be a good alternative

dask_cuda/cli/dask_cuda_worker.py
pentschev: This is the example that is in the argument description itself, see https://github.com/rapidsai/dask-cuda/pull/854/files#diff-b3d47096fc004d0fe86d1630b46845c70e79a9914b065ba2b52c4fc073dd1621R283-R285 and https://github.com/rapidsai/dask-cuda/pull/854/files#diff-245fed94d1f88f78dbd6898226b22c3cb2084eb9e8e7c8994beee67c44461f6aR169-R171 . I'm not sure I see what more is being suggested we do here.

pentschev:
> Not sure how doable this would be, but would it make sense to add a test?

Quite time consuming to get smooth tests for this, but now done in https://github.com/rapidsai/dask-cuda/pull/854/commits/b21f791efa1ba25449c54ed3f751c947e2b47a7a .

jakirkham:
Ouch! Sorry to hear that Peter 😞

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks Peter! 😄

dask_cuda/cli/dask_cuda_worker.py
jakirkham: Yeah agree docs are sufficient already then

Maybe @randerzander could share a bit more of what he is looking for with Dask-CUDA profiles in a new issue and we could discuss that? 😉

Issue #855, opened 2022-02-10
[BUG] Spilling  on demand with CuPy arrays not working as expected


VibhuJawa:
We seem to be running into issues when trying to spill CuPY arrays with dask-cuda cluster . On demand spilling fails to spill even though it should be able to spill some of these cupy arrays, allowing for successful execution. 

This seems like a probable cause of errors and hangs we see in GPU-BDB ML queries (05,20,25,27) as they use cuML and SpaCY under the hood  which in turn use CuPY.  Currently to work around these we set device_memory_limit really low (15GB) on 32GB cards to succeed which kills a lot of performance due to lack of available memory.   More details on the impact of this on gpu-bdb like workflows in the end of this issue. 

**Example:**

```python
from dask import persist
from distributed import Client,wait
from dask_cuda import LocalCUDACluster

import cupy as cp
import dask.array as da

cluster = LocalCUDACluster(CUDA_VISIBLE_DEVICES='0,1',
                           rmm_pool_size="25GB",
                          rmm_maximum_pool_size="29GB",
                           device_memory_limit="10GB",
                           log_spilling=False,
                           jit_unspill=True)  

client = Client(cluster)

def use_allocator():
    import rmm
    import cupy
    cupy.cuda.set_allocator(rmm.rmm_cupy_allocator)
_ = client.run(use_allocator);
del _;

### Main Reproducerr
rs =  da.random.RandomState(RandomState=cp.random.RandomState)
x = rs.random((5_000_000, 1000), chunks=(100, 1000))


u, s, v = da.linalg.svd(x)
u, s, v = persist(u, s, v)
%time _ = wait([u, s, v]) 
```

Main Trace: 

```python
distributed.worker - WARNING - RMM allocation of 4.91 MiB failed, spill-on-demand couldn't find any device memory to spill:
<ProxyManager dev_limit=9.31 GiB host_limit=503.90 GiB disk=0 B(0) host=0 B(0) dev=0 B(0)>: Empty
traceback:
  File "/datasets/vjawa/miniconda3/envs/rapids-22.04/lib/python3.9/threading.py", line 930, in _bootstrap
    self._bootstrap_inner()
  File "/datasets/vjawa/miniconda3/envs/rapids-22.04/lib/python3.9/threading.py", line 973, in _bootstrap_inner
    self.run()
  File "/datasets/vjawa/miniconda3/envs/rapids-22.04/lib/
```

<details>

```python

[...]

PR #856, opened 2022-02-11
Introduce incompatible-types and enables spilling of CuPy arrays

madsbk:
This PR replaces `ignore_types` introduced in https://github.com/rapidsai/dask-cuda/pull/568 with `incompatible_types`, which is a list of types that `ProxifyHostFile` will unproxify on retrieval. This makes it possible to spill types we previously ignored completely such as  `cupy.ndarray`.

To mark a type incompatible, add it to the comma separated config value `"jit-unspill-incompatible"` or environment variable `DASK_JIT_UNSPILL_INCOMPATIBLE`.

The default value is: `DASK_JIT_UNSPILL_INCOMPATIBLE="cupy.ndarray"`

Closes https://github.com/rapidsai/dask-cuda/issues/855

Notice, I have marked the PR `breaking` because the `DASK_JIT_UNSPILL_IGNORE` option has been removed.

cc. @VibhuJawa

### DIFF
--- a/dask_cuda/proxify_device_objects.py
+++ b/dask_cuda/proxify_device_objects.py
@@ -2,7 +2,7 @@ import functools
 import pydoc
 from collections import defaultdict
 from functools import partial
-from typing import List, MutableMapping, TypeVar
+from typing import List, MutableMapping, Optional, Tuple, TypeVar
 
 import dask
 from dask.utils import Dispatch
@@ -10,33 +10,30 @@ from dask.utils import Dispatch
 from .proxy_object import ProxyObject, asproxy
 
 dispatch = Dispatch(name="proxify_device_objects")
-ignore_types = None
+incompatible_types: Optional[Tuple[type]] = None
 
 T = TypeVar("T")
 
 
-def _register_ignore_types():
-    """Lazy register types that shouldn't be proxified
+def _register_incompatible_types():
+    """Lazy register types that ProxifyHostFile should unproxify on retrieval.
 
-    It reads the config key "jit-unspill-ignore" (DASK_JIT_UNSPILL_IGNORE),
-    which should be a comma seperated list of types to ignore. The default
-    value is:
-        DASK_JIT_UNSPILL_IGNORE="cupy.ndarray"
-
-    Notice, it is not possible to ignore types explicitly handled by this
-    module such as `cudf.DataFrame`, `cudf.Series`, and `cudf.Index`.
+    It reads the config key "jit-unspill-incompatible"
+    (DASK_JIT_UNSPILL_INCOMPATIBLE), which should be a comma seperated
+    list of types. The default value is:
+        DASK_JIT_UNSPILL_INCOMPATIBLE="cupy.ndarray"
     """
-    global ignore_types
-    if ignore_types is not None:
+    global incompatible_types
+    if incompatible_types is not None:
         return  # Only register once
     else:
-        ignore_types = ()
+        incompatible_types = ()
 
-    ignores = dask.config.get("jit-unspill-ignore", "cupy.ndarray")
-    ignores = ignores.split(",")
+    incompatibles = dask.config.get("jit-unspill-incompatible", "cupy.ndarray")
+    incompatibles = incompatibles.split(",")
 
     toplevels = defaultdict(set)
-    for path in ignores:
+    for path in incompatibles:
         if path:
             toplevel = path.split(".", maxsplit=1)[0].strip()
             toplevels[toplevel].add(path.strip())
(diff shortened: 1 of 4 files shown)


dask_cuda/proxify_device_objects.py
```diff
@@ -2,50 +2,49 @@
...
+    if incompatible_types is not None:
         return  # Only register once
     else:
-        ignore_types = ()
+        incompatible_types = ()
 
-    ignores = dask.config.get("jit-unspill-ignore", "cupy.ndarray")
-    ignores = ignores.split(",")
+    incompatibles = dask.config.get("jit-unspill-incompatible", "cupy.ndarray")
+    incompatibles = incompatibles.split(",")
```
VibhuJawa, line 33: Thanks for working on this @madsbk . 

I am curious, What about stuff like [cupyx.scipy.sparse](https://docs.cupy.dev/en/v9.0.0/reference/sparse.html#module-cupyx.scipy.sparse) and [CumlArray](https://github.com/rapidsai/cuml/blob/818bd2503a840f6f85190a8d3dd719a36504bdf8/python/cuml/common/array.py#L36-L46)  . 

Will this work with those too ?

dask_cuda/proxify_host_file.py
```diff
@@ -496,6 +503,7 @@ def __init__(
         gds_spilling: bool = None,
     ):
         self.store: Dict[Hashable, Any] = {}
+        self.key_containts_incompatible_type: Dict[Hashable, bool] = {}
```
shwina, line 506:
```suggestion
        self.key_contains_incompatible_type: Dict[Hashable, bool] = {}
```

ditto below?

dask_cuda/proxify_host_file.py
```diff
@@ -604,19 +612,26 @@ def __setitem__(self, key, value):
...
+            # Notice, we only call `unproxify_device_objects()` when `key`
+            # contains incompatible types.
+            ret = unproxify_device_objects(ret, only_incompatible_types=True)
+            self.manager.maybe_evict()
         return ret
 
     def __delitem__(self, key):
         with self.lock:
             del self.store[key]
+            del self.key_contains_incompatible_type[key]
```
pentschev, line 634: Wouldn't it be simpler to keep a single `dict` storing a `(obj, incompatible_type_found)`? Maybe I'm missing some context, but a single `dict` would potentially make it less error prone on forgetting to update one of them.

dask_cuda/tests/test_proxify_host_file.py
```diff
@@ -324,29 +326,23 @@ def test_externals():
     assert dhf.manager._dev._mem_usage == 0
 
 
-def test_proxify_device_objects_of_cupy_array():
-    """Check that a proxied array behaves as a regular cupy array
+@patch("dask_cuda.proxify_device_objects.incompatible_types", (cupy.ndarray,))
+def test_incompatible_types():
+    """Check that ProxifyHostFile unproxify `cupy.ndarray` on retrieval
```
pentschev, line 331:
```suggestion
    """Check that ProxifyHostFile unproxifies `cupy.ndarray` on retrieval
```

pentschev (approved):
Thanks @madsbk , I've added a few minor suggestions, but otherwise LGTM.

dask_cuda/proxify_host_file.py
shwina: I added a comment explaining exactly what `self.store` is. Please let me know if you think the comment is accurate.

pentschev: Thanks @shwina , I think the changes to combine `dict`s look good, but I didn't see any comment, any chance you missed pushing the change?

shwina: Done?

pentschev: I'm confused, I still don't see any comments. 😅 

Could you paste a link to that diff where the comment is contained?

pentschev:
rerun tests

dask_cuda/proxify_host_file.py
shwina: I kept pushing to my fork instead of Mads' :( Sorry for the confusion. Should be updated on this PR now.

pentschev: No worries Ashwin, I see your comment now and I think it's explaining correctly, thanks for adding that.

pentschev (approved):
LGTM, thanks @madsbk and @shwina for the work in this!

pentschev:
@gpucibot merge

madsbk:
Thanks @shwina for finishing the PR and thanks @pentschev for the review :)

PR #868, opened 2022-03-07
Use conda compilers

Ethyling:
This PR enables the usage of conda compilers to build conda packages. This is required to use `mambabuild`

### DIFF
--- a/conda/recipes/dask-cuda/conda_build_config.yaml
+++ b/conda/recipes/dask-cuda/conda_build_config.yaml
@@ -1,0 +1,8 @@
+c_compiler_version:
+  - 9
+
+cxx_compiler_version:
+  - 9
+
+sysroot_version:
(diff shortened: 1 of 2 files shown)


jakirkham:
Sorry for missing this PR previously, with Dask-CUDA the code is all pure Python. So there isn't any compilation here. Given this do we still need this change? Or are there other changes in the background that are necessitating this for some reason?

Ethyling:
> Sorry for missing this PR previously, with Dask-CUDA the code is all pure Python. So there isn't any compilation here. Given this do we still need this change? Or are there other changes in the background that are necessitating this for some reason?

Ah, my bad, I totally miss this. No, I think we can just close this

jakirkham:
Oh ok, thanks Jordan. Sorry for not catching this sooner 😞

PR #883, opened 2022-03-30
Resolve build issues / consistency with conda-forge packages

charlesbluca:
This PR:

- bumps dask-cuda's minimum python version to 3.8 to coincide with other RAPIDS projects
- bumps the pynvml minimum version to 11.0.0 to match up with the [conda-forge package](https://github.com/conda-forge/dask-cuda-feedstock/blob/main/recipe/meta.yaml)
- adds a max version constraint for `click` to resolve the build issues arising from click 8.1.0

### DIFF
--- a/conda/recipes/dask-cuda/meta.yaml
+++ b/conda/recipes/dask-cuda/meta.yaml
@@ -32,6 +32,7 @@ requirements:
     - pynvml>=11.0.0
     - numpy>=1.16.0
     - numba>=0.53.1
+    - click<8.1
 
 test:
   imports:
--- a/conda/recipes/dask-cuda/meta.yaml
+++ b/conda/recipes/dask-cuda/meta.yaml
@@ -29,7 +29,7 @@ requirements:
     - python
     - dask==2022.03.0
     - distributed==2022.03.0
-    - pynvml>=8.0.3
+    - pynvml>=11.0.0
     - numpy>=1.16.0
     - numba>=0.53.1
 
--- a/conda/recipes/dask-cuda/meta.yaml
+++ b/conda/recipes/dask-cuda/meta.yaml
@@ -32,7 +32,7 @@ requirements:
     - pynvml>=11.0.0
     - numpy>=1.16.0
     - numba>=0.53.1
-    - click<8.1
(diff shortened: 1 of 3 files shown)


conda/recipes/dask-cuda/meta.yaml
```diff
@@ -29,9 +29,10 @@ requirements:
     - python
     - dask==2022.03.0
     - distributed==2022.03.0
-    - pynvml>=8.0.3
+    - pynvml>=11.0.0
     - numpy>=1.16.0
     - numba>=0.53.1
+    - click<8.1
```
jakirkham, line 35: Should we add this to [`requirements.txt`]( https://github.com/rapidsai/dask-cuda/blob/branch-22.04/requirements.txt )?

charlesbluca: Good catch - will do that now

charlesbluca, line 35: Pinning click to 8.0.4 as suggested:

```suggestion
    - click==8.0.4
```

requirements.txt
```diff
@@ -3,3 +3,4 @@ distributed==2022.03.0
 pynvml>=11.0.0
 numpy>=1.16.0
 numba>=0.53.1
+click<8.1
```
charlesbluca, line 6:
```suggestion
click==8.0.4
```

conda/recipes/dask-cuda/meta.yaml
```diff
@@ -32,7 +32,7 @@ requirements:
     - pynvml>=11.0.0
     - numpy>=1.16.0
     - numba>=0.53.1
-    - click<8.1
+    - click==8.0.4
```
jakirkham, line 35: We discussed offline and decided to match what was done in PR ( https://github.com/rapidsai/integration/pull/453 ). The `distributed` issue this stems from is a bit different than the one that `black` encountered, but this pinning should work just as well here

pentschev (approved):
LGTM, thanks @charlesbluca !

pentschev:
@gpucibot merge

PR #885, opened 2022-03-31
Raise `ImportError` when platform is not Linux

pentschev:
To prevent users inadvertently trying to use Dask-CUDA (and potentially other RAPIDS components) on a non-supported platform such as Windows, raise an `ImportError`.

### DIFF
--- a/dask_cuda/__init__.py
+++ b/dask_cuda/__init__.py
@@ -1,3 +1,9 @@
+import sys
+
+if sys.platform != "linux":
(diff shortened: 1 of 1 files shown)


galipremsagar (approved):

pentschev:
Thanks @galipremsagar , let's just wait for @charlesbluca's confirmation that it doesn't break WSL2 before merging.

charlesbluca (approved):
Still able to import successfully on WSL2 🙂

pentschev:
Thanks @charlesbluca for confirming!

PR #886, opened 2022-03-31
Remove legacy `check_python_3`

pentschev:
The legacy `check_python_3` function was recently removed from Distributed, which breaks with `main` builds and with the future Dask/Distributed releases, but it's anyway not needed anymore since we only support Python 3.8 and above.


galipremsagar (approved):

pentschev:
@gpucibot merge

jakirkham:
Thanks Peter! 😄

Yeah upstream dropped this as the recent `click` 8.1.0 release broke this function. So they removed it.

PR #893, opened 2022-04-13
Pull `requirements.txt` into Conda recipe

jakirkham:
Leverages [`load_setup_py_data`]( https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html?highlight=load_setup_py_data#templating-with-jinja ) to extract the requirements from `setup.py` and add them to the Conda recipe.

### DIFF
--- a/ci/cpu/build.sh
+++ b/ci/cpu/build.sh
@@ -68,7 +68,7 @@ pip install git+https://github.com/dask/distributed.git@main
 # BUILD - Package builds
 ################################################################################
 
-gpuci_logger "Build conda pkg for libcudf"
+gpuci_logger "Build conda pkg for dask-cuda"
 gpuci_conda_retry build conda/recipes/dask-cuda --python=${PYTHON}
 
 rm -rf dist/
--- a/conda/recipes/dask-cuda/meta.yaml
+++ b/conda/recipes/dask-cuda/meta.yaml
@@ -2,6 +2,8 @@
 
 # Usage:
 #   conda build -c conda-forge .
+{% set data = load_setup_py_data() %}
+
 {% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev').lstrip('v') + environ.get('VERSION_SUFFIX', '') %}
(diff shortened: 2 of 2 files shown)


ci/cpu/build.sh
```diff
@@ -68,7 +68,7 @@ pip install git+https://github.com/dask/distributed.git@main
 # BUILD - Package builds
 ################################################################################
 
-gpuci_logger "Build conda pkg for libcudf"
+gpuci_logger "Build conda pkg for dask-cuda"
 gpuci_conda_retry build conda/recipes/dask-cuda --python=${PYTHON}
```
jakirkham, line 72: It's worth noting that it is important for us to start in the directory relative to `setup.py` as is already being done. Using a different directory doesn't work.

There isn't a good way to start from other directories atm. Please see issue ( https://github.com/conda/conda-build/issues/3754 ) for context.

jakirkham:
The other thing worth noting is there is a Setuptools deprecation warning that comes up ( https://github.com/conda/conda-build/issues/4428 )

pentschev:
Is this ready for review or is there more to be addressed @jakirkham ?

jakirkham:
Technically it works.

That said, think it is worthwhile to read up on the linked issues (particular the last one).

charlesbluca (approved):

charlesbluca:
@jakirkham is your preference for https://github.com/conda/conda-build/pull/4430 to get merged in (and incorporated into gpuCI containers) before merging this?

jakirkham:
Sorry for the slow reply. So we can do this. The issue with Setuptools is in part they are changing the API (easy to fix in Conda-Build), but they are also flagging the new API as provisional. So there may be issues we run into in the future. I'm ok with trying this out. We can always go back to what we were doing before if there are issues with Setuptools, but wanted to make sure others here are aware of that and ok with that.

ajschmidt8 (approved):

ajschmidt8 (approved):

pentschev:
@gpucibot merge

PR #896, opened 2022-04-21
Update auto host_memory computation when threads per worker > 1

### LINKED ISSUE #874: [Bug] host memory limit incorrectly calculated for threads_per_worker > 1
When a `LocalCUDACluster` is created with `threads_per_worker > 1` the memory_limit per worker is incorrectly computed to be `threads_per_worker * expected_value`.

If I have 8 workers and 800GB host memory, I get 100GB/worker if tpw=1, but 200GB/worker if tpw=2

ayushdg:
Fixes #874 .

Updates the arguments to `parse_memory_limit` to set the right memory_limit values for cases where threads_per_worker > 1 as discussed in https://github.com/rapidsai/dask-cuda/issues/874#issuecomment-1101770211

### DIFF
--- a/dask_cuda/cuda_worker.py
+++ b/dask_cuda/cuda_worker.py
@@ -85,7 +85,7 @@ class CUDAWorker(Server):
         if nthreads < 1:
             raise ValueError("nthreads must be higher than 0.")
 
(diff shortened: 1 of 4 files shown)


pentschev (approved):
LGTM, thanks @ayushdg for spotting and fixing this!

pentschev:
@gpucibot merge

PR #901, opened 2022-04-26
Increase spill tests timeout to 30 seconds

pentschev:
Timeouts have occurred already in a few gpuCI instances, and locally those are consistently reproducible when the spill test is the first to run, where it takes just over 20 seconds probably due to added overhead of CUDA context creation. Increasing the timeout to 30 seconds seems sufficient.

### DIFF
--- a/dask_cuda/tests/test_spill.py
+++ b/dask_cuda/tests/test_spill.py
@@ -118,7 +118,7 @@ def delayed_worker_assert(total_size, device_chunk_overhead, serialized_chunk_ov
         },
     ],
 )
-@gen_test(timeout=20)
+@gen_test(timeout=30)
 async def test_cupy_cluster_device_spill(params):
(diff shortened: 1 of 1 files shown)


Ethyling (approved):

madsbk (approved):
LGTM, but maybe increase to `60 sec` just to be on the safe side?

pentschev:
> LGTM, but maybe increase to `60 sec` just to be on the safe side?

I think 60 seconds may be too much, if we really need to increase that high, maybe we should consider redesigning tests so that they run faster. TBH, 30 seconds already feels too high for me, but spilling is hard to keep that much under.

pentschev:
Thanks for reviews/approvals @Ethyling @madsbk .

pentschev:
@gpucibot merge

PR #903, opened 2022-04-27
Disable `include-ignored` coverage warnings

pentschev:
Disable warnings that are printed hundreds of times in CI:

```
/opt/conda/envs/rapids/lib/python3.8/site-packages/coverage/inorout.py:472:CoverageWarning: --include is ignored because --source is set (include-ignored)
  self.warn("--include is ignored because --source is set", slug="include-ignored")
```

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,5 @@
 # Configuration file for Python coverage tests
 [run]
+disable_warnings = include-ignored
 include = dask_cuda/*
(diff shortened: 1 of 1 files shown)


raydouglass (approved):

pentschev:
Thanks @raydouglass !

pentschev:
@gpucibot merge

PR #897, opened 2022-04-21
Use conda to build python packages during GPU tests

Ethyling:
This PR convert the `from sources` build we are doing in GPU test job to a `conda build`. This is done for the following reasons:
- This is required step to improve the Ops CI/CD setup to a more convenient pipeline
- This is required to start using `conda` compilers and `mamba` to build RAPIDS packages
- This prevent us from manually managing and installing the dependencies in GPU job
- This ensure the packages can be installed
- This ensure the tests are running and working against the package content and not the build results. Currently the Python packages are not tested.

This may increase the global pipeline time, but the usage of `mamba` should resolve this as `mamba` is faster than `conda` to build packages

PR #916, opened 2022-05-25
[REVIEW] Pin `dask` & `distributed` for release

galipremsagar:
Pinnings to be in-line with: https://github.com/rapidsai/cudf/pull/10965

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -71,7 +71,11 @@ if [[ "${INSTALL_DASK_MAIN}" == 1 ]]; then
   gpuci_mamba_retry install -c dask/label/dev \
     "dask/label/dev::dask" \
     "dask/label/dev::distributed"
+else
+  gpuci_logger "gpuci_mamba_retry install conda-forge::dask==2022.05.1 conda-forge::distributed==2022.05.1 conda-forge::dask-core==2022.05.1 --force-reinstall"
+  gpuci_mamba_retry install conda-forge::dask==2022.05.1 conda-forge::distributed==2022.05.1 conda-forge::dask-core==2022.05.1 --force-reinstall
 fi
+
 
 gpuci_logger "Check versions"
 python --version
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -37,6 +37,9 @@ export NUMPY_EXPERIMENTAL_ARRAY_FUNCTION=1
 # development time and disabled before a new dask-cuda release.
 export INSTALL_DASK_MAIN=0
 
+# Dask version to install when `INSTALL_DASK_MAIN=0`
+export DASK_STABLE_VERSION="2022.05.1"
+
 ################################################################################
 # SETUP - Check environment
 ################################################################################
@@ -72,8 +75,8 @@ if [[ "${INSTALL_DASK_MAIN}" == 1 ]]; then
     "dask/label/dev::dask" \
     "dask/label/dev::distributed"
 else
-  gpuci_logger "gpuci_mamba_retry install conda-forge::dask==2022.05.1 conda-forge::distributed==2022.05.1 conda-forge::dask-core==2022.05.1 --force-reinstall"
-  gpuci_mamba_retry install conda-forge::dask==2022.05.1 conda-forge::distributed==2022.05.1 conda-forge::dask-core==2022.05.1 --force-reinstall
+  gpuci_logger "gpuci_mamba_retry install conda-forge::dask==${DASK_STABLE_VERSION} conda-forge::distributed==${DASK_STABLE_VERSION} conda-forge::dask-core==${DASK_STABLE_VERSION} --force-reinstall"
+  gpuci_mamba_retry install conda-forge::dask==${DASK_STABLE_VERSION} conda-forge::distributed==${DASK_STABLE_VERSION} conda-forge::dask-core==${DASK_STABLE_VERSION} --force-reinstall
 fi
 
 
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -35,7 +35,7 @@ export NUMPY_EXPERIMENTAL_ARRAY_FUNCTION=1
 
 # Install dask and distributed from main branch. Usually needed during
 # development time and disabled before a new dask-cuda release.
-export INSTALL_DASK_MAIN=1
+export INSTALL_DASK_MAIN=0
 
 ################################################################################
 # SETUP - Check environment
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
-dask>=2022.03.0
-distributed>=2022.03.0
+dask==2022.05.1
+distributed==2022.05.1
 pynvml>=11.0.0
 numpy>=1.16.0
 numba>=0.54


pentschev:
The failures seem to arise from inconsistent versions being picked from conda packages:

```
15:49:34 dask_cuda/tests/test_dask_cuda_worker.py::test_cuda_visible_devices_and_memory_limit_and_nthreads 
15:49:34 --------------------------- Subprocess stdout/stderr---------------------------
15:49:34 Traceback (most recent call last):
15:49:34   File "/opt/conda/envs/rapids/lib/python3.8/site-packages/pkg_resources/__init__.py", line 579, in _build_master
15:49:34     ws.require(__requires__)
15:49:34   File "/opt/conda/envs/rapids/lib/python3.8/site-packages/pkg_resources/__init__.py", line 897, in require
15:49:34     needed = self.resolve(parse_requirements(requirements))
15:49:34   File "/opt/conda/envs/rapids/lib/python3.8/site-packages/pkg_resources/__init__.py", line 788, in resolve
15:49:34     raise VersionConflict(dist, req).with_context(dependent_req)
15:49:34 pkg_resources.ContextualVersionConflict: (dask 2022.5.0+46.g4c0cdfdbd (/opt/conda/envs/rapids/lib/python3.8/site-packages), Requirement.parse('dask==2022.05.1'), {'distributed', 'dask-cuda'})
```

@Ethyling @ajschmidt8 @charlesbluca @jakirkham would be good to get your expertise here in what would be a safe way to prevent gpuCI from picking an inconsistent set of `dask`/`dask-core`/`distributed`/etc packages.

pentschev:
To be more specific, I think those issues are related to the changes from https://github.com/rapidsai/dask-cuda/pull/897 .

charlesbluca:
I think the issue here is that when `INSTALL_DASK_MAIN=0`, we need to explicitly install stable Dask packages in the GPU build script; I think this should be resolved by replacing this block:

https://github.com/rapidsai/dask-cuda/blob/63529e891b52aee6c1bfeeecd0c8ff272628d4d0/ci/gpu/build.sh#L69-L74

With something like this (used in cuDF's CI):

https://github.com/rapidsai/cudf/blob/df5dc08710c9cda8aa5f9b2f0b26a863301b5e01/ci/gpu/build.sh#L92-L99

ci/gpu/build.sh
```diff
@@ -35,7 +35,7 @@ export NUMPY_EXPERIMENTAL_ARRAY_FUNCTION=1
 
 # Install dask and distributed from main branch. Usually needed during
 # development time and disabled before a new dask-cuda release.
-export INSTALL_DASK_MAIN=1
+export INSTALL_DASK_MAIN=0
```
pentschev, line 38:
```suggestion
export INSTALL_DASK_MAIN=0

# Dask version to install when `INSTALL_DASK_MAIN=0`
export DASK_STABLE_VERSION="2022.05.1"
```

ci/gpu/build.sh
```diff
@@ -71,8 +71,12 @@ if [[ "${INSTALL_DASK_MAIN}" == 1 ]]; then
   gpuci_mamba_retry install -c dask/label/dev \
     "dask/label/dev::dask" \
     "dask/label/dev::distributed"
+else
+  gpuci_logger "gpuci_mamba_retry install conda-forge::dask==2022.05.1 conda-forge::distributed==2022.05.1 conda-forge::dask-core==2022.05.1 --force-reinstall"
+  gpuci_mamba_retry install conda-forge::dask==2022.05.1 conda-forge::distributed==2022.05.1 conda-forge::dask-core==2022.05.1 --force-reinstall
```
pentschev, line 76:
```suggestion
  gpuci_logger "gpuci_mamba_retry install conda-forge::dask==${DASK_STABLE_VERSION} conda-forge::distributed==${DASK_STABLE_VERSION} conda-forge::dask-core==${DASK_STABLE_VERSION} --force-reinstall"
  gpuci_mamba_retry install conda-forge::dask==${DASK_STABLE_VERSION} conda-forge::distributed==${DASK_STABLE_VERSION} conda-forge::dask-core==${DASK_STABLE_VERSION} --force-reinstall
```

pentschev:
> We could try waiting for new images & packages to be published after [rapidsai/integration#476](https://github.com/rapidsai/integration/pull/476) is merged. If that works we wouldn't need else condition.

Yes, that would be good. But perhaps we can get this merged right away and then change it afterwards. Also added a few suggestions.

pentschev:
> removing the Dask channel from our conda config, we weren't guaranteed to have stable packages, and would want a step to explicitly ensure they are installed?

How so? If that's the case, it then sounds to me that it's not guaranteed that pinning works for the release process which would be a major issue IMO.

My hope was always that the integration repo versions _MUST_ be respected, if not that feels like an enormous flaw to me.

@Ethyling @ajschmidt8 could you comment on this?

charlesbluca:
Think this is more of a consequence of the fact that the Dask channel is placed at a higher priority than conda-forge during gpuCI image builds?

https://github.com/rapidsai/gpuci-build-environment/blob/9c199e81bdd6bcf157be79976be10a06afd8d305/context/.condarc

In that case, installing `dask=2022.5.1` would still end up preferring the nightly packages over the stable release (at least this seems to be my behavior trying this out locally).

We could potentially remove the `else` blocks throughout the build scripts by centralizing the solution to the gpuCI image builds, either by:

- explicitly or conditionally removing the `dask/label/dev` channel around code freeze
- potentially changing the channel priority policy, as I think that in general `2022.5.1` > `2022.5.1aXXXXXX` and if conda was allowed to prefer lower priority channels it would probably end up going for the stable package over nightlies

galipremsagar:
> explicitly or conditionally removing the dask/label/dev channel around code freeze
We do this for release builds I suppose.

The `force-reinstall` in else block(coming from cudf) is because we have an image/package that already has pacakges from `dask/label/dev` channel and those get higher priority for the same version `2022.5.1`, hence we do a `force-reinstall` in else. - This is what I was able to summarise from all the experimentation I did in https://github.com/rapidsai/cudf/pull/10182.

pentschev:
> The `force-reinstall` in else block(coming from cudf) is because we have an image/package that already has pacakges from `dask/label/dev` channel and those get higher priority for the same version `2022.5.1`, hence we do a `force-reinstall` in else to get `conda-forge` packages installed. - This is what I was able to summarise from all the experimentation I did in [rapidsai/cudf#10182](https://github.com/rapidsai/cudf/pull/10182).

Maybe I'm misunderstanding something, but this still seems problematic with the release process. When we release, won't a non-stable Dask version be picked then? Additionally, if anyone has `dask/label/dev` channel when installing RAPIDS, they can pick a `2022.05.1XXXX` version that will be breaking for them.

raydouglass (approved):

jakirkham:
Maybe we should follow up on this discussion after the release?

jakirkham:
rerun tests

galipremsagar:
> When we release, won't a non-stable Dask version be picked then?

Nope, at least in cudf & docker image creations for release build we drop those channels: 
1. https://github.com/rapidsai/cudf/blob/4e66281f48c55735edb4b610e0f859ee2de32a75/ci/cpu/build.sh#L51
2. https://github.com/rapidsai/docker/blob/fea6a2fbe3c858ead7c0174dab231cec1be30cf1/generated-dockerfiles/rapidsai-core_ubuntu20.04-base.amd64.Dockerfile#L22

> Additionally, if anyone has dask/label/dev channel when installing RAPIDS, they can pick a 2022.05.1XXXX version that will be breaking for them.

Yea, that is possible.

pentschev:
> Nope, at least in cudf & docker image creations for release build we drop those channels:

Got it. It seems a bit different in Dask-CUDA, where Dask and Distributed `main` are installed from `pip`:

https://github.com/rapidsai/dask-cuda/blob/63529e891b52aee6c1bfeeecd0c8ff272628d4d0/ci/cpu/build.sh#L64-L65

I don't mean to hold off on this PR, but I want to make sure that we're not making any mistakes for the release process and that nightly packages after this PR is merged -- and ultimately the release packages -- indeed pin to the correct Dask version. If there's a mistake with that, it can be critical in the future.

If there are reasons to believe that I'm overcomplicating things, please let me know, but I would like to confirm who/what/where is the actual pinning coming from/happening.

requirements.txt
```diff
@@ -1,5 +1,5 @@
-dask>=2022.03.0
-distributed>=2022.03.0
+dask==2022.05.1
+distributed==2022.05.1
```
jakirkham, line 2: These will ensure `dask` & `distributed` are pinned correctly in the Conda packages. In PR ( https://github.com/rapidsai/dask-cuda/pull/893 ), we ensure these are pulled into the Conda packages

pentschev:
Thanks @galipremsagar for working on this, and everyone for clarifications. It seems like release process is correct, I will keep an eye on this just to be safe, but I think we're good here.

pentschev:
@gpucibot merge

PR #918, opened 2022-05-26
Reduce `test_cudf_cluster_device_spill` test and speed it up

pentschev:
This test has been non-deterministically timing out in gpuCI, instead of increasing its timeout, reduce problem size to make it faster.


galipremsagar (approved):
Thanks @pentschev !

galipremsagar:
@gpucibot merge

PR #933, opened 2022-06-15
Add --multiprocessing-method option to benchmarks

### LINKED ISSUE #930: trouble scaling cudf merge benchmark with ucx ~64 nodes on NERSC perlmutter
Dear dask-cuda devs,

We've been working with @quasiben to do some benchmarking with Dask UCX on NERSC's Perlmutter system. (You can see some background in @rcthomas issue [here](https://github.com/rapidsai/dask-cuda/issues/817)). So far I've been able to get the TCP-based [cudf merge benchmark](https://github.com/rapidsai/dask-cuda/blob/branch-22.08/dask_cuda/benchmarks/local_cudf_merge.py) to scale to about 256 nodes (it hit timeout but I think it would have finished) but I cannot get the UCX-based benchmark to run successfully at or above 64 nodes. The good news is that when it runs, UCX is quite a bit faster than TCP, but of course it would be nice to get it to scale up. You can view a [summary](https://gitlab.com/NERSC/python-benchmark/-/blob/main/dask-tcp-vs-ucx/output/summary.md) of the benchmarking data I've gathered so far. 

More details:

A few months ago @quasiben  kindly provided his [benchmark script](https://gist.github.com/quasiben/5e5c6a394d1a5fab8fb1ee0cb338ed8c) which I have lightly modified for my recent benchmarking efforts. You can view my modified version [here](https://gitlab.com/NERSC/python-benchmark/-/blob/main/dask-tcp-vs-ucx/nersc-ben-benchmark.sh).

I am using UCX 1.12.1 and the 22.06 rapids-nightly build. You can view my buildscript [here](https://gitlab.com/NERSC/python-benchmark/-/blob/main/dask-tcp-vs-ucx/build-dask-ucx-baremetal.sh). Note this is all being done outside a container using the NERSC shared filesystems. 

I have set `export UCXPY_LOG_LEVEL=DEBUG` and included the logs of all of my trials [here](https://gitlab.com/NERSC/python-benchmark/-/tree/main/dask-tcp-vs-ucx/output). Probably the most useful log is the [UCX 64 node run](https://gitlab.com/NERSC/python-benchmark/-/blob/main/dask-tcp-vs-ucx/output/dask-merge-bench-ucx-64-nodes/slurm-2319681.out). It's really huge and hard to navigate (sorry), but around line 4103066 you start to see errors like

```
[1654722743.050706] [nid001897:44756:0]         ib_mlx5.h:672  UCX  ERROR   madvise(DONTFORK, buf=0x55790ec0a000, len=131072) failed: Cannot allocate memory
[1654722743.050979] [nid001897:44756] UCXPY  DEBUG Error callback for endpoint 0x7f31609bad40 called with status -3: Input/output error
```

Seeing these I shrank the problem a bit compared to what @quasiben had done in his benchmark (40k vs 50k). I also made `--rmm-pool-size 35GB` a bit smaller. I also bumped up the ulimit as far as it can go on Perlmutter, although I don't know if this matters. 

Do you think this is really a memory issue? Based on [this](https://mymellanox.force.com/mellanoxcommunity/s/question/0D51T000080mq0jSAA/ucx-error-with-driver-51258-on-rhel-79) it sounds like it could be an RDMA config issue, but I am way out of my depth here so I'd appreciate any advice you have. 

Thank you very much,
Laurie

wence-:
Allows selection of the method multiprocessing uses to start child
processes. Additionally, in the forkserver case, ensure the fork
server is up and running before any computation happens.

Potentially fixes #930. Needs dask/distributed#6580.

cc: @pentschev, @quasiben

### DIFF
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -488,4 +488,12 @@ def parse_args():
 
 
 if __name__ == "__main__":
-    main(parse_args())
+    args = parse_args()
+    if args.multiprocessing_method == "forkserver":
+        import multiprocessing.forkserver as f
+
+        f.ensure_running()
+    with dask.config.set(
+        {"distributed.worker.multiprocessing-method": args.multiprocessing_method}
+    ):
+        main(args)
--- a/dask_cuda/benchmarks/local_cudf_shuffle.py
+++ b/dask_cuda/benchmarks/local_cudf_shuffle.py
@@ -305,4 +305,12 @@ def parse_args():
 
 
 if __name__ == "__main__":
-    main(parse_args())
+    args = parse_args()
+    if args.multiprocessing_method == "forkserver":
+        import multiprocessing.forkserver as f
+
+        f.ensure_running()
+    with dask.config.set(
+        {"distributed.worker.multiprocessing-method": args.multiprocessing_method}
+    ):
+        main(args)
--- a/dask_cuda/benchmarks/local_cupy.py
+++ b/dask_cuda/benchmarks/local_cupy.py
@@ -7,6 +7,7 @@ from warnings import filterwarnings
 import numpy as np
 from nvtx import end_range, start_range
 
+import dask
 from dask import array as da
 from dask.distributed import Client, performance_report, wait
 from dask.utils import format_bytes, format_time, parse_bytes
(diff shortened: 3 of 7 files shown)


pentschev:
Btw, the errors are driver issues that are being investigated by @galipremsagar , so we must wait for those to be fixed.

wence-:
> Does it make more sense to add that change directly into `LocalCUDACluster`/`dask-cuda-worker` as a `multiprocessing_method`/`--multiprocess-method`? We could then again expose that for the benchmarks, but perhaps it would be useful for other workflows rather than only these benchmarks. But that may raise other issues, for example, would there be additional concerns as to import/initialization ordering that we should document, or even better, detect and raise an error?

Probably yes. I would like to get confirmation that this is sufficient to fix the issues seen in #930 first.

It seems that ucx and ucx-py initialise lazily, so that as long as the forkserver is up and running before they touch the devices everything is fine. I'm not sure if it is possible to check for that programmatically.

I think certainly some documentation on how to get this working is worthwhile, certainly how to check whether one has managed the desired "no fork after IB bringup".

pentschev:
rerun tests

wence-:
This is now ready for review. From my testing, the combo of advice I have added to the docs seems sufficient to get past the node count limits we were seeing in #930.

pentschev (approved):
LGTM, this is really great work, thanks so much @wence- !

pentschev:
@gpucibot merge

PR #937, opened 2022-06-22
Add util functions to simplify printing benchmarks results

pentschev:
Two new utility functions are added to print benchmarks results to avoid the need to rearrange number of individual white spaces or separator lengths any time a new longer row is added, also preventing the count of spaces given indentation of code.


wence- (approved):
I think these look fine.

quasiben (approved):

quasiben:
@gpucibot merge

PR #938, opened 2022-06-22
Add communications bandwidth to benchmarks

pentschev:
The existing throughput would only give an overview of total data processed for the complete workflow, but no insight on communications which the new bandwidth value now does. Additionally, moved common peer-to-peer bandwidth computation as utility function.

Depends on #937

### DIFF
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -15,6 +15,8 @@ from dask.utils import format_bytes, format_time, parse_bytes
 from dask_cuda.benchmarks.utils import (
     get_cluster_options,
     get_scheduler_workers,
+    hmean,
+    hstd,
     parse_benchmark_args,
     peer_to_peer_bandwidths,
     plot_benchmark,
@@ -297,14 +299,25 @@ def main(args):
     times = np.asarray(times)
     bandwidths_all = np.asarray(bandwidths_all)
     print_separator(separator="=")
+
     print_key_value(
         key="Throughput",
         value=f"{format_bytes(t_p.mean())}/s +/- {format_bytes(t_p.std())}/s",
     )
     print_key_value(
+        key="ThroughputH",
+        value=f"{format_bytes(hmean(t_p))}/s +/- {format_bytes(hstd(t_p))}/s",
+    )
+    print_key_value(
         key="Bandwidth",
         value=f"{format_bytes(bandwidths_all.mean())}/s +/- "
         f"{format_bytes(bandwidths_all.std())}/s",
+    )
+
+    print_key_value(
+        key="BandwidthH",
+        value=f"{format_bytes(hmean(bandwidths_all))}/s +/- "
+        f"{format_bytes(hstd(bandwidths_all))}/s",
     )
     print_key_value(
         key="Wall clock",
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -299,12 +299,12 @@ def main(args):
     print_separator(separator="=")
     print_key_value(
         key="Throughput",
-        value=f"{format_bytes(t_p.mean())} +/- {format_bytes(t_p.std())}",
+        value=f"{format_bytes(t_p.mean())}/s +/- {format_bytes(t_p.std())}/s",
     )
     print_key_value(
         key="Bandwidth",
-        value=f"{format_bytes(bandwidths_all.mean())} +/- "
-        f"{format_bytes(bandwidths_all.std())}",
+        value=f"{format_bytes(bandwidths_all.mean())}/s +/- "
+        f"{format_bytes(bandwidths_all.std())}/s",
     )
     print_key_value(
         key="Wall clock",
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -298,11 +298,11 @@ def main(args):
     bandwidths_all = np.asarray(bandwidths_all)
     print_separator(separator="=")
     print_key_value(
-        key="Throughput (Total)",
+        key="Throughput",
         value=f"{format_bytes(t_p.mean())} +/- {format_bytes(t_p.std())}",
     )
     print_key_value(
-        key="Throughput (Comms only)",
+        key="Bandwidth",
         value=f"{format_bytes(bandwidths_all.mean())} +/- "
         f"{format_bytes(bandwidths_all.std())}",
     )
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -299,23 +299,12 @@ def main(args):
     times = np.asarray(times)
     bandwidths_all = np.asarray(bandwidths_all)
     print_separator(separator="=")
-
     print_key_value(
         key="Throughput",
-        value=f"{format_bytes(t_p.mean())}/s +/- {format_bytes(t_p.std())}/s",
-    )
-    print_key_value(
-        key="ThroughputH",
         value=f"{format_bytes(hmean(t_p))}/s +/- {format_bytes(hstd(t_p))}/s",
     )
     print_key_value(
         key="Bandwidth",
-        value=f"{format_bytes(bandwidths_all.mean())}/s +/- "
-        f"{format_bytes(bandwidths_all.std())}/s",
-    )
-
-    print_key_value(
-        key="BandwidthH",
         value=f"{format_bytes(hmean(bandwidths_all))}/s +/- "
         f"{format_bytes(hstd(bandwidths_all))}/s",
     )
--- a/dask_cuda/benchmarks/local_cudf_merge.py
+++ b/dask_cuda/benchmarks/local_cudf_merge.py
@@ -18,6 +18,8 @@ from dask_cuda.benchmarks.utils import (
     get_scheduler_workers,
     parse_benchmark_args,
     plot_benchmark,
+    print_key_value,
+    print_separator,
     setup_memory_pool,
 )
 from dask_cuda.utils import all_to_all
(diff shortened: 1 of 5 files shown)


dask_cuda/benchmarks/local_cupy.py
```diff
@@ -210,63 +212,72 @@ async def run(args):
...
+                times.append(took)
+                print_key_value(key=f"{m}", value=npartitions)
+            times = np.asarray(times)
+            bandwidths_all = np.asarray(bandwidths_all)
+            print_separator(separator="=")
+            print_key_value(
+                key="Bandwidth",
+                value=f"{format_bytes(bandwidths_all.mean())}/s +/- "
+                f"{format_bytes(bandwidths_all.std())}/s",
+            )
```
wence-, line 264: FWIW, and this is just the same as before, the arithmetic mean is generally the wrong thing for rate-like data, and the more correct option is the harmonic mean `len(xs)/sum(1/xs)`.

dask_cuda/benchmarks/local_cudf_shuffle.py
```diff
@@ -171,12 +156,34 @@ def main(args):
...
+    times = np.asarray(times)
+    bandwidths_all = np.asarray(bandwidths_all)
     print_separator(separator="=")
+    print_key_value(
+        key="Throughput (Total)",
+        value=f"{format_bytes(t_p.mean())} +/- {format_bytes(t_p.std())}",
+    )
+    print_key_value(
+        key="Throughput (Comms only)",
+        value=f"{format_bytes(bandwidths_all.mean())} +/- "
```
wence-: Whereas this bandwidth measure is the mean of the all the p2p data movement observeed by dask?

dask_cuda/benchmarks/local_cudf_shuffle.py
```diff
@@ -171,12 +156,34 @@ def main(args):
...
+        t_p.append(throughput)
         print_key_value(key=f"{m}", value=f"{format_bytes(throughput)}/s")
         t_runs[idx] = float(format_bytes(throughput).split(" ")[0])
+    t_p = np.asarray(t_p)
+    times = np.asarray(times)
+    bandwidths_all = np.asarray(bandwidths_all)
     print_separator(separator="=")
+    print_key_value(
+        key="Throughput (Total)",
+        value=f"{format_bytes(t_p.mean())} +/- {format_bytes(t_p.std())}",
```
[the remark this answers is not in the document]
wence-: Perhaps it is worth clarifying these differences?

wence- (commented):
Minor queries

dask_cuda/benchmarks/local_cupy.py
pentschev: Yeah, I was actually wondering if that was correct but didn't spend the time in investigating what would be the correct way to compute. Harmonic means do make sense indeed. However, it looks depressing 🙁 

```
Bandwidth                 | 10.18 GiB/s +/- 8.16 GiB/s
BandwidthH                | 3.47 GiB/s +/- 1.25 GiB/s
```

We _seemingly_ go down from 10.18GB/s to 3.47GB/s, of course the former is incorrect anyway, but it is sad.

I now replaced mean and standard deviation in https://github.com/rapidsai/dask-cuda/pull/938/commits/268f893b98faaccc87bea4f97060d502d513a15a, I am not sure whether standard deviation is correct, can you double-check?

dask_cuda/benchmarks/local_cudf_shuffle.py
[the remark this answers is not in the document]
pentschev: That is correct, although you commented on an old commit, the newest modifications read "Throughput" and "Bandwidth", do you think that naming is still unclear?

dask_cuda/benchmarks/local_cudf_shuffle.py
pentschev: Correct.

dask_cuda/benchmarks/local_cudf_merge.py
```diff
@@ -308,12 +293,18 @@ def main(args):
...
     print_key_value(
-        key="Throughput",
-        value=f"{format_bytes(t_p.mean())} +/- {format_bytes(t_p.std()) }",
+        key="Throughput (Total)",
+        value=f"{format_bytes(t_p.mean())} +/- {format_bytes(t_p.std())}",
+    )
+    print_key_value(
+        key="Throughput (Comms only)",
+        value=f"{format_bytes(bandwidths_all.mean())} +/- "
+        f"{format_bytes(bandwidths_all.std())}",
```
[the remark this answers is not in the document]
pentschev: Doesn't seem to be the case, although it's possible this is only for normal message passing (e.g., heartbeats):

```
$ DASK_DISTRIBUTED__COMM__COMPRESSION=None OPENBLAS_NUM_THREADS=1 python dask_cuda/benchmarks/local_cudf_merge.py -p ucx -d 0,1,2,3,4,5,6,7 -c 1_000_000 --runs 30 -b explicit-comms
...
Throughput                | 1.07 GiB/s +/- 209.63 MiB/s
Bandwidth                 | 103.27 MiB/s +/- 90.80 MiB/s
```

dask_cuda/benchmarks/local_cupy.py
wence-: Standard deviations of inverse random variables is a huge can of worms :(, but as per [this stats.stackexchange post](https://stats.stackexchange.com/q/7471) there are a few options. Suppose we have our rates $X_i, \quad i=[1,\dots,n]$, then define the reciprocal weights $Y_i$ and calculate

$$
\begin{align}
\mu_Y &= \frac{1}{n}\sum_i Y_i \\
\text{Var}(Y) &= \frac{1}{n}\sum_i (Y_i - \mu_Y)^2
\end{align}
$$

Then the variance of the harmonic mean of $X_i$ ($H_X := \frac{1}{\mu_Y}$) is 

$$
\text{Var}_H = \frac{\text{Var}(Y)}{n \mu_Y^4}
$$

The above-linked post shows that this is a good estimate for uniformly distributed data at least.

So I think we can calculate harmonic mean and standard deviation as

```python
def hmean(data):
    return 1/np.mean(1/data)

def hstd(data):
    rmean = np.mean(1/data)
    rvar = np.var(1/data)
    return np.sqrt(rvar / (len(data) * rmean**4))
```

dask_cuda/benchmarks/local_cudf_shuffle.py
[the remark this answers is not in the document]
wence-: I think those are OK.

dask_cuda/benchmarks/local_cudf_merge.py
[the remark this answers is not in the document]
wence-: So this must be background messaging since I don't think that the bandwidth can be lower than the throughput if we're actually measuring data movement of the computation.

dask_cuda/benchmarks/local_cupy.py
pentschev: Thanks, I saw that yesterday but wasn't totally sure that would be necessarily correct either. I've made the changes in https://github.com/rapidsai/dask-cuda/pull/938/commits/d120cd3e042382553b1b187a77abb6b08fe12c41 , thanks for the code as well!

dask_cuda/benchmarks/local_cudf_merge.py
```diff
@@ -308,12 +295,18 @@ def main(args):
...
-    t_p = numpy.asarray(t_p)
-    times = numpy.asarray(times)
+    t_p = np.asarray(t_p)
+    times = np.asarray(times)
+    bandwidths_all = np.asarray(bandwidths_all)
     print_separator(separator="=")
     print_key_value(
         key="Throughput",
-        value=f"{format_bytes(t_p.mean())} +/- {format_bytes(t_p.std()) }",
+        value=f"{format_bytes(hmean(t_p))}/s +/- {format_bytes(hstd(t_p))}/s",
```
wence-, line 304: I had a check to compare with how standard errors are propagated for reciprocals and these formulae are consistent with that approach. For this case `data_processed` is the same for every run and the times vary. So we can compute the mean throughput with `data_processed / np.mean(times)` and the standard deviation with `data_processed * np.std(times) / np.mean(times)**2`. But these are just special cases of the harmonic approach, so I'm happy.

dask_cuda/benchmarks/local_cupy.py
```diff
@@ -263,11 +249,25 @@ async def run(args):
...
             print_key_value(key="Worker thread(s)", value=f"{args.threads_per_worker}")
             print_separator(separator="=")
-            print_key_value(key="Wall clock", value="npartitions")
+            print_key_value(key="Wall clock", value="Partitions")
             print_separator(separator="-")
-            for (took, npartitions) in took_list:
-                t = format_time(took)
-                print_key_value(key=f"{t}", value=f"{npartitions}")
+            times = []
+            for idx, (took, npartitions) in enumerate(took_list):
```
wence-, line 255: `idx` unused.

dask_cuda/benchmarks/local_cupy_map_overlap.py
```diff
@@ -148,11 +134,25 @@ async def run(args):
...
             print_key_value(key="Worker thread(s)", value=f"{args.threads_per_worker}")
             print_separator(separator="=")
-            print_key_value(key="Wall clock", value="npartitions")
+            print_key_value(key="Wall clock", value="Partitions")
             print_separator(separator="-")
-            for (took, npartitions) in took_list:
-                t = format_time(took)
-                print_key_value(key=f"{t}", value=f"{npartitions}")
+            times = []
+            for idx, (took, npartitions) in enumerate(took_list):
```
wence-, line 140: `idx` unused

dask_cuda/benchmarks/local_cupy.py
```diff
@@ -263,11 +249,25 @@ async def run(args):
...
+                m = format_time(took)
+                times.append(took)
+                print_key_value(key=f"{m}", value=npartitions)
+            times = np.asarray(times)
+            bandwidths_all = np.asarray(bandwidths_all)
+            print_separator(separator="=")
+            print_key_value(
```
wence-, line 262: I think we can deduce throughput for these operations as well, but it would need to be done for each op in turn, and we'd need to pick a cache model. For example does `(A + A.T).sum()` move `A.nbytes` of data or `2*A.nbytes` (or more)?

wence- (approved):
nitpicks, but otherwise fine

quasiben (approved):
based on @wence- 

Probably good to address Lawrence's comments before merging in

dask_cuda/benchmarks/local_cupy.py
pentschev: The amount of data that gets moved (if by move you mean the same as transferring between workers) in the cuDF benchmarks is irrelevant for throughput, instead the throughput is based on the `data_processed` which is a sum of the DataFrames sizes:

https://github.com/rapidsai/dask-cuda/blob/a2e322362515bf612f9368eee99e150e6fa49437/dask_cuda/benchmarks/local_cudf_merge.py#L169-L170

In the same manner, I believe the equivalent `data_processed` for CuPy benchmarks here would be the sum of sizes of input arrays, e.g., `x.nbytes` for 

https://github.com/rapidsai/dask-cuda/blob/a2e322362515bf612f9368eee99e150e6fa49437/dask_cuda/benchmarks/local_cupy.py#L35

and `x.nbytes + y.nbytes` for

https://github.com/rapidsai/dask-cuda/blob/a2e322362515bf612f9368eee99e150e6fa49437/dask_cuda/benchmarks/local_cupy.py#L44-L45

IOW, the operation `A.T` size is not an input and thus doesn't get accounted for in `data_processed`. Does the above make sense?

wence-: I meant data movement in the local processing, but I guess this is a slightly philosophical argument. I think I was thinking from the point of view of what is the streaming data cost? e.g. if I just compute `B = A.T` then I need to stream `A` and stream the output to `B` so the data movement is 2*A.nbytes (on cpus at least it would normally be 3*A.nbytes since caches are usually write-allocate without hinting that the output can be done with a non-temporal store). This kind of metric tells us something about the performance on the hardware.

But I agree that a throughput of "this is how much data I started with, how fast can I process it?" is also a good metric, and is generally the one we actually care about since that's the wallclock time question. 

In conclusion: happy to go with sum of input array sizes.

dask_cuda/benchmarks/local_cupy_map_overlap.py
pentschev: Done in https://github.com/rapidsai/dask-cuda/commit/dc1a58715b0053eac7662229e8b892707a7d82e3

wence-:
> @wence- I think all your suggestions/requests have been addressed, do you think we're good to merge?

Yes, go ahead.

pentschev:
Thanks @wence- and @quasiben for reviews!

PR #943, opened 2022-07-04
LocalCUDACluster's memory limit: `None` means no limit

madsbk:
Use a regular `dict` when creating a `LocalCUDACluster` with no host and no device memory limit.

Currently, setting `device_memory_limit=None` translate into the total available GPU memory.  However,  in some cases `DeviceHostFile` overestimate the GPU memory usage, which can trigger spilling even though `device_memory_limit=None`.

### DIFF
--- a/dask_cuda/local_cuda_cluster.py
+++ b/dask_cuda/local_cuda_cluster.py
@@ -288,7 +288,9 @@ class LocalCUDACluster(LocalCluster):
 
         data = kwargs.pop("data", None)
         if data is None:
-            if self.jit_unspill:
+            if device_memory_limit is None and memory_limit is None:
+                data = {}
+            elif self.jit_unspill:
                 data = (
                     ProxifyHostFile,
                     {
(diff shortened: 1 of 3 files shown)


dask_cuda/local_cuda_cluster.py
```diff
@@ -288,7 +288,9 @@ def __init__(
 
         data = kwargs.pop("data", None)
         if data is None:
-            if self.jit_unspill:
+            if device_memory_limit is None and memory_limit is None:
+                data = {}
+            elif self.jit_unspill:
```
wence-, line 293: FWIW, and you're not changing this code, `self.jit_unspill` is a property that is only ever read in `__init__` so it could presumably not be set at all and you could just use `dask.config.get("jit-unspill", default=jit_unspill)`

wence-: Do you want to do that or shall I?

madsbk: If you have time, please do :)

pentschev (commented):
LGTM, thanks @madsbk .

pentschev:
@gpucibot merge

pentschev (approved):

PR #948, opened 2022-07-12
Always use versioneer command classes in setup.py

### LINKED ISSUE #336: Odd `__version__` reported

wence-:
The `GIT_DESCRIBE_TAG` and `VERSION_SUFFIX` environment variables are used
to control the name and version of the created conda/pypi package.
They should, however, not be used to control the version of the
installed package by overriding the versioneer cmdclass since that
leaves an unmodified _version.py file in the installed package
directory. A consequence is that the version reported by
`dask_cuda.__version__` is `"0+unknown"`. 

We cannot always use the versioneer-provided cmdclass unmodified since
PEP440 specifically forbids PyPI from accepting packages that have local
version identifiers (as used by versioneer). To get around this, when setup.py
detects it is building a PyPI package (`GIT_DESCRIBE_TAG` is in the environment),
patch the version returned from versioneer with a PyPI-compatible one.

While we're here, bring the conda version string into line with the
rest of the rapids ecosystem.

Closes #336.

### DIFF
--- a/conda/recipes/dask-cuda/meta.yaml
+++ b/conda/recipes/dask-cuda/meta.yaml
@@ -5,7 +5,6 @@
 {% set data = load_setup_py_data() %}
 
 {% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev').lstrip('v') + environ.get('VERSION_SUFFIX', '') %}
-{% set git_revision_count=environ.get('GIT_DESCRIBE_NUMBER', 0) %}
 {% set py_version=environ.get('CONDA_PY', 36) %}
 package:
   name: dask-cuda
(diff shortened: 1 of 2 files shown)


wence-:
Right now, `dask-cuda` installed from either pypi or the rapidsai and rapidsai-nightly conda channels reports:

```
$ python -c 'import dask_cuda; print(dask_cuda.__version__)'
0+unknown
```

I _think_ this is the right fix but am not really able to test if this will break package version resolution. I also don't really know much about how the conda packages are built, so careful eyes are welcome.

wence-:
Related to #336, and this approach was previously used in #448 which had breakages, and was partially reverted in #466.

wence-:
OK, I think I have a horrible hack that will work.

ajschmidt8 (approved):

jakirkham:
rerun tests

wence-:
rerun tests

pentschev:
rerun tests

jakirkham:
Let's keep an eye on [the pre-releases on PyPI]( https://pypi.org/project/dask-cuda/#history ) to make sure they are still getting published.

wence-:
> Let's keep an eye on [the pre-releases on PyPI](https://pypi.org/project/dask-cuda/#history) to make sure they are still getting published.

Looks like it: https://pypi.org/project/dask-cuda/22.8.0a220726/
```
>>> import dask_cuda
>>> dask_cuda.__version__
'v22.08.00a220726'
```

jakirkham:
Thanks Lawrence! 🙏

PR #955, opened 2022-07-21
Restore `CUDA_VISIBLE_DEVICES` overwritten by `Nanny`

pentschev:
Restore `CUDA_VISIBLE_DEVICES`, which are now overwritten by `Nanny`. This works around https://github.com/dask/distributed/issues/6749 .


pentschev:
The change for which we were attempting to work around was reverted in https://github.com/dask/distributed/pull/6777, and should be permanently dealt with in https://github.com/dask/distributed/issues/6780, therefore this PR should not be needed anymore, closing.

Issue #959, opened 2022-07-29
`@` operator fails with `ProxyObject` in CuPy 11.0.0


pentschev:
[CuPy 11.0.0 was released yesterday](https://github.com/cupy/cupy/releases/tag/v11.0.0) and some of the `ProxyObject` tests have begun failing, seemingly due to lack of support for the `@` operator providing `matmul` functionality:

<details>

```python
$ pytest dask_cuda/tests/test_proxy.py -k matmul -vs
========================================================= test session starts =========================================================
platform linux -- Python 3.8.13, pytest-7.1.2, pluggy-1.0.0 -- /datasets/pentschev/miniconda3/envs/rn-220729/bin/python
cachedir: .pytest_cache
rootdir: /datasets/pentschev/src/dask-cuda, configfile: pytest.ini
plugins: asyncio-0.19.0, repeat-0.9.1, rerunfailures-10.2, timeout-2.1.0
asyncio: mode=auto
collected 97 items / 95 deselected / 2 selected

dask_cuda/tests/test_proxy.py::test_cupy_matmul FAILED
dask_cuda/tests/test_proxy.py::test_cupy_imatmul FAILED

============================================================== FAILURES ===============================================================
__________________________________________________________ test_cupy_matmul ___________________________________________________________

    def test_cupy_matmul():
        cupy = pytest.importorskip("cupy")
        a, b = cupy.arange(10), cupy.arange(10)
        c = a @ b
        assert c == proxy_object.asproxy(a) @ b
>       assert c == a @ proxy_object.asproxy(b)

dask_cuda/tests/test_proxy.py:655:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   TypeError: Cannot convert ProxyObject to cupy._core.core._ndarray_base

cupy/_core/core.pyx:1287: TypeError
__________________________________________________________ test_cupy_imatmul __________________________________________________________

    def test_cupy_imatmul():
        cupy = pytest.importorskip("cupy")
        a = cupy.arange(9).reshape(3, 3)
        c = a.copy()
        c @= a

        a1 = a.copy()
>       a1 @= proxy_object.asproxy(a)

[...]

PR #960, opened 2022-07-29
Add `__rmatmul__` to `ProxyObject`

jakirkham:
Fixes https://github.com/rapidsai/dask-cuda/issues/959

If another object (like a `cupy.ndarray`) doesn't support `__matmul__` with `ProxyObject`, Python will try to fallback to `__rmatmul__`. As `__rmatmul__` was not defined before (and some cases like with CuPy now raise), this would raise. To fix, that this PR defines `__rmatmul__` for `ProxyObject`s to provide for this fallback layer.

### DIFF
--- a/dask_cuda/proxy_object.py
+++ b/dask_cuda/proxy_object.py
@@ -660,6 +660,9 @@ class ProxyObject:
     def __ror__(self, other):
         return other | self._pxy_deserialize()
 
+    def __rmatmul__(self, other):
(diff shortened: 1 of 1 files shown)


pentschev:
LGTM, thanks @jakirkham for working on this.

cc @madsbk for awareness.

pentschev:
@gpucibot merge

pentschev (approved):

PR #971, opened 2022-08-08
Unpin `dask` and `distributed` for development

galipremsagar:
This PR unpins `dask` & `distributed` for `22.10` development.

xref: https://github.com/rapidsai/cudf/pull/11492

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
-dask==2022.7.1
-distributed==2022.7.1
+dask>=2022.7.1
+distributed>=2022.7.1
 pynvml>=11.0.0
 numpy>=1.16.0
 numba>=0.54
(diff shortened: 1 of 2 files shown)


jakirkham:
rerun tests

( trying to make sure we are testing with PR https://github.com/dask/distributed/pull/6841 )

raydouglass (approved):

galipremsagar:
@gpucibot merge

pentschev:
FYI, no issues with https://github.com/dask/distributed/pull/6841 on a DGX-1 so far: https://github.com/dask/distributed/pull/6841#issuecomment-1208539135

requirements.txt
```diff
@@ -1,5 +1,5 @@
-dask==2022.7.1
-distributed==2022.7.1
+dask>=2022.7.1
+distributed>=2022.7.1
```
wence-, line 2: Should this be `>=2022.8` now that the August release is out? Or do we
not mind because the version resolution will pick the latest version
given the constraint?

pentschev: No, we only pin versions around code freeze time. During development we don't need to enforce a specific minimum version unless it's because some old version is problematic, given that's not the case here, it's useful to be able to go back if needed.

Issue #750, opened 2021-10-05
Document GPU best practices


jrbourbeau:
Some colleagues and I have started using distributed Dask clusters with GPU-enabled workers. During a quick chat with @quasiben, a few best practices came up (@quasiben feel free to correct me if I misunderstood your suggestions):

- Make sure to use `dask_cuda.CUDAWorker` for your cluster workers
- Set `rmm_pool_size` to be 80-90% of machine’s memory 
- Set `device_memory_limit` to be 60% of machine’s memory
- ...

I was wondering if there are other best practices the `dask-cuda` developers have for someone getting started with using GPUs in Dask clusters? Additionally, is there any interest in having a "Best Practices" page, similar to https://docs.dask.org/en/stable/best-practices.html, in the `dask-cuda` docs?

ayushdg:
Totally agree that a best practices page would be valuable. Based on my experience with gpu-bdb and other benchmarks here's what I've observed:
- device_memory_limit: ~ 55-60% if `DASK_JIT_UNSPILL=True`,  ~ 80% if false (since spilling is much faster and it enables more memory intensive operations at a much lower perf cost)
- rmm_pool_size: leave around 2GB of total gpu memory free. This usually amounts to ~90-98% of GPU memory on most cards. With the caveat that the workflow needs to account for any large data that might be computed and brought back to the client process with `compute`.
- Set `DASK_RMM__POOL_SIZE` ~0.5GB when working with the `ucx` protocol.

Potentially also worth either adding a separate section for UCX specific configuration recommendations or linking to [this page](https://docs.rapids.ai/api/dask-cuda/nightly/ucx.html#configuration) in the dask-cuda docs from the best practices page as well.

PR #976, opened 2022-08-15
Dask-cuda best practices

quasiben:
closes https://github.com/rapidsai/dask-cuda/issues/750

I've been getting more questions lately about how best to used RAPIDS/Dask-CUDA and found some time to jot down some quick thoughts on how "best" to setup Dask-CUDA for RAPIDS.  

This is by no means complete and if folks have time to amend/add please feel free to directly commit to the branch

### DIFF
--- a/docs/source/examples/best-practices.rst
+++ b/docs/source/examples/best-practices.rst
@@ -24,9 +24,9 @@ Accelerated Networking`_ for more discussion
 GPU Memory Management
 ~~~~~~~~~~~~~~~~~~~~~
 
-When using Dask-CUDA, especially with RAPIDS, it's best to use `RMM <https://docs.rapids.ai/api/rmm/stable/>`_ (RAPIDS Memory Manager)
-to pre-allocate memory on the GPU.  Allocating memory, while fast, takes a small amount of times, however, one can easily make
-hundreds of thousand or even millions of allocations in trivial workflows.  With RMM, allocations are sub-sampled from a larger
+When using Dask-CUDA, especially with RAPIDS, it's best to use an `RMM pool <https://docs.rapids.ai/api/rmm/stable/>`_ (RAPIDS Memory Manager)
+to pre-allocate memory on the GPU.  Allocating memory, while fast, takes a small amount of time, however, one can easily make
+hundreds of thousand or even millions of allocations in trivial workflows causing significant performance degradations.  With an RMM pool, allocations are sub-sampled from a larger
 pool and greatly reduces the allocation time and thereby increases performance:
 
 .. code-block:: python
--- a/docs/source/examples/best-practices.rst
+++ b/docs/source/examples/best-practices.rst
@@ -11,7 +11,7 @@ two reasons:
 
 - 1. Moving data between GPUs is costly and performance decreases when computation stops due to communication overheads, Host-to-Device/Device-to-Host transfers, etc
 - 2. Mutli-GPU instances often come with accelerated networking like `NVLink <https://www.nvidia.com/en-us/data-center/nvlink/>`_.  These accelerated
-networking paths usually have much higher throughput/bandwidth compared with traditional networking *and* don't force and H-to-D/D-to-H transfers.  See `
+networking paths usually have much higher throughput/bandwidth compared with traditional networking *and* don't force and Host-to-Device/Device-to-Host transfers.  See `
 Accelerated Networking`_ for more discussion
 
 .. code-block:: python
@@ -45,7 +45,7 @@ Accelerated Networking
 ~~~~~~~~~~~~~~~~~~~~~~
 
 As discussed in `Multi-GPUs`_, accelerated networking has better bandwidth/throughput compared with traditional networking hardware and does
-not force any costly H-to-D/D-to-H transfers.  Dask-CUDA can leverage accelerated networking hardware with `UCX-Py <https://ucx-py.readthedocs.io/en/latest/>`_.
+not force any costly Host-to-Device/Device-to-Host transfers.  Dask-CUDA can leverage accelerated networking hardware with `UCX-Py <https://ucx-py.readthedocs.io/en/latest/>`_.
 
 As an example, let's compare a merge benchmark when using 2 GPUs connected with NVLink.  First we'll run with standard TCP comms:
 
@@ -82,7 +82,7 @@ when Dask needed to move data back-and-forth between workers results in an avera
 
 To compare, we'll now change the ``procotol`` from ``tcp`` to ``ucx``:
 
-    python local_cudf_merge.py -d 0,1 -p ucx -c 50_000_000 --rmm-pool-size 28GB
+    python local_cudf_merge.py -d 0,1 -p ucx -c 50_000_000 --rmm-pool-size 30GB
 
 
 
--- a/docs/source/examples/best-practices.rst
+++ b/docs/source/examples/best-practices.rst
@@ -10,7 +10,7 @@ DGX, a cloud instance with `multi-gpu options <https://rapids.ai/cloud>`_ , a hi
 two reasons:
 
 - 1. Moving data between GPUs is costly and performance decreases when computation stops due to communication overheads, Host-to-Device/Device-to-Host transfers, etc
-- 2. Mutli-GPU instances often come with accelerated networking like `NVLink <https://www.nvidia.com/en-us/data-center/nvlink/>`_.  These accelerated
+- 2. Multi-GPU instances often come with accelerated networking like `NVLink <https://www.nvidia.com/en-us/data-center/nvlink/>`_.  These accelerated
 networking paths usually have much higher throughput/bandwidth compared with traditional networking *and* don't force and Host-to-Device/Device-to-Host transfers.  See `
 Accelerated Networking`_ for more discussion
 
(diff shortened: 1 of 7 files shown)


docs/source/examples/best-practices.rst
```diff
@@ -0,0 +1,112 @@
...
+
+    from dask_cuda import LocalCUDACluster
+
+    cluster = LocalCUDACluster(CUDA_VISIBLE_DEVICES="0,1",
+                               protocol="ucx")
+
+GPU Memory Management
+~~~~~~~~~~~~~~~~~~~~~
+
+When using Dask-CUDA, especially with RAPIDS, it's best to use `RMM <https://docs.rapids.ai/api/rmm/stable/>`_ (RAPIDS Memory Manager)
```
shwina, line 27: This is nitpicky, but RMM is used regardless. What's important is  to use an RMM _pool_.

pentschev: Agreed about making RMM _pool_ explicit. But saying it's used regardless is incorrect for the general case. it's true for cuDF, but CuPy will not automatically use RMM, for example.

quasiben: rewrote in d28dc55

docs/source/examples/best-practices.rst
```diff
@@ -0,0 +1,112 @@
...
+.. code-block:: python
+
+    from dask_cuda import LocalCUDACluster
+
+    cluster = LocalCUDACluster(CUDA_VISIBLE_DEVICES="0,1",
+                               protocol="ucx",
+                               rmm_pool_size="30GB")
+
+
+We also recommend allocating most, though not all, of the GPU memory space.  We do this so that temporary allocations, like those made deep inside CUDA libraries (libcudf, cupy, xgboost, etc) have room and the job is not halted due to an Out-Of-Memory (``OOM``) error.
```
pentschev, line 41: This is also not totally correct. cuDF and CuPy will use RMM to allocate memory, even temporary allocations (for CuPy that's only the case if it's internal memory management was replaced by RMM), I don't about xgboost. The problem with allocating the entire size is we don't know for example how large the CUDA context will be beforehand.

docs/source/examples/best-practices.rst
```diff
@@ -0,0 +1,112 @@
...
+Multi-GPUs
+~~~~~~~~~~
+
+When using multiple GPUs, if possible, it's best to co-locate as many GPUs on the same physical node.  This could be a
+DGX, a cloud instance with `multi-gpu options <https://rapids.ai/cloud>`_ , a high-density GPU HPC instance, etc.  This is done for
+two reasons:
+
+- 1. Moving data between GPUs is costly and performance decreases when computation stops due to communication overheads, Host-to-Device/Device-to-Host transfers, etc
+- 2. Mutli-GPU instances often come with accelerated networking like `NVLink <https://www.nvidia.com/en-us/data-center/nvlink/>`_.  These accelerated
+networking paths usually have much higher throughput/bandwidth compared with traditional networking *and* don't force and H-to-D/D-to-H transfers.  See `
```
pentschev, line 14: I would expand H-to-D/D-to-H to Host-to-Device/Device-to-Host. This will not be always immediately clear to a newcomer.

docs/source/examples/best-practices.rst
```diff
@@ -0,0 +1,112 @@
...
+
+
+We also recommend allocating most, though not all, of the GPU memory space.  We do this so that temporary allocations, like those made deep inside CUDA libraries (libcudf, cupy, xgboost, etc) have room and the job is not halted due to an Out-Of-Memory (``OOM``) error.
+
+
+Accelerated Networking
+~~~~~~~~~~~~~~~~~~~~~~
+
+As discussed in `Multi-GPUs`_, accelerated networking has better bandwidth/throughput compared with traditional networking hardware and does
+not force any costly H-to-D/D-to-H transfers.  Dask-CUDA can leverage accelerated networking hardware with `UCX-Py <https://ucx-py.readthedocs.io/en/latest/>`_.
```
pentschev, line 48: Ditto about expanding H-to-D/D-to-H.

docs/source/examples/best-practices.rst
```diff
@@ -0,0 +1,112 @@
...
+    Worker index              | Worker address
+    --------------------------------------------------------------------------------
+    0                         | tcp://127.0.0.1:44055
+    1                         | tcp://127.0.0.1:41095
+    ================================================================================
+
+
+To compare, we'll now change the ``procotol`` from ``tcp`` to ``ucx``:
+
+    python local_cudf_merge.py -d 0,1 -p ucx -c 50_000_000 --rmm-pool-size 28GB
```
pentschev, line 85:
```suggestion
    python local_cudf_merge.py -d 0,1 -p ucx -c 50_000_000 --rmm-pool-size 30GB
```

pentschev: Just to make the examples totally compatible.

pentschev (changes requested):

docs/source/examples/best-practices.rst
wence-: Note as well that if `rmm_pool_size=None` then RMM will pick an initial pool for size `total_device_memory / 2`

docs/source/examples/best-practices.rst
```diff
@@ -0,0 +1,112 @@
...
+not force any costly H-to-D/D-to-H transfers.  Dask-CUDA can leverage accelerated networking hardware with `UCX-Py <https://ucx-py.readthedocs.io/en/latest/>`_.
+
+As an example, let's compare a merge benchmark when using 2 GPUs connected with NVLink.  First we'll run with standard TCP comms:
+
+::
+
+    python local_cudf_merge.py -d 0,1 -p tcp -c 50_000_000 --rmm-pool-size 30GB
+
+
+In the above, we used 2 GPUs (2 dask-cuda-workers), pre-allocated 30GB of GPU ram (to make gpu memory allocations faster), and used TCP comms
```
wence-, line 57:
```suggestion
In the above, we used 2 GPUs (2 dask-cuda-workers), pre-allocated 30GB of GPU RAM (to make gpu memory allocations faster), and used TCP comms
```

docs/source/examples/best-practices.rst
```diff
@@ -0,0 +1,112 @@
...
+when Dask needed to move data back-and-forth between workers results in an average wall clock time of: ``19.72 s +/- 694.36 ms``::
```
wence-, line 58:
```suggestion
when Dask needed to move data back-and-forth between workers. This setup results in an average wall clock time of: ``19.72 s +/- 694.36 ms``::
```

docs/source/examples/best-practices.rst
```diff
@@ -0,0 +1,112 @@
...
+    cluster = LocalCUDACluster(CUDA_VISIBLE_DEVICES="0,1",
+                               protocol="ucx")
+
+GPU Memory Management
+~~~~~~~~~~~~~~~~~~~~~
+
+When using Dask-CUDA, especially with RAPIDS, it's best to use an `RMM pool <https://docs.rapids.ai/api/rmm/stable/>`_ (RAPIDS Memory Manager)
+to pre-allocate memory on the GPU.  Allocating memory, while fast, takes a small amount of time, however, one can easily make
+hundreds of thousand or even millions of allocations in trivial workflows causing significant performance degradations.  With an RMM pool, allocations are sub-sampled from a larger
+pool and greatly reduces the allocation time and thereby increases performance:
```
wence-, line 30: This also has a big positive effect when using UCX with cuda IPC: we only need to register a single IPC handle for the whole pool (which is expensive, but only done once) since from the IPC point of viewer there's only a single allocation. As opposed to just using RMM without a pool where each new allocation must be registered with IPC.

docs/source/examples/best-practices.rst
```diff
@@ -0,0 +1,112 @@
+Best Practices
+==============
+
+
+Multi-GPUs
+~~~~~~~~~~
+
+When using multiple GPUs, if possible, it's best to co-locate as many GPUs on the same physical node.  This could be a
```
wence-, line 8: Do we need to say _how_ to co-locate GPUs? Or do you just mean "When choosing between two multi-GPU setups, it is best to pick the one where most GPUs are co-located with one-another".

docs/source/examples/best-practices.rst
```diff
@@ -0,0 +1,112 @@
...
+DGX, a cloud instance with `multi-gpu options <https://rapids.ai/cloud>`_ , a high-density GPU HPC instance, etc.  This is done for
```
wence-, line 9: Link to the DGX nvidia pages?

docs/source/examples/best-practices.rst
pentschev: That isn't true. In Dask `rmm_pool_size=None` means it's disabled by default. The behavior you're thinking is within `rmm.reinitialize`, where passing `pool_allocator=True` will cause RMM to pick a pool with initial size of half the GPU's memory.

docs/source/examples/best-practices.rst
```diff
@@ -0,0 +1,112 @@
...
+
+    from dask_cuda import LocalCUDACluster
+
+    cluster = LocalCUDACluster(CUDA_VISIBLE_DEVICES="0,1",
+                               protocol="ucx")
+
+GPU Memory Management
+~~~~~~~~~~~~~~~~~~~~~
+
+When using Dask-CUDA, especially with RAPIDS, it's best to use an `RMM pool <https://docs.rapids.ai/api/rmm/stable/>`_ (RAPIDS Memory Manager)
```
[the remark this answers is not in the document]
quasiben: Nesting is annoying.  I was able to implement your suggestion with `replace`:

```
  .. |rmm-pool| replace:: :abbr:`RMM (RAPIDS Memory Manager)` pool
  __ https://docs.rapids.ai/api/rmm/stable/
```

quasiben:
Thanks all for the comments.  I think I've addressed all of them in the last commit.  Could folks review again ?  

Also, should we include a section on spilling.  @ayushdg had some thoughts awhile back:
https://github.com/rapidsai/dask-cuda/issues/750#issuecomment-934899888

docs/source/examples/best-practices.rst
quasiben: done

docs/source/examples/best-practices.rst
wence-: Ah, sorry, yes.

pentschev:
CI failures are being resolved in https://github.com/rapidsai/dask-cuda/pull/977 .

pentschev (approved):

madsbk (approved):

quasiben:
Is this ok to merge and we can continue to iterate with recommendations for spilling ?

wence- (approved):
Thanks, looks good!

wence-:
> Is this ok to merge and we can continue to iterate with recommendations for spilling ?

Yes, I think so.

quasiben:
@gpucibot merge

PR #979, opened 2022-08-18
[WIP] Add Groupby benchmark

rjzamora:
Add simple groupby benchmark using the same template as `local_cudf_merge.py` and `local_cudf_shuffle.py`.

### DIFF
--- a/dask_cuda/benchmarks/local_cudf_groupby.py
+++ b/dask_cuda/benchmarks/local_cudf_groupby.py
@@ -17,14 +17,18 @@ from dask_cuda.benchmarks.utils import (
 )
 
 
-def apply_groupby(df, shuffle=None):
+def apply_groupby(df, split_out=1, split_every=8, shuffle=None):
+    # TODO: Add support for shuffle
+    if shuffle:
+        raise NotImplementedError(
+            "shuffle-based groupby benchmark not fully implemented yet."
+        )
     wait(
-        # TODO: Add params for split_out and sort
         df.groupby("key", sort=False)
         .agg(
-            {"mean": ["x", "y"], "count": "x"},
-            split_out=1,
-            split_every=8,
+            {"int64": ["max", "count"], "float64": "mean"},
+            split_out=split_out,
+            split_every=split_every,
             # shuffle=shuffle,
         )
         .persist()
@@ -63,7 +67,7 @@ def get_random_ddf(args):
     return dd.from_map(
         generate_chunk,
         [(i, args.chunk_size) for i in range(args.in_parts)],
-        meta=generate_chunk(0, **chunk_kwargs),
+        meta=generate_chunk((0, 0), **chunk_kwargs),
         **chunk_kwargs,
     )
 
@@ -83,7 +87,12 @@ def bench_once(client, args, write_profile=None):
 
     with ctx:
         t1 = clock()
-        apply_groupby(df, shuffle=args.shuffle)
+        apply_groupby(
+            df,
+            split_out=args.split_out,
+            split_every=args.split_every,
+            shuffle=args.shuffle,
+        )
         t2 = clock()
 
     return (data_processed, t2 - t1)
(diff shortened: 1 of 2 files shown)


dask_cuda/benchmarks/utils.py
```diff
@@ -279,6 +279,7 @@ def get_cluster_options(args):
             "CUDA_VISIBLE_DEVICES": args.devs,
             "interface": args.interface,
             "device_memory_limit": args.device_memory_limit,
+            "local_directory": "/raid/dask-space/rzamora/scratch",
```
rjzamora, line 282:
```suggestion
```

dask_cuda/benchmarks/local_cudf_groupby.py
```diff
@@ -0,0 +1,273 @@
...
+            "help": "Chunk size (default 1_000_000)",
+        },
+        {
+            "name": "--unique-ratio",
+            "default": 0.01,
+            "type": float,
+            "help": "Fraction of rows that are unique groups",
+        },
+        {
+            "name": "--sort",
```
rjzamora, line 211: Note that I am seeing errors when I run the benchmark with `--shuffle "explicit-comms" --split_out 2 --sort`, and the cause seems to be that dask is using [a shallow copy](https://github.com/dask/dask/blob/f382d9e8439f5ed200da6ce66df7b2b33a0fc500/dask/dataframe/utils.py#L726) in [DataFrame.drop](https://github.com/dask/dask/blob/f382d9e8439f5ed200da6ce66df7b2b33a0fc500/dask/dataframe/core.py#L5304), and [this](https://github.com/rapidsai/cudf/blob/4023b65233d8ade1b19f883b1e48abf614813254/python/dask_cudf/dask_cudf/sorting.py#L293) `"_partitions"`-column `drop` is somehow not working right after an explicit-comms shuffle.

I'm not exactly sure why the `"_partitions"` column is not getting dropped properly here (I am struggling to create a simple-enough reproducer to properly explore). However, I suspect that the custom round-tripping of of the input dataframe from a persisted list of futures is somehow running us into an undefined edge case.

wence-: I see a (possibly different) error:

```
python -m dask_cuda.benchmarks.local_cudf_groupby --shuffle explicit-comms  --split_out 2 --sort -d 0,1,3,4   -p tcp
=>   File "/root/dask/dask/dataframe/utils.py", line 429, in check_matching_columns
    raise ValueError(
ValueError: The columns in the computed data do not match the columns in the provided metadata
Order of columns does not match
```
With the columns being:
```
Index(['key', 'int64___count', 'int64___max', 'float64___count',
       'float64___sum'],
      dtype='object')

Index(['key', 'int64___max', 'int64___count', 'float64___sum',
       'float64___count'],
      dtype='object')
```

rjzamora: >This occurs also with --shuffle tasks

Oh no! You see this same error for the command you shared above, but with `--shuffle tasks`?

wence-: yes (not always), more likely with more devices

rjzamora: Thanks for pointing this out @wence-.  It seems like we need to tweak the groupby-aggregation functions (defined in `dask_cudf`) to output intermediate `DataFrame` results with **sorted** column names. Otherwise the ordering is non-deterministic, and we open ourselves up to failed metadata-enforcement checks in down-stream operations :/

rjzamora: **Update**: https://github.com/rapidsai/cudf/pull/11836 should resolve this problem in `dask_cudf`.

madsbk (changes requested):
Thanks @rjzamora LGTM, I only have a minor change

dask_cuda/benchmarks/local_cudf_groupby.py
```diff
@@ -0,0 +1,273 @@
...
+        {
+            "name": "--runs",
+            "default": 3,
+            "type": int,
+            "help": "Number of runs",
+        },
+    ]
+
+    return parse_benchmark_args(
+        description="Distributed shuffle (dask/cudf) benchmark", args_list=special_args
```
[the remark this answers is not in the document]
rjzamora: Oops. Thanks @madsbk !

quasiben:
@madsbk can you accept the change then we can merge in

madsbk (approved):

quasiben:
@gpucibot merge

rjzamora:
Thanks @madsbk and @quasiben !

PR #981, opened 2022-08-24
Switch to the new dask CLI

jacobtomlinson:
In https://github.com/dask/dask/pull/9283 we are adding a new top level `dask` CLI command which can be extended by other modules using entry points. A primary motivation here is to improve discoverability by uniting everything under one tool and allowing folks to run `dask --help` and `dask <subcommand> --help` to learn more about the various tools.

This PR adds a new `click` group called `cuda` and moves the `dask-cuda-worker` command under that group with the name `worker`. This means the `dask-cuda-worker` becomes `dask cuda worker` in the new CLI tool.

I haven't made any changes to the existing `dask-cuda-worker` console script so that will still continue to work, but maybe we should add a deprecation warning to it?

I went with this name rather than `dask cuda-worker` because I think it is more readable and also leaves us open to adding more subcommands in the future without cluttering up the top-level `dask` namespace.

```console
$ dask --help             
Usage: dask [OPTIONS] COMMAND [ARGS]...

  Dask command line interface.

Options:
  --version   Show the version and exit.
  -h, --help  Show this message and exit.

Commands:
  cluster    Manage dask clusters.
  cuda       GPU subcommands.
  docs       Open Dask documentation (https://docs.dask.org/) in a web browser.
  info       Information about your dask installation.
  scheduler  Launch a distributed scheduler.
  ssh        Launch a distributed cluster over SSH.
  worker     Launch a distributed worker attached to an existing SCHEDULER.
```

```console
$ dask cuda --help        
Usage: dask cuda [OPTIONS] COMMAND [ARGS]...

  GPU subcommands.

Options:
  -h, --help  Show this message and exit.

Commands:
  worker  Launch a distributed worker with GPUs attached to an existing SCHEDULER.
```

```console
$ dask cuda worker --help  
Usage: dask cuda worker [OPTIONS] [SCHEDULER] [PRELOAD_ARGV]...

  Launch a distributed worker with GPUs attached to an existing SCHEDULER.

  See https://docs.rapids.ai/api/dask-cuda/stable/quickstart.html#dask-cuda-worker for
  info.

Options:
  --host TEXT                     IP address of serving host; should be visible to the
                                  scheduler and other workers. Can be a string (like
                                  ``"127.0.0.1"``) or ``None`` to fall back on the
                                  address of the interface specified by
                                  ``--interface`` or the default interface.
  --nthreads INTEGER              Number of threads to be used for each Dask worker
                                  process.  [default: 1]
...
```

The CLI PR needs to be merged and released before this can be merged.

Fixes https://github.com/rapidsai/dask-cuda/issues/1038


pentschev:
The idea seems nice, but this is a very intricate component in many scripts and I imagine this will be annoying to abruptly change. Would we be able to keep `dask-cuda-worker` for some time with a `DeprecationWarning` added to it?

wence-:
> The idea seems nice, but this is a very intricate component in many scripts and I imagine this will be annoying to abruptly change. Would we be able to keep `dask-cuda-worker` for some time with a `DeprecationWarning` added to it?

AIUI, the old entry-point continues to exist and will work. I agree that a deprecation warning is a good idea, but not sure exactly how to hook that up.

pentschev:
> AIUI, the old entry-point continues to exist and will work.

Yes, you're right. That was my bad, I didn't pay a lot of attention to the actual changes.

> I agree that a deprecation warning is a good idea, but not sure exactly how to hook that up.

I think we can simply check if `dask_cuda_worker.py` is running without the `cuda` argument and then raise the warning.

jacobtomlinson:
This is still blocked on the upstream PR

jacobtomlinson:
The new CLI has been merged and released so we should be good to merge this now.

I've bumped the minimum version of `dask` and `distributed` to one with the CLI although I'm not sure this is actually necessary as folks can still use `dask-cuda-worker` and the new `dask cuda` CLI namespace is opt-in.

requirements.txt
```diff
@@ -1,5 +1,5 @@
-dask==2022.9.2
-distributed==2022.9.2
+dask>=2022.10.1
+distributed>=2022.10.1
```
pentschev, line 2: Not strictly necessary, but not harmful either, we currently pin `>=2022.9.2` so we rely on the latest Dask/Distributed, only at code freeze we will increase the pin.

pentschev (commented):
Overall LGTM, thanks @jacobtomlinson . I now wonder, could we have a `dask cuda scheduler` without increasing code duplication too much now? I think this could be useful to have some default overrides, more specifically `DASK_DISTRIBUTED__COMM__UCX__CREATE_CUDA_CONTEXT=True`.

setup.py
```diff
@@ -59,5 +59,7 @@ def get_versions():
         [console_scripts]
         dask-cuda-worker=dask_cuda.cli.dask_cuda_worker:go
         dask-config=dask_cuda.cli.dask_config:go
```
jakirkham, line 61: What would we like to do with this entry point?

jakirkham:
It would be good to add the entry point to the Conda recipe as well

https://github.com/conda-forge/dask-cuda-feedstock/blob/c870fd90153b47cc0fe8060593bf309e27da3b32/recipe/meta.yaml#L17-L18

setup.py
pentschev: I assume you're referring to the `dask-config` one, this is from https://github.com/rapidsai/dask-cuda/pull/1006.

jacobtomlinson: Ideally I'd like that functionality to move from `dask-cuda` to `dask-ctl`, especially given that it isn't particularly GPU specific. cc @quasiben 

Alternatively if it is going to stay here I expect it would become `dask cuda config` and we would need to update the click groups accordingly.

jakirkham: Yep we were discussing renaming in issue ( https://github.com/rapidsai/dask-cuda/issues/1038 ). Feel free to share thoughts there 🙂

If it will take a while to move this functionality to `dask-ctl`, renaming may be prudent as an intermediate step in any event

dask_cuda/cli.py
```diff
@@ -5,21 +5,28 @@
...
+from .utils import print_cluster_config
 
 logger = logging.getLogger(__name__)
 
 
 pem_file_option_type = click.Path(exists=True, resolve_path=True)
 
 
-@click.command(context_settings=dict(ignore_unknown_options=True))
+@click.group
```
charlesbluca, line 24: Not sure if all `cuda.commands` need to reside in the same file as the initial definition to get picked up correctly? Made a few failed attempts to keep the initial structure of the CLI modules before settling on having them all in one shared module, which is similar to what dask-ctl does for its grouped `cluster`/`discovery` commands.

Assuming this is the only way to organize to the CLI commands, we could probably consolidate things up a bit by reusing shared click arguments/options.

dask_cuda/cli.py
```diff
@@ -5,21 +5,28 @@
...
+def cuda():
+    """GPU subcommands."""
```
charlesbluca, line 26: Was thinking a more detailed description like

```suggestion
    """Subcommands to launch or query distributed workers with GPUs."""
```

Could be nice here but happy to hear feedback

conda/recipes/dask-cuda/meta.yaml
```diff
@@ -41,6 +41,11 @@ requirements:
 test:
   imports:
     - dask_cuda
+  commands:
+    - dask-cuda-worker --help
+    - dask-cuda-config --help
+    - dask cuda worker --help
+    - dask cuda config --help
```
charlesbluca, line 48: Considered setting these CLI tests programmatically, but don't see any obvious way to grab the new CLI commands from the pyproject.toml.

cc @jakirkham in case there's something I'm missing

dask_cuda/cli.py
```diff
@@ -404,9 +423,52 @@ async def run():
...
+@scheduler_file
+@click.option(
+    "--get-cluster-configuration",
+    "get_cluster_conf",
+    default=False,
+    is_flag=True,
+    required=False,
+    show_default=True,
+    help="""Print a table of the current cluster configuration""",
+)
```
pentschev, line 438: I guess this is the only argument and without it nothing will happen. Does it make more sense to remove and make the default action (without any non-mandatory arguments) print the configuration?

dask_cuda/cli.py
```diff
@@ -324,6 +337,12 @@ def main(
     multiprocessing_method,
     **kwargs,
 ):
+    """Launch a distributed worker with GPUs attached to an existing SCHEDULER.
```
pentschev, line 340: Not a big deal, but are there any reasons for `SCHEDULER` to be all uppercase? Is it referring to something that is normally written in uppercase? If not, I would suggest rewriting this.

pentschev:
```suggestion
    """Launch a distributed worker with GPUs attached to an existing dask scheduler.
```

dask_cuda/cli.py
```diff
@@ -5,25 +5,62 @@
...
+scheduler = click.argument("scheduler", type=str, required=False)
+preload_argv = click.argument(
     "preload_argv", nargs=-1, type=click.UNPROCESSED, callback=validate_preload_argv
 )
+scheduler_file = click.option(
+    "--scheduler-file",
+    type=str,
+    default=None,
+    help="""Filename to JSON encoded scheduler information. To be used in conjunction
+    with the equivalent ``dask-scheduler`` option.""",
```
pentschev, line 31:
```suggestion
    with the equivalent ``dask scheduler`` option.""",
```

pentschev: Not absolutely sure, but I imagine the hyphen here is also not necessary anymore (or deprecated)?

pentschev (commented):
Thanks @charlesbluca for picking this up, I've left a few suggestions/questions.

conda/recipes/dask-cuda/meta.yaml
jakirkham: There's some code above for the all `-` ones

The ones with ` `s can't be done programmatically AFAIK

ajschmidt8:
I canceled the latest GH Action run because it looks like the python tests were in-progress for over 4 hours, which is dramatically longer than the usual ~30 minutes that they take.

The logs are below, it looks like there were some errors throughout the tests.

- https://github.com/rapidsai/dask-cuda/actions/runs/3912570589/jobs/6687484798


To prevent this from happening again, I will open a PR to put a reasonable timeout on the Python tests.

This also happened on #1081.

dask_cuda/cli.py
charlesbluca: My assumption was that it was to make it clear that it corresponded to the `SCHEDULER` argument that crops up in the `--help` output:

```
Usage: dask cuda worker [OPTIONS] [SCHEDULER] [PRELOAD_ARGV]...

  Launch a distributed worker with GPUs attached to an existing SCHEDULER.
```

It looks like this is also done for `dask worker`, though I can see how this description could seem somewhat strange from the `dask cuda` output.

Is there a way in click to render documentation for arguments? If so, that could be a good place to give more detail on what the expected input for `SCHEDULER` is; if not, I guess a reasonable compromise is to take your suggestion and dump info about the `SCHEDULER` argument in the subsequent lines of the docstring so that it only appears when running `dask cuda worker`.

dask_cuda/cli.py
charlesbluca: Good catch! Yeah I can go through and switch out instances of the old Dask CLI for the new.

dask_cuda/cli.py
charlesbluca: Yeah was thinking that as well, just wasn't sure if it made more sense to do in a separate PR since the diff here is fairly large already.

(cc @quasiben in case there's something non-obvious this command does when `--get-cluster-configuration` isn't passed as an option)

dask_cuda/cli.py
pentschev: I see your point, it feels a little unusual but I'm ok with either one, if your prefer we can leave it as it is. I have no idea about click rendering documentation, not exactly sure how that would even work on the CLI, so my guess is that this probably doesn't exist.

dask_cuda/cli.py
pentschev: I _think_ @jacobtomlinson had done it all (or almost all), so I hope there won't be many more.

conda/recipes/dask-cuda/meta.yaml
```diff
@@ -41,6 +41,13 @@ requirements:
 test:
   imports:
     - dask_cuda
+  commands:
+    {% for e in data.get("project", {}).get("scripts", {}).keys() %}
+    - {{ e }} --help
+    {% endfor %}
+    - dask cuda --help
+    - dask cuda worker --help
+    - dask cuda config --help
```
jakirkham, line 50: Actually we might be able to do this to consolidate them further

```suggestion
    - dask cuda --help
    {% for e in data.get("project", {}).get("scripts", {}).keys() %}
    - {{ e }} --help
    - {{ e|replace("-", " ") }} --help
    {% endfor %}
```

charlesbluca: Good catch! Yeah for now we can do this, interested in if we intend to deprecate and eventually remove the old entrypoints, which would force us to hardcode this some time in the future (cc @jacobtomlinson if you've had any discussion around this)

dask_cuda/tests/test_dask_cuda_worker.py
```diff
@@ -194,18 +208,20 @@ def test_pre_import(loop):  # noqa: F811
     module = None
 
     # Pick a module that isn't currently loaded
-    for m in pkg_resources.working_set:
-        if m.key not in sys.modules.keys():
-            module = m.key
+    for m in pkgutil.iter_modules():
+        if m.ispkg and m.name not in sys.modules.keys():
+            module = m.name
```
charlesbluca, line 213: Changed this because IIUC `pkg_resources.working_set` is listing the installed distributions and not necessarily the importable modules; this becomes an issue if the distribution and module names aren't the same (i.e. one would `conda install pillow` and then `import PIL`), which was causing some failures in CI that seem unrelated to the changes here.

charlesbluca: Decided to break this off into https://github.com/rapidsai/dask-cuda/pull/1085 since it seems like this issue is blocking CI elsewhere (https://github.com/rapidsai/dask-cuda/pull/1081 for example)

conda/recipes/dask-cuda/meta.yaml
jakirkham: Yeah that's fair. For that reason wasn't sure whether we would want to do this or not

dask_cuda/tests/test_dask_cuda_worker.py
pentschev: Thanks @charlesbluca , the fix should get merged as soon as tests pass.

jakirkham: Had a question about using [`importlib.metadata`]( https://docs.python.org/3/library/importlib.metadata.html ) from the Python standard library ( https://github.com/rapidsai/dask-cuda/pull/1085#issuecomment-1385921939 )

ajschmidt8 (approved):
Approving `ops-codeowner` file changes

jakirkham (approved):

pentschev (approved):
LGTM, thanks @jacobtomlinson and @charlesbluca for working on this!

jakirkham:
Have noted the corresponding changes for the conda-forge recipe in issue ( https://github.com/conda-forge/dask-cuda-feedstock/issues/17 ). Please feel free to note anything there I may have missed

PR #985, opened 2022-08-31
benchmarks: Add option to suppress output of point to point data

wence-:
When running with many workers the point to point data is not that
insightful, so add command-line flag to suppress it (default is still
to show data).

### DIFF
--- a/dask_cuda/benchmarks/utils.py
+++ b/dask_cuda/benchmarks/utils.py
@@ -186,15 +186,8 @@ def parse_benchmark_args(description="Generic dask-cuda Benchmark", args_list=[]
         "one worker per GPU will be launched.",
     )
     parser.add_argument(
-        "--show-p2p-bandwidth",
-        action="store_true",
-        help="Produce detailed point to point bandwidth stats in output",
-        default=True,
-    )
-    parser.add_argument(
         "--no-show-p2p-bandwidth",
-        action="store_false",
-        dest="show_p2p_bandwidth",
+        action="store_true",
         help="Do not produce detailed point to point bandwidth stats in output",
     )
     parser.add_argument(
@@ -561,7 +554,7 @@ def print_throughput_bandwidth(
         key="Wall clock",
         value=f"{format_time(durations.mean())} +/- {format_time(durations.std()) }",
     )
-    if args.show_p2p_bandwidth:
+    if not args.no_show_p2p_bandwidth:
         print_separator(separator="=")
         if args.markdown:
             print("<details>\n<summary>Worker-Worker Transfer Rates</summary>\n\n```")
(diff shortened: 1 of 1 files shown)


dask_cuda/benchmarks/utils.py
```diff
@@ -185,6 +185,18 @@ def parse_benchmark_args(description="Generic dask-cuda Benchmark", args_list=[]
         "Note: --devs is currently ignored in multi-node mode and for each host "
         "one worker per GPU will be launched.",
     )
+    parser.add_argument(
+        "--show-p2p-bandwidth",
+        action="store_true",
+        help="Produce detailed point to point bandwidth stats in output",
+        default=True,
+    )
```
pentschev, line 193:
```suggestion
```

pentschev: We don't need this, just the `--no-...` option should suffice to assume the enabled option is the default (e.g., https://github.com/rapidsai/dask-cuda/blob/86a9846b1b8f1bea7bdc37c1df09a75f981f1fbf/dask_cuda/benchmarks/utils.py#L193-L197 ).

dask_cuda/benchmarks/utils.py
```diff
@@ -185,6 +185,18 @@ def parse_benchmark_args(description="Generic dask-cuda Benchmark", args_list=[]
...
+    parser.add_argument(
+        "--no-show-p2p-bandwidth",
+        action="store_false",
+        dest="show_p2p_bandwidth",
```
pentschev, line 197:
```suggestion
        action="store_true",
```

dask_cuda/benchmarks/utils.py
```diff
@@ -549,28 +561,29 @@ def print_throughput_bandwidth(
...
-            )
-    print_separator(separator="=")
-    print_key_value(key="Worker index", value="Worker address")
-    print_separator(separator="-")
-    for address, index in sorted(address_to_index.items(), key=itemgetter(1)):
-        print_key_value(key=index, value=address)
-    print_separator(separator="=")
-    if args.markdown:
-        print("```\n</details>\n")
+    if args.show_p2p_bandwidth:
```
pentschev, line 564:
```suggestion
    if not args.no_show_p2p_bandwidth:
```

pentschev (changes requested):
Looks good, but I think we can shorten the code quite a bit by only adding the `--no-...` option. Added suggestions for that.

wence-:
> Looks good, but I think we can shorten the code quite a bit by only adding the `--no-...` option. Added suggestions for that.

Thanks, done.

pentschev (approved):
LGTM, thanks @wence- !

PR #992, opened 2022-09-22
Update rearrange_by_column patch for explicit comms

rjzamora:
Motivated by discussions in https://github.com/rapidsai/cudf/pull/11576 (primarilty [here](https://github.com/rapidsai/cudf/pull/11576#discussion_r977813158)). This PR updates the `rearrange_by_columns_tasks` wrapper to target `rearrange_by_columns` instead. The updated wrapper also handles the case that `shuffle="explicit-comms"`.

### DIFF
--- a/dask_cuda/explicit_comms/dataframe/shuffle.py
+++ b/dask_cuda/explicit_comms/dataframe/shuffle.py
@@ -344,7 +344,7 @@ def shuffle(
     return new_dd_object(dsk, name, meta, divs).persist()
 
 
-def get_rearrange_by_column_tasks_wrapper(func):
+def get_rearrange_by_column_wrapper(func):
     """Returns a function wrapper that dispatch the shuffle to explicit-comms.
 
     Notice, this is monkey patched into Dask at dask_cuda import
@@ -354,23 +354,29 @@ def get_rearrange_by_column_tasks_wrapper(func):
 
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
-        if dask.config.get("explicit-comms", False):
+        # Use explicit-comms shuffle if the shuffle kwarg is
+        # set to "explicit-comms", or if it is set to "tasks"
+        # and the `dask.config` specifies "explicit-comms"
+        shuffle_arg = kwargs.pop("shuffle", None) or dask.config.get("shuffle", "disk")
+        if shuffle_arg == "explicit-comms" or (
+            dask.config.get("explicit-comms", False) and shuffle_arg == "tasks"
+        ):
             try:
                 import distributed.worker
 
                 # Make sure we have an activate client.
                 distributed.worker.get_client()
             except (ImportError, ValueError):
-                pass
+                shuffle_arg = "tasks"  # Fall back to task-based shuffle
             else:
                 # Convert `*args, **kwargs` to a dict of `keyword -> values`
                 kw = func_sig.bind(*args, **kwargs)
                 kw.apply_defaults()
                 kw = kw.arguments
-                column = kw["column"]
+                column = kw["col"]
                 if isinstance(column, str):
                     column = [column]
                 return shuffle(kw["df"], column, kw["npartitions"], kw["ignore_index"])
-        return func(*args, **kwargs)
+        return func(*args, shuffle=shuffle_arg, **kwargs)
 
     return wrapper
--- a/dask_cuda/explicit_comms/dataframe/shuffle.py
+++ b/dask_cuda/explicit_comms/dataframe/shuffle.py
@@ -1,6 +1,7 @@
 import asyncio
 import functools
 import inspect
+import warnings
 from collections import defaultdict
 from operator import getitem
 from typing import Dict, List, Optional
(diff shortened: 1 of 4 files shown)


rjzamora:
cc @madsbk @wence-

rjzamora:
@madsbk - Do you have any intuition on [these failures](https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/dask-cuda/job/prb/job/dask-cuda-gpu-test/CUDA=11.2,GPU_LABEL=driver-495,LINUX_VER=ubuntu18.04,PYTHON=3.9/899/testReport/junit/dask_cuda.tests/test_proxify_host_file/test_compatibility_mode_dataframe_shuffle_True_1_/)

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -354,23 +354,29 @@ def get_rearrange_by_column_tasks_wrapper(func):
...
     @functools.wraps(func)
     def wrapper(*args, **kwargs):
-        if dask.config.get("explicit-comms", False):
+        # Use explicit-comms shuffle if the shuffle kwarg is
+        # set to "explicit-comms", or if it is set to "tasks"
+        # and the `dask.config` specifies "explicit-comms"
+        shuffle_arg = kwargs.pop("shuffle", None) or dask.config.get("shuffle", "disk")
+        if shuffle_arg == "explicit-comms" or (
+            dask.config.get("explicit-comms", False) and shuffle_arg == "tasks"
+        ):
```
wence-, line 363: Is this the minimally-surprising behaviour? My feeling is that the narrower specification (`shuffle="tasks"` as a keyword argument) should take precedence over the configuration variable.

Before this PR, there was one way to enable an explicit-comms shuffle: set `dask.config["explicit-comms"]` to True, which changed the meaning of `shuffle="tasks"` to be `shuffle="explicit-comms"`.

Now that we are hooking in one level higher up, I think we can (unless we want to maintain backwards-compat) switch to a setup where the only way to request the explicit-comms shuffle is via `dask.config["shuffle"] = "explicit-comms"` or `shuffle="explicit-comms"`, and remove the `dask.config["explicit-comms"]` option (unless we anticipate needing it for other things as well).

Concretely, I'm suggesting:

```
use_explicit_comms = kwargs.get("shuffle", dask.config.get("shuffle", None)) == "explicit-comms"
if use_explicit_comms:
    try:
        import distributed.worker
        distributed.worker.get_client()
    except (ImportError, ValueError):
        warning("Requested explicit-comms shuffle, but no distributed client available, using task-based shuffle")
        # slight action-at-a-distance here because this affects the call to `func` below in the (implicit) "not use_explicit_comms" branch.
        kwargs["shuffle"] = "tasks"
    else:
        ... # do the explicit-comms thing
return func(*args, **kwargs)

**Review by wence- — CHANGES_REQUESTED** [review]
More for discussion than definite change...

File: dask_cuda/explicit_comms/dataframe/shuffle.py (hunk 1 above)
Comment (rjzamora) [reply dask_cuda/explicit_comms/dataframe/shuffle.py:363]: >Now that we are hooking in one level higher up, I think we can (unless we want to maintain backwards-compat) switch to a setup where...

Yes. I agree that we should move away from the `"explicit-comms"` config, but I was indeed trying to avoid a "breaking" change. With that said, I have no evidence that anyone is actually **unsing** the `"explicit-comms"` config in the wild.  So, perhaps I was being too conservative.

Even if people **are** using the `"explicit-comms"` config, we should probably add a deprecation warning here, and tell them to use the `"shuffle"` option and config instead. What do you think about allowing the `"explicit-comms"`/`"tasks"` for now, but with a `FutureWarning`?

File: docs/source/explicit_comms.rst [hunk 2]
```diff
@@ -11,7 +11,6 @@ implementation of the Dataframe `shuffle <https://github.com/rapidsai/dask-cuda/
 Usage
 -----
 
-In order to use explicit-comms in Dask/Distributed automatically, simply define the environment variable ``DASK_EXPLICIT_COMMS=True`` or setting the ``"explicit-comms"``
-key in the `Dask configuration <https://docs.dask.org/en/latest/configuration.html>`_.
+In order to use explicit-comms in Dask/Distributed for shuffle-based DataFrame operations, simply pass the ``shuffle="explicit-comms"`` key-word argument to the appropriate DataFrame API (e.g. ``ddf.sort_values(..., shuffle="explicit-comms")```). You can also change the default shuffle algorithm to explicit-comms by define the ``DASK_SHUFFLE=explicit-comms`` environment variable, or by setting the ``"shuffle"`` key to ``"explicit-comms"`` in the `Dask configuration <https://docs.dask.org/en/latest/configuration.html>`_.
```
rjzamora, line 14: @madsbk - Does this change seem okay to you? My understanding is that the `EXPLICIT_COMMS` config/env-variable was only being used for the shuffle/rarrange_by_column_tasks patch.  As far as I can tell, we never use this config to automatically use the explicit-comms merge anywhere. Is that correct?

madsbk: Yes, this is correct.

dask_cuda/explicit_comms/dataframe/shuffle.py
madsbk: Agree, this is a good solution.

madsbk (approved):
LGTM, thanks @rjzamora

docs/source/explicit_comms.rst
```diff
@@ -2,7 +2,7 @@ Explicit-comms
 ==============
 
 Communication and scheduling overhead can be a major bottleneck in Dask/Distributed. Dask-CUDA addresses this by introducing an API for explicit communication in Dask tasks.
-The idea is that Dask/Distributed spawns workers and distribute data as usually while the user can submit tasks on the workers that communicate explicitly.
+The idea is that Dask/Distributed spawns workers and distribute data as usuall while the user can submit tasks on the workers that communicate explicitly.
```
wence-:
```suggestion
The idea is that Dask/Distributed spawns workers and distributes data as usual while the user can submit tasks on the workers that communicate explicitly.
```

wence- (approved):
Thanks!

rjzamora:
@gpucibot merge

PR #994, opened 2022-09-26
Add benchmark option to use dask-noop

wence-:
To add some metrics on distributed overheads to our benchmark suite, I propose using https://github.com/gjoseph92/dask-noop/ to rewrite the compute graphs into no-ops. Now we just measure the scheduler and worker-to-worker performance on zero-sized work (or something close thereto).

Interestingly the cupy-based transpose runs take _longer_ if I noopify (cc @gjoseph92 any insight here?), but I haven't done anything systematic yet.

### DIFF
--- a/dask_cuda/benchmarks/common.py
+++ b/dask_cuda/benchmarks/common.py
@@ -153,7 +153,7 @@ def run_client_from_existing_scheduler(args: Namespace, config: Config):
             client.shutdown()
 
 
-def run_create_client(args, config):
+def run_create_client(args: Namespace, config: Config):
     """Create a client + cluster and run
 
     Shuts down the cluster at the end of the benchmark"""
--- a/dask_cuda/benchmarks/local_cudf_groupby.py
+++ b/dask_cuda/benchmarks/local_cudf_groupby.py
@@ -11,6 +11,7 @@ from dask.utils import format_bytes, parse_bytes
 
 from dask_cuda.benchmarks.common import Config, execute_benchmark
 from dask_cuda.benchmarks.utils import (
+    as_noop,
     parse_benchmark_args,
     print_key_value,
     print_separator,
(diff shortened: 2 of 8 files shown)


madsbk:
rerun tests

pentschev (approved):
LGTM. I wasn't aware of dask-noop, but it seems like an interesting tool to benchmark indeed. Thanks @wence- .

pentschev:
I believe the failing test `dask_cuda.tests.test_proxify_host_file.test_worker_force_spill_to_disk` may be flaky. I've seen it a few times in my nightly runs (e.g., https://github.com/pentschev/ucx-py-ci/issues/1116) but I think I didn't see it in gpuCI before.

@madsbk any ideas on what could be the cause?

pentschev:
rerun tests

pentschev:
I've opened https://github.com/rapidsai/dask-cuda/issues/997 to track the `test_worker_force_spill_to_disk` issue.

pentschev:
Still looks fine, anything else missing here or are we good to merge @wence- ?

wence-:
I added noop to the groupby benchmark. I think we're good to merge, thanks!

pentschev:
Thanks for the work here @wence- !

pentschev:
@gpucibot merge

gjoseph92:
@wence- dask-noop is pretty experimental and I'm not sure how well it works. One thing I notice at a glance is that it doesn't apply graph optimizations. So when it switches the collection type from Array to Delayed, you'll lose out on all the array graph optimizations. So that could explain why transpose is slower?

wence-:
> @wence- dask-noop is pretty experimental and I'm not sure how well it works. One thing I notice at a glance is that it doesn't apply graph optimizations. So when it switches the collection type from Array to Delayed, you'll lose out on all the array graph optimizations. So that could explain why transpose is slower?

Ah, thanks. I will have a poke.

Issue #989, opened 2022-09-20
[FEA] Client script to export dask and dask-cuda configuration of  a dask-cuda cluster


VibhuJawa:
## Request
We should have a script that a user can run on a client to export configuration and environment variables of  a dask-cuda/dask cluster . 

## Motivation:

Cluster configuration is often complex and requires the users to setup both in `dask-cuda` and `dask`   . It would really help if there was a way to get those exported easily into a file  to ensure that their setup is indeed correct. 

Common configuration errors seen are with the following configurations: 

##### Dask-Cuda Level:

- Transfer Protocol being userd (ucx/tcp) 
- Communication Hardware actually being used under the hood
-  Jit-Unspill 
- rmm-pool-size
- device-memory-limit

##### Dask Level 
 
- DASK_DISTRIBUTED__WORKERS__MEMORY__LIMIT
- DASK_DISTRIBUTED__COMM__TIMEOUTS__CONNECT

Potential UX:

```python
from dask_cuda import get_configration 
client = Client(.....)

get_configration(client, 'current_configuration.json')
```

Then the `current_configuration.json` can be easily shared with us to use. 

CC: @quasiben , @rlratzel , @randerzander , @jnke2016

quasiben:
What do you think of the following:

## Cluster Setup 1

> cluster = LocalCUDACluster(rmm_pool_size='5GB', protocol='ucx')

```python
In [80]: pprint(list(client.run(lambda dask_worker: worker_config(dask_worker)).values())[0])
{'async_alloc': False,
 'distributed.comm.timeouts': {'connect': '30s', 'tcp': '30s'},
 'initial_pool_size': 5000000000,
 'jit-unspill': False,
 'libraries': [],
 'log_directory': None,
 'logging': False,
 'managed_memory': False,
 'maximum_pool_size': None,
 'memory_limit': 135263113216,
 'memory_pause_fraction': 0.8,
 'memory_spill_fraction': 0.7,
 'memory_target_fraction': 0.6,
 'protocol': 'ucx',
 'rmm_track_allocations': False,
 'ucx-transports': {'cma',
                    'cuda_copy',
                    'cuda_ipc',
                    'dc_mlx5',
                    'posix',
                    'rc_mlx5',
                    'rc_verbs',
                    'self',
                    'sysv',
                    'tcp',
                    'ud_mlx5',
                    'ud_verbs'}}
```

## Cluster Setup 2

[...]

PR #1006, opened 2022-10-04
Add feature to get cluster configuration

quasiben:
Early attempt at resolving https://github.com/rapidsai/dask-cuda/issues/989

Screen Shot 2022-10-04 at 6 37 20 PM

In an ideal world, we could call `dask.config.config` or `dask.config.get(...)` and that will have been updated with live values.   Unfortunately, that is not the case.  Config options are set in all manner of places: cli, env, yaml, kwargs, etc and we are not consistent about storing them.  For example, in this PR `self.device_memory_limit = device_memory_limit` is added to `device_host_file.py` just so `client.run` could be accessed from within the worker.

This change employs a hodge-podge of techniques to get at options we _think_ may be potentially useful for debugging but in no way is this exhaustive.

### DIFF
--- a/dask_cuda/cli/dask_config.py
+++ b/dask_cuda/cli/dask_config.py
@@ -93,9 +93,5 @@ def main(
         sys.exit(0)
 
 
-def go():
-    main()
-
-
 if __name__ == "__main__":
-    go()
+    main()
--- a/dask_cuda/cli/dask_config.py
+++ b/dask_cuda/cli/dask_config.py
@@ -90,7 +90,6 @@ def main(
         else:
             client = Client(scheduler, security=security)
         print_cluster_config(client)
-        sys.exit(0)
 
 
 if __name__ == "__main__":
--- a/dask_cuda/cli/dask_config.py
+++ b/dask_cuda/cli/dask_config.py
@@ -1,0 +1,101 @@
+from __future__ import absolute_import, division, print_function
+
+import logging
+import sys
+
+import click
+
+from distributed import Client
+from distributed.preloading import validate_preload_argv
+from distributed.security import Security
+
+from ..utils import print_cluster_config
+
+logger = logging.getLogger(__name__)
+
+
+pem_file_option_type = click.Path(exists=True, resolve_path=True)
+
+
+@click.command(context_settings=dict(ignore_unknown_options=True))
+@click.argument("scheduler", type=str, required=False)
+@click.argument(
+    "preload_argv", nargs=-1, type=click.UNPROCESSED, callback=validate_preload_argv
+)
+@click.option(
+    "--scheduler-file",
+    type=str,
+    default=None,
+    help="""Filename to JSON encoded scheduler information. To be used in conjunction
+    with the equivalent ``dask-scheduler`` option.""",
+)
+@click.option(
+    "--get-cluster-configuration",
+    "get_cluster_conf",
+    default=False,
+    is_flag=True,
+    required=False,
+    show_default=True,
+    help="""Print a table of the current cluster configuration""",
+)
+@click.option(
+    "--tls-ca-file",
+    type=pem_file_option_type,
+    default=None,
+    help="""CA certificate(s) file for TLS (in PEM format). Can be a string (like
+    ``"path/to/certs"``), or ``None`` for no certificate(s).""",
+)
+@click.option(
+    "--tls-cert",
+    type=pem_file_option_type,
+    default=None,
+    help="""Certificate file for TLS (in PEM format). Can be a string (like
+    ``"path/to/certs"``), or ``None`` for no certificate(s).""",
+)
+@click.option(
+    "--tls-key",
+    type=pem_file_option_type,
+    default=None,
+    help="""Private key file for TLS (in PEM format). Can be a string (like
+    ``"path/to/certs"``), or ``None`` for no private key.""",
+)
+def main(
+    scheduler,
+    scheduler_file,
+    get_cluster_conf,
+    tls_ca_file,
+    tls_cert,
+    tls_key,
+    **kwargs,
+):
+    if tls_ca_file and tls_cert and tls_key:
+        security = Security(
+            tls_ca_file=tls_ca_file,
+            tls_worker_cert=tls_cert,
+            tls_worker_key=tls_key,
+        )
+    else:
+        security = None
+
+    if isinstance(scheduler, str) and scheduler.startswith("-"):
+        raise ValueError(
+            "The scheduler address can't start with '-'. Please check "
+            "your command line arguments, you probably attempted to use "
+            "unsupported one. Scheduler address: %s" % scheduler
+        )
+
+    if get_cluster_conf:
+        if scheduler_file is not None:
+            client = Client(scheduler_file=scheduler_file, security=security)
+        else:
+            client = Client(scheduler, security=security)
+        print_cluster_config(client)
+        sys.exit(0)
+
+
+def go():
+    main()
+
+
+if __name__ == "__main__":
+    go()
--- a/dask_cuda/cli/dask_config.py
+++ b/dask_cuda/cli/dask_config.py
@@ -1,7 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
 import logging
-import sys
 
 import click
 
--- a/dask_cuda/cli/dask_cuda_worker.py
+++ b/dask_cuda/cli/dask_cuda_worker.py
@@ -179,7 +179,7 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
 @click.option(
     "--scheduler-file",
     type=str,
-    default="",
+    default=None,
     help="""Filename to JSON encoded scheduler information. To be used in conjunction
     with the equivalent ``dask-scheduler`` option.""",
 )
@@ -336,11 +336,6 @@ def main(
     get_cluster_conf,
     **kwargs,
 ):
-    if get_cluster_conf:
-        client = Client(scheduler)
-        print_cluster_config(client)
-        sys.exit(0)
-
     if multiprocessing_method == "forkserver":
         import multiprocessing.forkserver as f
 
@@ -360,6 +355,14 @@ def main(
             "your command line arguments, you probably attempted to use "
             "unsupported one. Scheduler address: %s" % scheduler
         )
+
+    if get_cluster_conf:
+        if scheduler_file is not None:
+            client = Client(scheduler_file=scheduler_file, security=security)
+        else:
+            client = Client(scheduler, security=security)
+        print_cluster_config(client)
+        sys.exit(0)
 
     if worker_class is not None:
         worker_class = import_term(worker_class)
--- a/dask_cuda/cli/dask_cuda_worker.py
+++ b/dask_cuda/cli/dask_cuda_worker.py
@@ -1,17 +1,20 @@
 from __future__ import absolute_import, division, print_function
 
 import logging
+import sys
 
 import click
 from tornado.ioloop import IOLoop, TimeoutError
 
 from dask import config
+from distributed import Client
 from distributed.cli.utils import install_signal_handlers
 from distributed.preloading import validate_preload_argv
 from distributed.security import Security
 from distributed.utils import import_term
 
 from ..cuda_worker import CUDAWorker
+from ..utils import get_cluster_configuration
 
 logger = logging.getLogger(__name__)
 
@@ -287,6 +290,15 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     type=click.Choice(["spawn", "fork", "forkserver"]),
     help="""Method used to start new processes with multiprocessing""",
 )
+@click.option(
+    "--get-cluster-configuration",
+    "get_cluster_conf",
+    default=False,
+    is_flag=True,
+    required=False,
+    show_default=True,
+    help="""Print a table of the current cluster configuration""",
+)
 def main(
     scheduler,
     host,
@@ -321,8 +333,14 @@ def main(
     worker_class,
     pre_import,
     multiprocessing_method,
+    get_cluster_conf,
     **kwargs,
 ):
+    if get_cluster_conf:
+        client = Client(scheduler)
+        get_cluster_configuration(client, table=True)
+        sys.exit(0)
+
     if multiprocessing_method == "forkserver":
         import multiprocessing.forkserver as f
 
--- a/dask_cuda/cli/dask_cuda_worker.py
+++ b/dask_cuda/cli/dask_cuda_worker.py
@@ -14,7 +14,7 @@ from distributed.security import Security
 from distributed.utils import import_term
 
 from ..cuda_worker import CUDAWorker
-from ..utils import get_cluster_configuration
+from ..utils import print_cluster_config
 
 logger = logging.getLogger(__name__)
 
@@ -338,7 +338,7 @@ def main(
 ):
     if get_cluster_conf:
         client = Client(scheduler)
-        get_cluster_configuration(client, table=True)
+        print_cluster_config(client)
         sys.exit(0)
 
     if multiprocessing_method == "forkserver":
--- a/dask_cuda/cli/dask_cuda_worker.py
+++ b/dask_cuda/cli/dask_cuda_worker.py
@@ -1,20 +1,17 @@
 from __future__ import absolute_import, division, print_function
 
 import logging
-import sys
 
 import click
 from tornado.ioloop import IOLoop, TimeoutError
 
 from dask import config
-from distributed import Client
 from distributed.cli.utils import install_signal_handlers
 from distributed.preloading import validate_preload_argv
 from distributed.security import Security
 from distributed.utils import import_term
 
 from ..cuda_worker import CUDAWorker
-from ..utils import print_cluster_config
 
 logger = logging.getLogger(__name__)
 
@@ -290,15 +287,6 @@ pem_file_option_type = click.Path(exists=True, resolve_path=True)
     type=click.Choice(["spawn", "fork", "forkserver"]),
     help="""Method used to start new processes with multiprocessing""",
 )
-@click.option(
-    "--get-cluster-configuration",
-    "get_cluster_conf",
-    default=False,
-    is_flag=True,
-    required=False,
-    show_default=True,
-    help="""Print a table of the current cluster configuration""",
-)
 def main(
     scheduler,
     host,
@@ -333,7 +321,6 @@ def main(
     worker_class,
     pre_import,
     multiprocessing_method,
-    get_cluster_conf,
     **kwargs,
 ):
     if multiprocessing_method == "forkserver":
@@ -355,14 +342,6 @@ def main(
             "your command line arguments, you probably attempted to use "
             "unsupported one. Scheduler address: %s" % scheduler
         )
-
-    if get_cluster_conf:
-        if scheduler_file is not None:
-            client = Client(scheduler_file=scheduler_file, security=security)
-        else:
-            client = Client(scheduler, security=security)
-        print_cluster_config(client)
-        sys.exit(0)
 
     if worker_class is not None:
         worker_class = import_term(worker_class)
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -180,8 +180,6 @@ class DeviceHostFile(ZictBase):
         local_directory=None,
         log_spilling=False,
     ):
-        # can now access value within dask-worker / client.run
-        self.device_memory_limit = device_memory_limit
         self.disk_func_path = os.path.join(
             local_directory or dask.config.get("temporary-directory") or os.getcwd(),
             "dask-worker-space",
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -180,6 +180,8 @@ class DeviceHostFile(ZictBase):
         local_directory=None,
         log_spilling=False,
     ):
+        # can now access value within dask-worker / client.run
+        self.device_memory_limit = device_memory_limit
         self.disk_func_path = os.path.join(
             local_directory or dask.config.get("temporary-directory") or os.getcwd(),
             "dask-worker-space",
--- a/dask_cuda/tests/test_local_cuda_cluster.py
+++ b/dask_cuda/tests/test_local_cuda_cluster.py
@@ -385,7 +385,7 @@ async def test_get_cluster_configuration():
     ) as cluster:
         async with Client(cluster, asynchronous=True) as client:
             ret = await get_cluster_configuration(client)
-            assert ret["RMMSetup"]["initial_pool_size"] == 2000000000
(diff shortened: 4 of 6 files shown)


dask_cuda/utils.py
```diff
@@ -649,3 +649,101 @@ def get_gpu_uuid_from_index(device_index=0):
     pynvml.nvmlInit()
     handle = pynvml.nvmlDeviceGetHandleByIndex(device_index)
     return pynvml.nvmlDeviceGetUUID(handle).decode("utf-8")
+
+
+def get_worker_config(dask_worker):
+    import dask
+    from distributed.comm import parse_address
```
madsbk, line 656: Do we need the imports inside the function?

dask_cuda/utils.py
```diff
@@ -649,3 +649,101 @@ def get_gpu_uuid_from_index(device_index=0):
...
+    for mem in [
+        "memory_limit",
+        "memory_pause_fraction",
+        "memory_spill_fraction",
+        "memory_target_fraction",
+    ]:
+        ret[mem] = getattr(dask_worker.memory_manager, mem)
+
+    # jit unspilling set
+    ret["jit-unspill"] = hasattr(dask_worker.data, "spill_on_demand_initialized")
```
madsbk, line 679: Maybe use `isinstance(dask_worker.data, ProxifyHostFile)` to make it clear what we consider JIT-Unspill

dask_cuda/device_host_file.py
```diff
@@ -180,6 +180,8 @@ def __init__(
         local_directory=None,
         log_spilling=False,
     ):
+        # can now access value within dask-worker / client.run
+        self.device_memory_limit = device_memory_limit
```
madsbk, line 184: Avoid this and use `data.device_buffer.n` to get device limit instead

dask_cuda/utils.py
```diff
@@ -649,3 +649,101 @@ def get_gpu_uuid_from_index(device_index=0):
...
+
+    # get optional device-memory-limit
+    if ret["jit-unspill"]:
+        ret["device-memory-limit"] = dask_worker.data.manager._device_memory_limit
+    else:
+        device_memory_limit = getattr(dask_worker.data, "device_memory_limit", False)
```
madsbk, line 685: Use `dask_worker.data.device_buffer.n` to get device limit.

madsbk (commented):
Looks good, I only have some minor suggestions

dask_cuda/utils.py
```diff
@@ -649,3 +649,101 @@ def get_gpu_uuid_from_index(device_index=0):
...
+
+        table = Table(title="Dask Cluster Configuration")
+        table.add_column("Parameter", justify="left", style="red")
+        table.add_column("Value", justify="left", style="green")
+
+        params = sorted(ret.keys())
+        for p in params:
+            table.add_row(p, str(ret[p]))
+        console = Console()
+        console.print(table)
```
pentschev, line 749: Maybe make a CLI out of this? I can imagine this causing all sorts of issues if accidentally called from code when `rich` is installed, as it may try to rich-print on the console from arbitrary places.

pentschev (commented):
Besides Mads suggestions, I have one more, but overall looks real nice @quasiben !

dask_cuda/utils.py
quasiben: I added a CLI but left the rich code inside of `utils`.  i don't think it will cause any `rich-printing` .  My thought was that folks like @VibhuJawa / users may still want a nicely formatted table from within a live session

pentschev: > I added a CLI

I think a separate CLI (maybe `dask-get-config` or `dask-print-config`) would be better, given that has nothing to do with `dask-cuda-worker`.

> but left the rich code inside of `utils`. i don't think it will cause any `rich-printing` . My thought was that folks like @VibhuJawa / users may still want a nicely formatted table from within a live session

I agree a nicely formatted table is a nice addition to a live session, and maybe this is my ignorance of `rich` but I thought it was a tool to print nicely formatted tables on a terminal, am I mistaken? What if you call that code arbitrarily from any place in the code, say you want to parse that output and feed it to some other code, what happens then (assuming you have `rich` installed)?

VibhuJawa: > I added a CLI but left the rich code inside of utils. i don't think it will cause any rich-printing . My thought was that folks like @VibhuJawa / users may still want a nicely formatted table from within a live session

I like the nicely formatted output .   I guess if we want to feed it into some other code we just stick with the default of `Table=False` .

VibhuJawa (changes requested):
Looks great to. 

Requesting a small change to make all the memory limits human readable as just parsing bytes is hard and this mainly for users to debug the cluster configuration.

dask_cuda/utils.py
quasiben: Rich was, I think, originally designed for terminals but notebook support has existed for a 2 years or so:
https://github.com/Textualize/rich/issues/26

If you'd prefer a text only output I can look into the version mismatch output in dask:

```
+-------------+----------+-----------+----------+
| Package     | client   | scheduler | workers  |
+-------------+----------+-----------+----------+
| dask        | 2022.9.2 | 2022.7.1  | 2022.9.2 |
| distributed | 2022.9.2 | 2022.7.1  | 2022.9.2 |
+-------------+----------+-----------+----------+
```

@jacobtomlinson do you know if there are known issues with using rich in a jupyter session ?

dask_cuda/utils.py
```diff
@@ -649,3 +653,105 @@ def get_gpu_uuid_from_index(device_index=0):
...
+    # device and host memory configuration
+    for p in plugin_vals:
+        for v in dir(p):
+            # ignore hidden attrs as well as setup and CPUAffinity
+            if v.startswith("_") or v == "setup" or v == "cores":
+                pass
+            else:
+                ret[v] = getattr(p, v)
+    for mem in [
+        "memory_limit",
```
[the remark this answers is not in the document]
quasiben: yup yup, i'll add that in

dask_cuda/utils.py
```diff
@@ -649,3 +653,105 @@ def get_gpu_uuid_from_index(device_index=0):
...
+    ret = client.sync(
+        _get_cluster_configuration, client=client, asynchronous=client.asynchronous
+    )
+    if not table:
+        return ret
+    else:
+        try:
+            from rich.console import Console
+            from rich.table import Table
+        except ImportError:
```
wence-, line 742: I would advocate a stronger separation of acquiring the data from its presentation. That is, `get_cluster_configuration` should just return "raw" data, and there should be a separate function that prints a tabulated version.

dask_cuda/utils.py
```diff
@@ -649,3 +653,105 @@ def get_gpu_uuid_from_index(device_index=0):
...
+            print(
+                "Please install rich `python -m pip install rich`"
+                / "to print a table of the current Dask Cluster Configuration"
+            )
+            sys.exit(1)
```
wence-, line 747: This is a very unfriendly way to complain about a missing package. Prefer `raise`ing some reasonable error here.

dask_cuda/utils.py
```diff
@@ -649,3 +653,105 @@ def get_gpu_uuid_from_index(device_index=0):
...
+
+        table = Table(title="Dask Cluster Configuration")
+        table.add_column("Parameter", justify="left", style="red")
+        table.add_column("Value", justify="left", style="green")
+
+        params = sorted(ret.keys())
+        for p in params:
+            table.add_row(p, str(ret[p]))
```
wence-, line 755:
```suggestion
        for p in sorted(ret.keys()):
            table.add_row(p, str(ret[p]))
```
Or:
```
for key, value in sorted(ret.items(), key=operator.itemgetter(0)):
    table.add_row(key, str(value))
```
If you like

wence-: Aside: why do we need to explicit `str` the value, but not the key? With my proposed change to providing nested dictionaries for providing configuration settings of plugins would suggest one could format as nested tables.

dask_cuda/utils.py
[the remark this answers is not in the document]
wence-: Please prefer doing this in the presentation, rather than data-gathering, phase.

dask_cuda/tests/test_local_cuda_cluster.py
```diff
@@ -350,7 +355,10 @@ async def test_gpu_uuid():
 async def test_rmm_track_allocations():
     rmm = pytest.importorskip("rmm")
     async with LocalCUDACluster(
-        rmm_pool_size="2GB", asynchronous=True, rmm_track_allocations=True
+        rmm_pool_size="2GB",
+        asynchronous=True,
+        rmm_track_allocations=True,
+        device_memory_limit="30B",
```
wence-, line 361: Why this change here?

dask_cuda/utils.py
```diff
@@ -649,3 +653,105 @@ def get_gpu_uuid_from_index(device_index=0):
     pynvml.nvmlInit()
     handle = pynvml.nvmlDeviceGetHandleByIndex(device_index)
     return pynvml.nvmlDeviceGetUUID(handle).decode("utf-8")
+
+
+def get_worker_config(dask_worker):
```
wence-, line 658: Why is this function not private (`_get_worker_config`)? Or, conversely, why are the other functions not public?

dask_cuda/utils.py
```diff
@@ -649,3 +653,105 @@ def get_gpu_uuid_from_index(device_index=0):
...
+    ret = {}
+
+    # device and host memory configuration
+    for p in plugin_vals:
+        for v in dir(p):
+            # ignore hidden attrs as well as setup and CPUAffinity
+            if v.startswith("_") or v == "setup" or v == "cores":
+                pass
+            else:
+                ret[v] = getattr(p, v)
```
wence-, line 670: This assumes that all plugins have disjoint attributes.

```
plugin_config = {}
for p in plugin_vals:
    plugin_config[type(p).__name__] = {v: getattr(p, v) for v in dir(p) if not (v.startswith("_") or v in {"setup", "cores"})}
```
(Untested)

dask_cuda/utils.py
```diff
@@ -649,3 +653,105 @@ def get_gpu_uuid_from_index(device_index=0):
...
+
+        table = Table(title="Dask Cluster Configuration")
+        table.add_column("Parameter", justify="left", style="red")
+        table.add_column("Value", justify="left", style="green")
+
+        params = sorted(ret.keys())
+        for p in params:
+            table.add_row(p, str(ret[p]))
+        console = Console()
+        console.print(table)
```
wence-, line 757:
```suggestion
        Console().print(table)
```

wence- (changes requested):
Some nitpicks

dask_cuda/utils.py
```diff
@@ -649,3 +653,105 @@ def get_gpu_uuid_from_index(device_index=0):
...
+        except ImportError:
+            print(
+                "Please install rich `python -m pip install rich`"
+                / "to print a table of the current Dask Cluster Configuration"
+            )
+            sys.exit(1)
+
+        table = Table(title="Dask Cluster Configuration")
+        table.add_column("Parameter", justify="left", style="red")
+        table.add_column("Value", justify="left", style="green")
```
wence-, line 751: Should we really be opinionated about colours here? (I noticed in your screenshot that red on black was not very readable).

quasiben: Yes!  I would love direction on styling.  If you have a specific set you think works well I'm happy to change

dask_cuda/utils.py
quasiben: Rich uses [renderables](https://rich.readthedocs.io/en/latest/protocol.html#console-render) which are either strings or have implemented valid `__rich__` / `__rich_console__` methods:
https://github.com/Textualize/rich/blob/84e628655a2981ee90413ca3f35001ec3954161d/rich/protocol.py#L10-L16

dask_cuda/tests/test_local_cuda_cluster.py
quasiben: This was an accident while I was mucking about with config tests.

dask_cuda/utils.py
wence-: I'm not sure I do have a set that works. We could use the styles that rich defines, rather than specific colours e.g. "bold" (see them with `python -m rich.default_styles`) and then if someone has overridden the styling with a configuration file that will be obeyed.

dask_cuda/utils.py
quasiben: I'm following a pattern used in dask where the fully async function definition is private and the public function is both sync/async.  For example, we do something similar for [all_to_all](https://github.com/rapidsai/dask-cuda/blob/23460f6f7300202a2d3a25ffedf6884755fe6fbc/dask_cuda/utils.py#L444-L470)

dask_cuda/utils.py
quasiben: This is a lot clean and works nicely.  Thank you!

quasiben: Hmm, on second thought, I think it's better to key something like I had before but flatten the key name like:

> key = type(p).__name__ + '.' + v

This makes it easier to read without having to scan a nested dictionary within the table

quasiben:
Thanks for the review comments @wence-  .  I believe it's ready for another round.  As for styling, what do you think about bold bright_green ?

Screen Shot 2022-10-11 at 10 08 05 AM

dask_cuda/utils.py
wence-: > This makes it easier to read without having to scan a nested dictionary within the table

Or one could translate (in the rich output), like so:

```python
from functools import singledispatch
from rich.pretty import Pretty
from rich.table import Table
from operator import itemgetter

@singledispatch
def pretty(obj):
    return Pretty(obj)

@pretty.register(dict)
def pretty_dict(obj):
    t = Table()
    t.add_column("key")
    t.add_column("value")
    for k, v in sorted(obj.items(), key=itemgetter(0)):
        t.add_row(pretty(k), pretty(v))
    return t

...

data = get_config(...)
Console().print(pretty(data))
```

WDYT?

dask_cuda/cli/dask_cuda_worker.py
```diff
@@ -321,8 +333,14 @@ def main(
...
     pre_import,
     multiprocessing_method,
+    get_cluster_conf,
     **kwargs,
 ):
+    if get_cluster_conf:
+        client = Client(scheduler)
+        print_cluster_config(client)
+        sys.exit(0)
+
```
wence-, line 343: This needs to move further down (and respect `scheduler_file` and `security` arguments).

dask_cuda/utils.py
```diff
@@ -649,3 +653,116 @@ def get_gpu_uuid_from_index(device_index=0):
...
+        ret["ucx-transports"] = ucp.get_active_transports()
+
+    # comm timeouts
+    ret["distributed.comm.timeouts"] = dask.config.get("distributed.comm.timeouts")
+    # ret.update(plugin_config)
+
+    return ret
+
+
+async def _get_sched_config(client):
```
wence-, line 705: nit: consistent naming? `_get_scheduler_config`; and then perhaps, given we've called it `get_cluster_configuration`, `_get_scheduler_configuration` and `get_worker_configuration` ?

dask_cuda/utils.py
```diff
@@ -649,3 +653,116 @@ def get_gpu_uuid_from_index(device_index=0):
...
+def print_cluster_config(client):
+    """print current Dask cluster configuration"""
+    data = get_cluster_configuration(client)
+    formatted_byte_keys = [
+        "memory_limit",
+        "device-memory-limit",
+        "initial_pool_size",
+        "maximum_pool_size",
+    ]
+
```
wence-, line 742: Should we add the size of the cluster? (Number of workers/theads/gpus/etc...)

wence- (commented):
Minor

wence-:
> ```diff
> diff --git a/dask_cuda/cli/dask_cuda_worker.py b/dask_cuda/cli/dask_cuda_worker.py
> index 464d988..92b61e8 100755
> --- a/dask_cuda/cli/dask_cuda_worker.py
> +++ b/dask_cuda/cli/dask_cuda_worker.py
> @@ -336,11 +336,6 @@ def main(
>      get_cluster_conf,
>      **kwargs,
>  ):
> -    if get_cluster_conf:
> -        client = Client(scheduler)
> -        print_cluster_config(client)
> -        sys.exit(0)
> -
>      if multiprocessing_method == "forkserver":
>          import multiprocessing.forkserver as f
>  
> @@ -361,6 +356,14 @@ def main(
>              "unsupported one. Scheduler address: %s" % scheduler
>          )
>  
> +    if get_cluster_conf:
> +        if scheduler_file is not None:
> +            client = Client(scheduler_file=scheduler_file, security=security)
> +        else:
> +            client = Client(scheduler, security=security)
> +        print_cluster_config(client)
> +        sys.exit(0)
> +
>      if worker_class is not None:
>          worker_class = import_term(worker_class)
> ```

This hunk is needed irrespective of the proposed formatting changes.

quasiben:
@wence- I'll your suggestion a lot. Made a minor adjustment (need to default to `None` for `--scheduler-file`.  I'll push changes shortly with the other suggestions as well

dask_cuda/utils.py
quasiben: 👍  - this is really cool!

dask_cuda/utils.py
```diff
@@ -649,3 +653,116 @@ def get_gpu_uuid_from_index(device_index=0):
...
+            value = format_bytes(value)
+        table.add_row(key, str(value))
+    Console().print(table)
+
+
+def get_cluster_configuration(client):
+    data = client.sync(
+        _get_cluster_configuration, client=client, asynchronous=client.asynchronous
+    )
+    return data
```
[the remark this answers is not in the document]
quasiben: Actually, I think we need to go the other way and build an async print function.  This would allow of for usage of `print_cluster_config` inside a jupyter notebook with an async client

dask_cuda/utils.py
```diff
@@ -649,3 +654,144 @@ def get_gpu_uuid_from_index(device_index=0):
...
+
+
+def get_worker_config(dask_worker):
+    # assume homogenous cluster
+    plugin_vals = dask_worker.plugins.values()
+    ret = {}
+
+    # device and host memory configuration
+    for p in plugin_vals:
+        ret[type(p).__name__] = {
```
wence-, line 666: Looking again (sorry!), shall we indicate that these are plugins? `ret[f"[plugin] {type(p).__name__}"] = ...` ?

wence-:
Thanks @quasiben, I'm pretty happy with the output, and we can iterate separately if we need to afterwards. I think the outstanding queries to address here are @pentschev's comment about whether this really makes sense as an additional flag to `dask-cuda-worker` or whether it should be a separate commandline tool. We've now separated the data gathering from printing, so I think a separate commandline tool makes more sense.

This would also fix the likelihood of making the error that I did when trying to use it which was to try and launch a new worker and print the cluster configuration. This new option is really "connect to an existing cluster and dump information".

WDYT?

dask_cuda/utils.py
quasiben: I'm happy to include it and we can continue iterating.  I doubt most users deeply care about what is label a plugin but we (developers/anyone helping to debug) may find it useful.

quasiben:
I'll build a new CLI: `dask-config` this is also related to an issue I'm going to file today around configuring "profiles" for dask/RAPIDS

quasiben:
@wence- when you have some time can you give this PR another review ?

dask_cuda/cli/dask_config.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+pem_file_option_type = click.Path(exists=True, resolve_path=True)
+
+
+@click.command(context_settings=dict(ignore_unknown_options=True))
+@click.argument("scheduler", type=str, required=False)
+@click.argument(
+    "preload_argv", nargs=-1, type=click.UNPROCESSED, callback=validate_preload_argv
+)
```
wence-, line 24: These arguments don't seem to go anywhere? [...], oh I see, `validate_preload_argv` side-effectfully imports the specified module. Do we need this here?

dask_cuda/cli/dask_config.py
```diff
@@ -0,0 +1,101 @@
...
+            "unsupported one. Scheduler address: %s" % scheduler
+        )
+
+    if get_cluster_conf:
+        if scheduler_file is not None:
+            client = Client(scheduler_file=scheduler_file, security=security)
+        else:
+            client = Client(scheduler, security=security)
+        print_cluster_config(client)
+        sys.exit(0)
```
wence-, line 93: No need to `sys.exit` here since this is the end of the function.
```suggestion
```

dask_cuda/cli/dask_config.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def go():
+    main()
+
+
+if __name__ == "__main__":
+    go()
```
wence-, line 101:
```suggestion
if __name__ == "__main__":
    main()
```

?

dask_cuda/utils.py
```diff
@@ -649,3 +654,147 @@ def get_gpu_uuid_from_index(device_index=0):
...
+    scheme, loc = parse_address(dask_worker.scheduler.address)
+    ret["protocol"] = scheme
+    if scheme == "ucx":
+        import ucp
+
+        ret["ucx-transports"] = ucp.get_active_transports()
+
+    # comm timeouts
+    ret["distributed.comm.timeouts"] = dask.config.get("distributed.comm.timeouts")
+    # ret.update(plugin_config)
```
wence-, line 701:
```suggestion
```

dask_cuda/utils.py
```diff
@@ -649,3 +654,147 @@ def get_gpu_uuid_from_index(device_index=0):
...
+@singledispatch
+def pretty_print(obj, toplevel):
+    from rich.pretty import Pretty
+
+    return Pretty(obj)
+
+
+@pretty_print.register(str)
+def pretty_print_str(obj, toplevel):
+    return obj
```
wence-, line 747:
```suggestion
    from rich.markup import escape
    return escape(obj)
```

Moving the escaping to `str`s

dask_cuda/utils.py
```diff
@@ -649,3 +654,147 @@ def get_gpu_uuid_from_index(device_index=0):
...
+
+
+@pretty_print.register(dict)
+def pretty_print_dict(obj, toplevel):
+    from rich.markup import escape
```
wence-, line 752:
```suggestion
```

wence- (changes requested):
Very minor nits, but otherwise looks great! I like the output.

quasiben:
Thanks @wence- for the additional comments!

wence-:
Think you need a little bit more linting?

wence- (approved):
Thanks!

quasiben:
Thanks again for the comments and review @wence- @pentschev @madsbk

quasiben:
@gpucibot merge

PR #1025, opened 2022-10-26
Explicit-comms-shuffle: fine control of task scheduling

madsbk:
In shuffle, use `Client.submit()` to control where tasks are executed and release temporary dataframes ASAP.

#### Context
In the final step in explicit-comms shuffle, we call `getitem()` to extract the final dataframe partitions from the result of the local shuffle. In some cases, these `getitem()` tasks would not run on the worker that ran the local shuffle, which would result in extra communication and spilling.
We now use `submit(..., worker=...)` to make sure that the worker running the local shuffle also runs the `getitem()` task afterwards.

Is it possible to do this without the use of `submit()` to avoid the overhead of creating a `Future` for each dataframe partition?

### DIFF
--- a/dask_cuda/explicit_comms/dataframe/shuffle.py
+++ b/dask_cuda/explicit_comms/dataframe/shuffle.py
@@ -325,8 +325,11 @@ def shuffle(
                 rank_to_out_part_ids,
                 ignore_index,
             )
-    distributed.wait(list(result_futures.values()))
-    del df_groups
+    wait(list(result_futures.values()))
+
+    # Release dataframes from step (a)
+    for fut in df_groups:
+        fut.release()
 
     # Step (c): extract individual dataframe-partitions
     name = f"explicit-comms-shuffle-getitem-{tokenize(name)}"
--- a/dask_cuda/explicit_comms/dataframe/shuffle.py
+++ b/dask_cuda/explicit_comms/dataframe/shuffle.py
@@ -11,7 +11,6 @@ from toolz import first
 
 import dask
 import dask.dataframe
-import distributed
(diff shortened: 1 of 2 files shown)


dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -325,25 +324,39 @@ def shuffle(
...
-                meta = delayed(make_meta)(
-                    delayed(getitem)(result_futures[rank], i)
-                ).compute()
-    assert meta is not None
+    for rank, worker in enumerate(c.worker_addresses):
+        if rank in workers:
+            for i, part_id in enumerate(rank_to_out_part_ids[rank]):
+                dsk[(name, part_id)] = c.client.submit(
+                    getitem, result_futures[rank], i, workers=[worker]
+                )
```
wence-, line 344: I _thought_ that the values of a `dsk` graph always had to be tuples. But this returns a `Future`. Does that work fine?

madsbk: This should be fine: https://docs.dask.org/en/stable/spec.html#definitions

pentschev (approved):
Only added a minor code duplication reduction, apart from that LGTM. Thanks @madsbk !

wence- (approved):
My only comment was similar to @pentschev's; don't know a better way to force location of the compute

madsbk:
rerun tests

pentschev:
rerun tests

pentschev:
Thanks @madsbk !

pentschev:
@gpucibot merge

PR #1026, opened 2022-10-28
Switch pre-import not found test to sync definition

pentschev:
Even though pytest-async is installed, it has not executed async tests that aren't wrapped in gen_cluster. Using gen_cluster is the proper way for Dask testing, but in few exceptions we may not be able to use it, such as when cleanup must be avoided.

### DIFF
--- a/dask_cuda/tests/test_local_cuda_cluster.py
+++ b/dask_cuda/tests/test_local_cuda_cluster.py
@@ -1,3 +1,4 @@
+import asyncio
 import os
 import sys
 from unittest.mock import patch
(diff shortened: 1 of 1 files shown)


pentschev:
@madsbk this should fix CI failures seen in https://github.com/rapidsai/dask-cuda/pull/1025 .

madsbk (approved):
Awesome, thanks @pentschev

madsbk:
@gpucibot merge

pentschev:
@gpucibot merge

pentschev:
Thanks Mads, not sure why gpuCI didn't merge after your comment.

madsbk:
Did I loose my mojo? :)

pentschev:
> Did I lose my mojo? :)

Doesn't look like it, the important part (approving) was accepted. No idea what happened... 🤷

PR #1028, opened 2022-10-31
Fix recorded time in merge benchmark

wence-:
With the merge of #994 I accidentally changed the recorded time in the merge benchmark for the explicit-comms case. Since explicit-comms is eager, we must time the entire operation, not just `wait(result.persist())`.


madsbk (approved):
LGTM, thanks @wence-

wence-:
@gpucibot merge

PR #1030, opened 2022-11-02
Re-implement shuffle using staging

madsbk:
Introduce staging in explicit-comms. The idea is to "stage" the keys of the input on the workers so that a later explicit-comms task can access **and** free the data associated with the keys. 

Notice, explicit-comms and this new staging approach is still experimental. If or when it gets to a state where it provides a significant performance improvements over a range of workflows, the plan is to tighten up the API.


dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -254,94 +222,54 @@ def shuffle(
...
+    c.client.cancel(df)  # Notice, since `df` has been staged, nothing is freed here.
+
+    # Find the output partition IDs for each worker
+    ranks = sorted(rank_to_inkeys.keys())
+    div = npartitions // len(ranks)
+    rank_to_out_part_ids: Dict[int, List[int]] = {}  # rank -> [list of partition id]
+    for i, rank in enumerate(ranks):
+        rank_to_out_part_ids[rank] = list(range(div * i, div * (i + 1)))
+    for rank, i in zip(ranks, range(div * len(ranks), npartitions)):
+        rank_to_out_part_ids[rank].append(i)
```
wence-, line 245: Given that these are contiguous sequences, is it not more comms-efficient to just communicate the extents?

wence- (commented):
A few minor comments, but no need to act if you feel it doesn't really make sense to.

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -48,133 +45,104 @@ async def recv(
...
+                    stage.popitem()[1],  # pop dataframe in any order
+                    column_names,
+                    0,
+                    npartitions,
+                    npartitions,
+                    ignore_index,
+                    npartitions,
+                )
+            )
+        )
```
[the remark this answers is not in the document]
madsbk: I want to delete stage items ASAP,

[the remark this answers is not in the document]
wence-: Ah, ok.

dask_cuda/explicit_comms/dataframe/shuffle.py
madsbk: Agree, we should compute the IDs on-the-fly eventually. However, I like to keep it simple while we experiment with different approaches.

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -5,201 +5,239 @@
...
+    if isinstance(worker.data, ProxifyHostFile):
+        data = worker.data
+        return lambda x: data.manager.proxify(x)[0]
+    return lambda x: x  # no-op
+
+
+def compute_map_index(df: Any, column_names, npartitions) -> Series:
+    """Return a Series that maps each row `df` to a partition ID"""
+    if column_names[0] == "_partitions":
+        ind = df[column_names[0]]
```
wence-, line 76: This is somewhat magic, can you describe it in the docstring?

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -5,201 +5,239 @@
...
+            df[column_names] if column_names else df, index=False
+        )
+    typ = numpy.min_scalar_type(npartitions * 2)
+    return (ind % npartitions).astype(typ, copy=False)
+
+
+def single_shuffle_group(
+    df: DataFrame, column_names, npartitions, ignore_index
+) -> Dict[int, DataFrame]:
+    """Split dataframe such that each partition hashes to the same"""
```
wence-, line 88: Or maybe describe the magic `_partitions` key here.

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -5,201 +5,239 @@
...
+
+def multi_shuffle_group(
+    df_meta: DataFrame,
+    dfs: Dict[str, DataFrame],
+    column_names,
+    npartitions,
+    ignore_index,
+    proxify,
+) -> Dict[int, DataFrame]:
+    """Split multiple dataframes such that each partition hashes to the same
```
wence-, line 101: "hashes to the same" needs a "what" (as in what does it hash the same as).

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -5,201 +5,239 @@
...
-    in_parts: list of dict of dataframes
-        List of dataframe groups that need to be shuffled.
-    rank_to_out_part_ids: dict
-        dict that for each worker rank specifices a list of partition IDs that
-        worker should return. If the worker shouldn't return any partitions,
-        it is excluded from the dict.
+    df_meta: DataFrame
+        An empty dataframe matching the expected output
+    dfs: dict of dataframes
+        The dataframes to spilt given as a map of stage keys to dataframes
```
wence-, line 111:
```suggestion
        The dataframes to split given as a map of stage keys to dataframes
```

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -5,201 +5,239 @@
...
+    df: DataFrame, column_names, npartitions, ignore_index
+) -> Dict[int, DataFrame]:
+    """Split dataframe such that each partition hashes to the same"""
+    map_index = compute_map_index(df, column_names, npartitions)
+    return group_split_dispatch(df, map_index, npartitions, ignore_index=ignore_index)
+
+
+def multi_shuffle_group(
+    df_meta: DataFrame,
+    dfs: Dict[str, DataFrame],
```
wence-, line 95: So all of these dataframes here are somehow related to each other (because we can split them based on the same column names).

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -5,201 +5,239 @@
...
+    while dfs:
+        dfs_grouped.append(
+            proxify(
+                single_shuffle_group(
+                    # pop dataframe in any order, to free staged memory ASAP
+                    dfs.popitem()[1],
+                    column_names,
+                    npartitions,
+                    ignore_index,
+                )
```
wence-, line 139: So this does a round of communication (one for each input partition)?

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -5,201 +5,239 @@
...
+    # an empty dataframe is used.
+    ret: Dict[int, DataFrame] = {}
+    for i in range(npartitions):  # Iterate over all possible output partition IDs
+        t = [df_grouped[i] for df_grouped in dfs_grouped]
+        if len(t) == 1:
+            ret[i] = t[0]
+        elif len(t) > 1:
+            ret[i] = proxify(dd_concat(t, ignore_index=ignore_index))
+        else:
+            ret[i] = df_meta  # Empty dataframe
```
wence-, line 153: This seems like a slightly odd way to "transpose" a sparse data structure, I think?

I think also that the length checks are all independent of the number of partitions?
So what about

```
ret = defaultdict(list)
for group in dfs_grouped:
    for i, df in enumerate(group):
        ret[i].append(df)

def unpack(dfs):
    n = len(dfs)
    if n == 0:
        return df_meta
    elif n == 1:
        return dfs[0]
    else:
        return proxify(dd_concat(dfs, ignore_index=ignore_index))

ret = {i: unpack(dfs) for i, dfs in ret.items()}
```

Although as I write it, it is a bit longer, so perhaps meh? WDYT?

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -5,201 +5,239 @@
...
-        futures.append(recv(eps, in_nparts, out_parts_list))
-    if myrank in in_nparts:
-        futures.append(send(eps, rank_to_out_parts_list))
-    await asyncio.gather(*futures)
+    # `out_part_id_to_dataframe_list` that for each output partition maps
+    # a list of dataframes received.
+    out_part_id_to_dataframe_list: Dict[int, List[DataFrame]] = defaultdict(list)
+    await asyncio.gather(
+        recv(eps, myrank, rank_to_out_part_ids, out_part_id_to_dataframe_list, proxify),
+        send(eps, myrank, rank_to_out_part_ids, out_part_id_to_dataframe),
```
wence-, line 219: A change here is that you don't collect all the futures for all the tasks and gather them at once, but wrap them up in nested asyncio.gather calls. I don't expect this will really change anything substantive about the performance, but just to note.

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -5,201 +5,239 @@
...
-            ret.append(dfs[0])
+    assert rank_to_out_part_ids[myrank] == out_part_id_to_dataframe.keys()
+    # We can now add them to the output dataframes.
+    for out_part_id, dataframe in out_part_id_to_dataframe.items():
+        out_part_id_to_dataframe_list[out_part_id].append(dataframe)
+    del out_part_id_to_dataframe
+
+    # Finally, we concatenate the output dataframes into the final output partitions
+    ret: List[DataFrame] = []
+    for out_part_id, dataframe_list in out_part_id_to_dataframe_list.items():
```
wence-, line 232:
```suggestion
    for dataframe_list in out_part_id_to_dataframe_list.values():
```

You never needed the id.

So probably a list comprehension here would be simpler:

```
ret = [
proxify(dd_concat(dfs, ignore_index=ignore_index)) for dfs in out_part_id_to_dataframe_list.values()
]
```

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -241,118 +279,80 @@ def shuffle(
...
-        output_key = (name, input_key[1])
-        dsk[output_key] = (
-            shuffle_group,
-            input_key,
+    df_meta: DataFrame = df._meta
+
+    # Stage all keys of `df` on the workers and cancel them, which makes it possible
+    # for the shuffle to free memory as the partitions of `df` are consumed.
+    rank_to_inkeys = c.stage_keys(name=name, keys=df.__dask_keys__())
+    c.client.cancel(df)  # Notice, since `df` has been staged, nothing is freed here.
```
wence-, line 311: This I do not understand. We just copy some keys around here as far as I can tell (and don't persist anything), so I'm not sure what this even does.

wence- (commented):
Mostly just some queries for understanding, with some relatively minor issues.

TBH: This code is quite hard to follow at this point, though that may be my unfamiliarity with it, especially to do with what is executing where.

dask_cuda/explicit_comms/comms.py
```diff
@@ -260,3 +275,24 @@ def run(self, coroutine, *args, workers=None, lock_workers=False):
                     )
                 )
             return self.client.gather(ret)
+
+    def stage_keys(self, name: str, keys: Iterable[Hashable]) -> Dict[int, set]:
```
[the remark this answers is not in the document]
madsbk: Right, but added the `Hashable` as a reminder that all keys in Dask must be hashable.

dask_cuda/explicit_comms/comms.py
```diff
@@ -260,3 +275,24 @@ def run(self, coroutine, *args, workers=None, lock_workers=False):
...
+
+        The workers storing the keys and the associated data moves the data to
+        its state dict.
+
+        Parameters
+        ----------
+        name: str
+            Name for the staging
+        keys: iterable
+            The keys to stage
```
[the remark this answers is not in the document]
madsbk: Added more doc in 7ea9f6fb32c144ae922ad515bad8152c69888fd2

dask_cuda/explicit_comms/dataframe/shuffle.py
madsbk: Added docs in feac8d5f56887cf7eff5591d1ea3821a2038c24a

dask_cuda/explicit_comms/dataframe/shuffle.py
madsbk: Added docs in feac8d5f56887cf7eff5591d1ea3821a2038c24a

dask_cuda/explicit_comms/dataframe/shuffle.py
madsbk: Added docs in feac8d5f56887cf7eff5591d1ea3821a2038c24a

dask_cuda/explicit_comms/dataframe/shuffle.py
madsbk: Yes

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -5,201 +5,239 @@
...
+
+def compute_map_index(df: Any, column_names, npartitions) -> Series:
+    """Return a Series that maps each row `df` to a partition ID"""
+    if column_names[0] == "_partitions":
+        ind = df[column_names[0]]
+    else:
+        ind = hash_object_dispatch(
+            df[column_names] if column_names else df, index=False
+        )
+    typ = numpy.min_scalar_type(npartitions * 2)
```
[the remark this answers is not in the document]
madsbk: I don't know, it is from https://github.com/dask/dask/blob/bfe07f63967494a5cfcfafbb2a6b5271a6714dcc/dask/dataframe/shuffle.py#L937. I guess `numpy.min_scalar_type(npartitions)` should be enough?

dask_cuda/explicit_comms/dataframe/shuffle.py
madsbk: No communication here, this is creating the local shuffle groups

dask_cuda/explicit_comms/dataframe/shuffle.py
madsbk: No strong opinion here :)

dask_cuda/explicit_comms/dataframe/shuffle.py
madsbk: Good point, fixed in 6464eb6b9f3b07a11276729eb2a25d37149377f7

madsbk:
Thanks for the review @wence- ! 
I think I addressed all of your comments?

dask_cuda/explicit_comms/dataframe/shuffle.py
madsbk: Added more doc in 7ea9f6fb32c144ae922ad515bad8152c69888fd2

dask_cuda/explicit_comms/dataframe/shuffle.py
wence-: Let's leave it.

dask_cuda/explicit_comms/dataframe/shuffle.py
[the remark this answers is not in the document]
wence-: Comes from dask/dask#2032, but I think it doesn't matter. Any time `2` is going to take us over a boundary, we're either in a situation where the number of partitions is small, so 8bit -> 16bit integers doesn't hurt. Or large 16bit->32bit, where we are already in a bad place.

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -241,118 +292,81 @@ def shuffle(
...
-    for input_key in df.__dask_keys__():
-        output_key = (name, input_key[1])
-        dsk[output_key] = (
-            shuffle_group,
-            input_key,
+    df_meta: DataFrame = df._meta
+
+    # Stage all keys of `df` on the workers and cancel them, which makes it possible
+    # for the shuffle to free memory as the partitions of `df` are consumed.
+    # See CommsContext.stage_keys() for a description of staging.
```
wence-, line 323: Thanks, this makes more sense now!

madsbk: Yeah, sorry should have written that doc from the start!

madsbk:
rerun tests

madsbk:
@gpucibot merge

PR #1031, opened 2022-11-02
Add NVTX annotations for JIT spilling and unspilling

shwina:
Are these the right functions to annotate?

### DIFF
--- a/dask_cuda/proxy_object.py
+++ b/dask_cuda/proxy_object.py
@@ -35,6 +35,7 @@ except ImportError:
 
 from .disk_io import SpillToDiskFile
 from .is_device_object import is_device_object
+from .utils import nvtx_annotate
 
 if TYPE_CHECKING:
     from .proxify_host_file import ProxyManager
@@ -285,6 +286,7 @@ class ProxyDetail:
         self.serializer = header["serializer"]
         return self.obj
 
+    @nvtx_annotate("JIT_UNSPILL", color="green", domain="dask_cuda")
     def deserialize(self, maybe_evict: bool = True, nbytes=None):
         """Inplace deserialization of the proxied object
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -189,7 +189,8 @@ epub_exclude_files = ["search.html"]
 # -- Extension configuration -------------------------------------------------
 
 
-
 def setup(app):
     app.add_css_file("https://docs.rapids.ai/assets/css/custom.css")
-    app.add_js_file("https://docs.rapids.ai/assets/js/custom.js", loading_method="defer")
+    app.add_js_file(
+        "https://docs.rapids.ai/assets/js/custom.js", loading_method="defer"
+    )
--- a/dask_cuda/proxify_host_file.py
+++ b/dask_cuda/proxify_host_file.py
@@ -42,6 +42,7 @@ from .disk_io import SpillToDiskProperties, disk_read, disk_write
 from .get_device_memory_objects import DeviceMemoryId, get_device_memory_ids
 from .proxify_device_objects import proxify_device_objects, unproxify_device_objects
 from .proxy_object import ProxyObject
+from .utils import nvtx_annotate
 
 T = TypeVar("T")
 
@@ -393,6 +394,7 @@ class ProxyManager:
                 serializer(p)
         return freed_memory
 
+    @nvtx_annotate("JIT_SPILL_DEVICE_TO_HOST", color="red", domain="dask_cuda")
     def maybe_evict_from_device(self, extra_dev_mem=0) -> None:
         """Evict buffers until total memory usage is below device-memory-limit
 
@@ -409,6 +411,7 @@ class ProxyManager:
                 serializer=lambda p: p._pxy_serialize(serializers=("dask", "pickle")),
             )
 
+    @nvtx_annotate("JIT_SPILL_HOST_TO_DISK", color="red", domain="dask_cuda")
     def maybe_evict_from_host(self, extra_host_mem=0) -> None:
         """Evict buffers until total memory usage is below host-memory-limit
 


docs/source/conf.py
```diff
@@ -189,7 +189,8 @@
 # -- Extension configuration -------------------------------------------------
 
 
-
 def setup(app):
     app.add_css_file("https://docs.rapids.ai/assets/css/custom.css")
-    app.add_js_file("https://docs.rapids.ai/assets/js/custom.js", loading_method="defer")
+    app.add_js_file(
+        "https://docs.rapids.ai/assets/js/custom.js", loading_method="defer"
+    )
```
shwina, line 196: pre-commit did this...

galipremsagar (approved):

dask_cuda/proxy_object.py
```diff
@@ -285,6 +286,7 @@ def serialize(self, serializers: Iterable[str]) -> Tuple[dict, list]:
         self.serializer = header["serializer"]
         return self.obj
 
+    @nvtx_annotate("JIT_UNSPILL", color="green", domain="dask_cuda")
```
pentschev, line 289: I think you may want to annotate `serialize` just above as well, or is there a reason I don't immediately see why you intentionally skipped that?

pentschev (commented):
Looks good to me, tagging @madsbk just so he's aware and may know if we missed something important.

pentschev:
rerun tests

dask_cuda/proxy_object.py
shwina: I think the `JIT_SPILL_DEVICE_TO_HOST` and `JIT_SPILL_HOST_TO_DISK` annotations I added give more information than we would get by annotating `serialize`. Would also love for Mads to weigh in here.

pentschev:
Failing tests are unrelated, opened an issue in https://github.com/rapidsai/dask-cuda/issues/1032 .

pentschev:
rerun tests

dask_cuda/proxy_object.py
madsbk: I don't have a strong opinion but notice, `serialize` and `deserialize` are called both when spilling and doing TCP-communication whereas `maybe_evict_from_device` and `maybe_evict_from_host` are called only when spilling.

shwina: I agree that we don't want to annotate TCP serialization. If anyone would like to take this PR over and add the annotations in the right place, that'd be welcome!

wence-: I think that the annotations on `maybe_evict_from_device/host` cover the movement of data from device memory to host (and then disk) [the spilling part of the equation]. I think @shwina's approach here was to handle the movement of spilled data _back_ into device memory.

I haven't managed to follow the logic of who does the unspilling yet, but presumably it happens in `ProxifyHostFile.__getitem__` ?

madsbk: > I haven't managed to follow the logic of who does the unspilling yet, but presumably it happens in ProxifyHostFile.__getitem__ ?

It is done JIT by [`ProxyObject._pxy_deserialize`](https://github.com/rapidsai/dask-cuda/blob/branch-22.12/dask_cuda/proxy_object.py#L412) when the proxy object is accessed.

shwina: I've moved the annotation to `_pxy_deserialize` -- does this look right to you, @madsbk?

shwina:
@madsbk @pentschev @wence- -- sorry for letting this PR go stale. If you could take another look and review/approve, that would be much appreciated!

madsbk (approved):
LGTM, thanks @shwina.

ajschmidt8:
Removing `ops-codeowners` from the required reviews since it doesn't seem there are any file changes that we're responsible for. Feel free to add us back if necessary.

dask_cuda/proxy_object.py
```diff
@@ -411,6 +412,7 @@ def _pxy_serialize(
         # Invalidate the (possible) cached "device_memory_objects"
         self._pxy_cache.pop("device_memory_objects", None)
 
+    @nvtx_annotate("JIT_UNSPILL", color="green", domain="dask_cuda")
```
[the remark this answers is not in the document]
pentschev: Yes, this is expected because of serialization I guess. Are we able to register the source of a deserialized object? I'm sure I'm ignoring the complexity of this, but maybe that could be part of the serialization/spilling metadata?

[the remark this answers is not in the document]
pentschev: Also, I'm fine merging this as is, but it may be worth considering whether it's useful to have deserialization of transferred data (i.e., not unspilled data) annotated just as unspilled data, I think this has the potential to be very misleading and thus cause more confusion that help.

Maybe @shwina could comment on whether this satisfies immediate needs or if we really need to think of a more sophisticated approach to prevent any potential misleading annotations.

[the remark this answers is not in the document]
shwina: I agree; we ideally want to annotate _just_ unspilling (and not conflate deserialization and unspilling). Any help/direction for how we could separate those in the code would be much appreciated.

dask_cuda/proxy_object.py
shwina: Adding a `do-not-merge` until this conversation is resolved.

dask_cuda/proxy_object.py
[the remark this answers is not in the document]
shwina: Adding a `do-not-merge` until this conversation is resolved.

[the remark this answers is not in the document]
wence-: > Notice, this is also called when the object is accessed after it has been received using TCP.

If we're definitely _not_ using the TCP transports, then is this a correct measure of data that is actually being unspilled?

If so, we could also annotate the tcp deserialize call (so you can distinguish by the nvtx call stack those unspills that are due to tcp, adn those that are not).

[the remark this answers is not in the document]
madsbk: > If we're definitely not using the TCP transports, then is this a correct measure of data that is actually being unspilled?

Yes, `ProxyObject` are serialized without spilling them when they are communicated using UCX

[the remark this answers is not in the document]
shwina: Thanks both -- I'm afraid I still don't understand fully.

> Otherwise, we have to mark spilled ProxyObject that we are retrieving so we know not to annotate them when they are unspilled later.

At the time `ob_pxy_dask_deserialize` is called, how do we know if we're deserializing post-TCP, or if we're deserializing to unspill an object?

> we could also annotate the tcp deserialize call (so you can distinguish by the nvtx call stack those unspills that are due to tcp, adn those that are not).

This seems reasonable (we'd have to document somewhere how to interpret the annotations). Also, do you mean annotating `ProxyDetail.deserialize` specifically here?

PR #1035, opened 2022-11-04
Migrate as much as possible to `pyproject.toml`

jakirkham:
* Upgrade to versioneer 0.28 & run formatting tools on those files
* Move most of `setup.py` to `pyproject.toml`
* Move all of `setup.cfg` to `pyproject.toml`
* Move `pytest.ini` to `pyproject.toml`
* Move `.coveragerc` to `pyproject.toml`
* Move Flake8 settings to `.flake8`


pyproject.toml
```diff
@@ -0,0 +1,113 @@
+[build-system]
+build-backend = "setuptools.build_meta:__legacy__"
```
jakirkham, line 2: Currently need to use legacy since versioneer is vendored ( https://github.com/python-versioneer/python-versioneer/issues/249#issuecomment-1235845640 ). Though we could devendor versioneer

conda/recipes/dask-cuda/meta.yaml
```diff
@@ -16,11 +16,16 @@ source:
 build:
   number: {{ GIT_DESCRIBE_NUMBER }}
   string: py{{ py_version }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
+  script: {{ PYTHON }} -m pip install . -vv
+  entry_points:
+    - dask-cuda-worker = dask_cuda.cli.dask_cuda_worker:go
+    - dask-config = dask_cuda.cli.dask_config:go
```
wence-, line 22: Why do we have to specify this stuff twice (here and in pyproject.toml)? I guess it's because conda doesn't know anything about setuptools/pyproject? But given we're explicitly installing with pip should that not be fine?

.flake8
```diff
@@ -0,0 +1,14 @@
+[flake8]
+exclude = docs, __init__.py, versioneer.py
```
pentschev, line 2: Similar to UCX-Py, we have a bit of actual code in `__init__.py`. Would there be any downsides of not excluding it from flake8?

pyproject.toml
```diff
@@ -0,0 +1,113 @@
...
+skip = [
+    ".eggs",
+    ".git",
+    ".hg",
+    ".mypy_cache",
+    ".tox",
+    ".venv",
+    "build",
+    "dist",
+    "__init__.py",
```
pentschev, line 84: Just like with flake8, do we need to skip `__init__.py`?

wence- (commented):
Mostly queries for my edification

.flake8
jakirkham: This comes from here

https://github.com/rapidsai/dask-cuda/blob/2c99f5a0fdf59a7f89e182c4add9121e0b8dd48d/setup.cfg#L10

pyproject.toml
```diff
@@ -0,0 +1,113 @@
...
+[tool.coverage.run]
+disable_warnings = "include-ignored"
+include = "dask_cuda/*"
+omit = "dask_cuda/tests/*"
+
+[tool.isort]
+line_length = 88
+multi_line_output = 3
+include_trailing_comma = true
+force_grid_wrap = 0
```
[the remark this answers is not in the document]
jakirkham: These are just taken from here

https://github.com/rapidsai/dask-cuda/blob/2c99f5a0fdf59a7f89e182c4add9121e0b8dd48d/setup.cfg#L24

Since there is enough happening here, maybe we can handle other updates separately

pyproject.toml
jakirkham: This is just taken from this line

https://github.com/rapidsai/dask-cuda/blob/2c99f5a0fdf59a7f89e182c4add9121e0b8dd48d/setup.cfg#LL54

pyproject.toml
```diff
@@ -0,0 +1,113 @@
+[build-system]
+build-backend = "setuptools.build_meta:__legacy__"
+requires = [
+    "setuptools>=64.0.0",
+    "tomli  ; python_version < '3.11'",
```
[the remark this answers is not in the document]
jakirkham: Support for TOML is directly integrated into Python 3.11.0. So no need for this dependency there. Hence this is a backport included for support of older Python versions.

This is just the format for the file. Please see [this example]( https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use )

pyproject.toml
```diff
@@ -0,0 +1,113 @@
...
+    "ignore:make_current is deprecated:DeprecationWarning:",
+]
+
+[tool.setuptools]
+license-files = ["LICENSE"]
+package-dir = {"" = "."}
+packages = ["dask_cuda"]
+
+[tool.setuptools.dynamic]
+dependencies = {file = "requirements.txt"}
```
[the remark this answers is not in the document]
jakirkham: Agreed. Inlined and file deleted.

jakirkham:
Just a note, CI fails due to having an old version of `conda-build` (am able to reproduce the same locally). Submitted PR ( https://github.com/rapidsai/gpuci-build-environment/pull/278 ) to update `conda-build`

pyproject.toml
```diff
@@ -0,0 +1,131 @@
...
+[project.scripts]
+dask-cuda-worker = "dask_cuda.cli.dask_cuda_worker:go"
+dask-config = "dask_cuda.cli.dask_config:go"
+
+[project.optional-dependencies]
+docs = [
+    "numpydoc",
+    "sphinx",
+    "sphinx-click",
+    "sphinx-rtd-theme",
```
pentschev, line 47: Some of those had version requirements (see https://github.com/rapidsai/dask-cuda/pull/1035/files#diff-2c9c11d26b09b8afde329980309d967121543a456e4592c76886a20b5cf56c90), are those not required anymore?

jakirkham: Those seem to be pinned very tightly to specific versions (that are likely pretty outdated at this point). So no don't think we want those here. We could set lower bounds if we wanted them

pyproject.toml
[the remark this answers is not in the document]
wence-: Ah ok, this is [PEP-508](https://peps.python.org/pep-0508/), which is just inconsistent like this for presumably reasons.

pyproject.toml
```diff
@@ -0,0 +1,113 @@
...
+    "version",
+    "dependencies",
+]
+description = "Utilities for Dask and CUDA interactions"
+readme = { file = "README.md", content-type = "text/markdown" }
+authors = [
+    { name = "NVIDIA Corporation" },
+]
+license= { text = "Apache-2.0" }
+requires-python = ">=3.8"
```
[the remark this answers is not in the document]
wence-: So my confusion was that `build-system.requires` is not saying anything about the python version that works, but rather conditional dependencies. So this is OK.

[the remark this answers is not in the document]
jakirkham: Right when building other artifacts (like wheels for example), the assumption is one has built in an environment that already contains the intended Python version (and is using that Python for the build).

The `requires-python` is effectively intended for both build and run time.

pyproject.toml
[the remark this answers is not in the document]
jakirkham: Filed as issue ( https://github.com/rapidsai/dask-cuda/issues/1039 )

pyproject.toml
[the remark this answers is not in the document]
jakirkham: The different types of quoting is just a Python trick to avoid needing to escape the quotes with `\`. Python views these two strings `"abc"` and `'abc'` as equivalent.

docs/requirements.txt
```diff
@@ -1,4 +1,5 @@
 dask_cuda
 numpydoc==1.1.0
 sphinx_click==2.7.1
-sphinx_rtd_theme==0.5.1
\ No newline at end of file
+sphinx_rtd_theme==0.5.1
+tomli
```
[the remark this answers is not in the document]
jakirkham: It was needed because the docs also did a build. Or at least that was true when the docs here were built on ReadTheDocs. However docs are now built by RAPIDS.

Plus we can do the same simplification that you noted elsewhere of dropping the requirements file and including these in `pyproject.toml`. So have since made that change.

jakirkham:
rerun tests

jakirkham:
rerun tests

jakirkham:
rerun tests

pentschev:
CI issues are unrelated, I opened https://github.com/rapidsai/dask-cuda/pull/1041 to attempt to temporarily unblock CI.

pyproject.toml
```diff
@@ -0,0 +1,142 @@
...
+dependencies = [
+    "dask >=2022.9.2",
+    "distributed >=2022.9.2",
+    "pynvml >=11.0.0",
+    "numpy >=1.18.0",
```
jakirkham, line 25: Bumped to [align with Dask's `numpy` requirement]( https://github.com/dask/dask/blob/2022.11.0/setup.py#L15 )

pyproject.toml
```diff
@@ -0,0 +1,142 @@
...
+    "numba >=0.54",
+    "pandas >=1.0",
```
jakirkham, line 27: Added as `pandas` is used here:

https://github.com/rapidsai/dask-cuda/blob/b3ed9029a1ad02a61eb7fbd899a5a6826bb5cfac/dask_cuda/proxy_object.py#L11

Also [aligned with Dask's `pandas` requirement]( https://github.com/dask/dask/blob/2022.11.0/setup.py#L17 )

pyproject.toml
```diff
@@ -0,0 +1,142 @@
...
+    "zict >=0.1.3",
```
jakirkham, line 28: Added as `zict` is used here:

https://github.com/rapidsai/dask-cuda/blob/b3ed9029a1ad02a61eb7fbd899a5a6826bb5cfac/dask_cuda/device_host_file.py#L8

Also [aligned with Distributed's `zict` requirement]( https://github.com/dask/distributed/blob/2022.11.0/requirements.txt#L14 )

pyproject.toml
```diff
@@ -0,0 +1,142 @@
...
+[project.optional-dependencies]
+docs = [
+    "numpydoc",
+    "sphinx",
+    "sphinx-click",
+    "sphinx-rtd-theme",
+]
+test = [
+    "pytest",
+]
```
jakirkham, line 53: Added an optional test requirement for clarity. Only saw `pytest` being used.

Not sure if `pytest-asyncio` is still needed ( or desired https://github.com/rapidsai/dask-cuda/issues/1044 )

Can add anything else needed here

pyproject.toml
jakirkham: Added those as lower bounds

The challenge with having them pinned that tightly generally is they will fall out of date, have unfixed bugs, and be incompatible with newer Python & Sphinx versions (the latter also not being pinned currently)

pentschev (approved):
Latest changes look good to me. Thanks @jakirkham !

raydouglass (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks all! 🙏 Happy to follow up on anything else separately

PR #1036, opened 2022-11-08
Fix version constraint

wence-:
Should apply custom __new__ if Python < 3.9.


pentschev (approved):
Oh yes, sorry, I should have catched that. Thanks for fixing this @wence- !

pentschev:
@gpucibot merge

PR #1046, opened 2022-11-16
[REVIEW] Pin `dask` and `distributed` for release

galipremsagar:
This PR pins `dask` and `distributed` to `2022.11.1` for `22.12` release.

xref: https://github.com/rapidsai/cudf/pull/12165


galipremsagar:
cc: @pentschev for review

pentschev:
I think we should wait until codefreeze, burndown is starting only tomorrow. Any reasons why we should do it this early?

galipremsagar:
> I think we should wait until codefreeze, burndown is starting only tomorrow. Any reasons why we should do it this early?

We can wait if you would want us to across all repos. cudf code-freeze begins tomorrow, hence opening the PRs for all repos.

pentschev:
I guess it makes sense to pin it down once we are somewhat confident this is the version we want. Since code freeze for ["others" (including Dask) starts December 1st](https://docs.rapids.ai/maintainers), I'm imagining there will be another Dask release until then. I have no strong objections on doing this immediately, but if we do this is effectively a code-freeze regarding Dask, which is what we wanted to avoid.

raydouglass (approved):

pentschev (changes requested):
Momentarily blocking this to raise a discussion.

pentschev:
Per our discussion, we decided we should pin during the second code freeze, for the upcoming release that is December 1st. I added the "blocked" label for this PR, which we can on that date update and merge. Thanks so much @galipremsagar for your effort here, it's really appreciated! 😄

galipremsagar:
pinging @pentschev for review

pentschev (approved):
LGTM, thanks @galipremsagar !

pentschev:
@gpucibot merge

PR #1049, opened 2022-11-17
[skip gpuci] Build dask-cuda internally to use for RAPIDS CI

sevagh:
During the pip wheel CI for RAPIDS projects (e.g. cudf/cugraph/cuml), we need to be able to install dask-cuda.

Dask-cuda is released to PyPI.org on a specific release schedule. 

This PR allows dask-cuda wheels to get built using the RAPIDS pip wheel workflows, published to the RAPIDS nightly index. That way, RAPIDS CI can have access to an installable `dask-cuda` pip wheel before the PyPI release.

This shouldn't affect external users, since our internal nightly index is not accessible to the public (yet).

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -42,6 +42,6 @@ jobs:
       # leave this empty to remove the default '-cu11' suffix
       python-package-cuda-suffix: ""
 
-      test-extras:  "test"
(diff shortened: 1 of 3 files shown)


sevagh:
Closing this since @vyasr suggested we install from source instead of pre-publishing a wheel.

PR #1053, opened 2022-11-23
Ensure linting checks for whole repo in CI

pentschev:
It often happens that `pre-commit` finds improperly formatted files that are not in the `dask_cuda` directory, therefore it is sensible to check the entire repo.

### DIFF
--- a/ci/checks/style.sh
+++ b/ci/checks/style.sh
@@ -12,50 +12,5 @@ PATH=/opt/conda/bin:$PATH
 . /opt/conda/etc/profile.d/conda.sh
 conda activate rapids
 
-# Print versions
-echo -e "\nVersions:"
-black --version
-echo "isort, `isort --vn`"
-echo "flake8, `flake8 --version`"
-
-# Run isort and get results/return code
-ISORT=`isort --check-only . 2>&1`
-ISORT_RETVAL=$?
-
-# Run black and get results/return code
-BLACK=`black --check . 2>&1`
-BLACK_RETVAL=$?
-
-# Run flake8 and get results/return code
-FLAKE=`flake8 . 2>&1`
-FLAKE_RETVAL=$?
-
-# Output results if failure otherwise show pass
-if [ "$ISORT_RETVAL" != "0" ]; then
-  echo -e "\n\n>>>> FAILED: isort style check; begin output\n\n"
-  echo -e "$ISORT"
-  echo -e "\n\n>>>> FAILED: isort style check; end output\n\n"
-else
-  echo -e "\n\n>>>> PASSED: isort style check\n\n"
-fi
-
-if [ "$BLACK_RETVAL" != "0" ]; then
-  echo -e "\n\n>>>> FAILED: black style check; begin output\n\n"
-  echo -e "$BLACK"
-  echo -e "\n\n>>>> FAILED: black style check; end output\n\n"
-else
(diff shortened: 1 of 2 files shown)


ci/checks/style.sh
```diff
@@ -19,15 +19,15 @@ echo "isort, `isort --vn`"
...
-BLACK=`black --check dask_cuda 2>&1`
+BLACK=`black --check . 2>&1`
 BLACK_RETVAL=$?
 
 # Run flake8 and get results/return code
-FLAKE=`flake8 dask_cuda 2>&1`
+FLAKE=`flake8 . 2>&1`
 FLAKE_RETVAL=$?
 
 # Output results if failure otherwise show pass
```
wence-, line 33: If we're doing this, we should just migrate to using precommit in CI by just running

```
pre-commit run --hook-stage manual --all-files
```

And remove this "by hand" handling.

pentschev: This would only apply to the local CI clone I presume, which won't actually do anything to the changes in the PR. I would not oppose to something automated, but I believe this has to be discussed internally as to why it isn't done that way today and what would be the correct approach to apply this.

wence-: I think (although I am a bit unsure) that this is mostly historical reasons.

Right now, the status quo is:

If someone has run `pre-commit install --install-hooks` in their local clone, then pre-commit will run when they try and commit something. This will obey the normal configuration mechanisms (`pyproject.toml` etc...).

When they then push, CI runs this file, and complains if things are bad.

So at present we have multiple sources of truth for:

1. What it means to pass style checks (is it pre-commit or is it this file?)
2. What versions to use (pre-commit specifies things, this file presumably picks up from the CI environment).

At least in cudf, we've moved to a situation where pre-commit (and its configuration) is the single source of truth. But I guess that if we want to do that here we should agree somehow to do so.

pentschev: Oh, I think what you mean is different than what I understood from your initial comment. What you're suggesting is running `pre-commit install --install-hooks` (or something equivalent) as the check itself as opposed to isort/black/flake8 individually (which is what you meant with "by hand"), and not that CI or merge procedure will auto-format when the PR is merged and thus we could eliminate style-checking. I agree this is a better solution and we can implement it that way here, I'll push changes shortly.

madsbk (approved):
LGTM, thanks @pentschev 
BTW, at some point we should also run `mypy`

pentschev:
> LGTM, thanks @pentschev BTW, at some point we should also run `mypy`

Thanks @madsbk . I think for `mypy` we would need to add type hints to everything, I'm not necessarily opposed to that, but it is going to be quite some work, and if anyone has bandwidth to do that it would be great. The only thing I don't like about `mypy` is how slow that is, the experience I have with Distributed is that it takes some 30+ seconds to `git commit` anything, which is really annoying.

pentschev:
@wence- anything else here or do you think we're good to merge?

wence- (approved):

wence-:
@gpucibot merge

wence-:
> @gpucibot merge

Ah no, it has no power here...

pentschev:
Thanks @wence- , I've asked for ops approval, should get merged once that it's approved.

AyodeAwe (approved):

PR #1055, opened 2022-11-28
[REVIEW] Fix `parse_memory_limit` function call

galipremsagar:
This PR addresses a breaking change that was made in upstream distributed: https://github.com/dask/distributed/pull/5669/

Error due to above change:
```
>       self.memory_limit = parse_memory_limit(
            memory_limit=memory_limit, nthreads=1, total_cores=n_workers
        )
E       TypeError: parse_memory_limit() missing 1 required keyword-only argument: 'logger'

```
Introduced a version-based function call because we haven't yet finalized the `dask` pinning for `22.12`, though as of now `2022.11.1` seems to be the most likely release pinning. Since the dask dev `2022.11.1` packages now contain this breaking change, this check will also help unblock any `22.12` CI jobs that are fetching `2022.11.1` dask dev packages.

Once `dask` is pinned to a specific version, this version-based check can be removed.

### DIFF
--- a/dask_cuda/cuda_worker.py
+++ b/dask_cuda/cuda_worker.py
@@ -16,9 +16,9 @@ from distributed.proctitle import (
     enable_proctitle_on_children,
     enable_proctitle_on_current,
 )
+from distributed.utils import has_arg
 from distributed.worker_memory import parse_memory_limit
 
-from ._compat import DISTRIBUTED_GT_2022_11_1
 from .device_host_file import DeviceHostFile
 from .initialize import initialize
 from .proxify_host_file import ProxifyHostFile
@@ -86,7 +86,8 @@ class CUDAWorker(Server):
             raise ValueError("nthreads must be higher than 0.")
 
         # Set nthreads=1 when parsing mem_limit since it only depends on nprocs
-        if DISTRIBUTED_GT_2022_11_1:
+        if has_arg(parse_memory_limit, "logger"):
+            # TODO: Remove has_arg check after 2022.11.1 support is dropped
             logger = logging.getLogger(__name__)
             memory_limit = parse_memory_limit(
                 memory_limit=memory_limit, nthreads=1, total_cores=nprocs, logger=logger
--- a/dask_cuda/cuda_worker.py
+++ b/dask_cuda/cuda_worker.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, division, print_function
 
 import asyncio
 import atexit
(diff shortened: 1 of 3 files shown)


dask_cuda/cuda_worker.py
```diff
@@ -84,9 +86,15 @@ def __init__(
...
-        )
+        if DISTRIBUTED_GT_2022_11_1:
+            logger = logging.getLogger(__name__)
+            memory_limit = parse_memory_limit(
+                memory_limit=memory_limit, nthreads=1, total_cores=nprocs, logger=logger
+            )
+        else:
+            memory_limit = parse_memory_limit(
+                memory_limit=memory_limit, nthreads=1, total_cores=nprocs
+            )
```
pentschev, line 97: I'd suggest we keep things simple, we don't want to keep this in the long-run, since we don't provide backwards compatibility anyway. For that reason I suggest the following instead:

```python
from distributed.utils import has_arg

if has_arg(parse_memory_limit, "logger"):
    logger = logging.getLogger(__name__)
    memory_limit = ...
else:
    memory_limit = ...
```

We can then drop the version compatibility code. Plus, I'd add a `TODO` comment or an issue to remove that in 23.02, when we'll definitely not support the old `2022.11.1` behavior.

pentschev (changes requested):
Thanks for finding this @galipremsagar , I added a request to make the code a bit simpler, and that we probably want to drop it later.

dask_cuda/cuda_worker.py
galipremsagar: TIL that this utility exists.. Done, this is quite simpler too Peter.

wence- (approved):
Thanks!

pentschev (approved):
Looks great, thanks @galipremsagar !

pentschev:
@gpucibot merge

galipremsagar:
rerun tests

galipremsagar:
rerun tests

galipremsagar:
@madsbk Spilling tests seem to be timing out on `centos7` jobs: https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/dask-cuda/job/prb/job/dask-cuda-gpu-test/CUDA=11.2,GPU_LABEL=driver-450,LINUX_VER=centos7,PYTHON=3.8/1032/testReport/junit/dask_cuda.tests/test_proxify_host_file/test_spill_on_demand/

PR #1061, opened 2022-12-05
Fix owner check when the owner is a cupy array

wence-:
A cupy array can't be used in a boolean setting (it is neither truthy nor falsy because at heart it's intuitionist) so we need to explicitly check that the owner is None.


wence-:
What's the right way to write a test for this?

madsbk (approved):
Thanks @wence- !

madsbk:
> What's the right way to write a test for this?

Test with  a cupy array in `test_cudf_get_device_memory_objects`

madsbk:
@gpucibot merge

PR #1062, opened 2022-12-05
Add GitHub Actions Workflows

bdice:
## Description

This PR adds GitHub Actions workflows to `dask-cuda`.

### Task list

Coverage required for this PR:
- [x] Python tests
- [x] Codecov
- [x] Style checks

Future work required:
- [Deploy sdist/wheels to PyPI](https://github.com/rapidsai/dask-cuda/blob/d6ff68daae638c30e1e2e25f2fb91ecc1ee8f6ea/ci/cpu/build.sh#L98)


bdice:
Thanks @pentschev for helping with the libnuma questions and @ajschmidt8 for updating the base images. Tests appear to be passing now. I’m going to open this for review but it *should not be merged until after the 22.12 release*, when we will migrate most repos to GitHub Actions. Can someone with privileges mark this DO NOT MERGE?

.github/workflows/build.yaml
```diff
@@ -0,0 +1,93 @@
...
+    runs-on:
+      - self-hosted
+      - linux
+      - amd64
+      - cpu4
+    container:
+      image: rapidsai/ci:latest
+      env:
+        RAPIDS_BUILD_TYPE: 'pull-request'
+        NVIDIA_VISIBLE_DEVICES: ${{ env.NVIDIA_VISIBLE_DEVICES }}
```
ajschmidt8, line 60:
```suggestion
```

This is only necessary when we're running on GPU machines. Otherwise it's extraneous.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,80 @@
...
+    runs-on:
+      - self-hosted
+      - linux
+      - amd64
+      - cpu4
+    container:
+      image: rapidsai/ci:latest
+      env:
+        RAPIDS_BUILD_TYPE: 'pull-request'
+        NVIDIA_VISIBLE_DEVICES: ${{ env.NVIDIA_VISIBLE_DEVICES }}
```
ajschmidt8, line 47:
```suggestion
```

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,80 @@
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Standardize repository information
+        run: |
+          echo "RAPIDS_REPOSITORY=${{ github.repository }}" >> "${GITHUB_ENV}"
+          echo "RAPIDS_SHA=$(git rev-parse HEAD)" >> "${GITHUB_ENV}"
+          echo "RAPIDS_REF_NAME=${{ github.ref_name }}" >> "${GITHUB_ENV}"
+          echo "RAPIDS_BUILD_TYPE=pull-request" >> "${GITHUB_ENV}"
```
ajschmidt8, line 60: I'm probably going to create a composite action with inputs to consolidate this logic into a single location. I don't like that it's copy/pasted in this repo and others at the moment.

ci/release/update-version.sh
```diff
@@ -33,3 +33,7 @@ function sed_runner() {
 
 # Update UCX-Py version
 sed_runner "s/export UCXPY_VERSION=.*/export UCXPY_VERSION="${NEXT_UCXPY_VERSION}"/g" ci/gpu/build.sh
+
+# Bump cudf and dask-cudf testing dependencies
+sed_runner "s/cudf=${CURRENT_SHORT_TAG}/cudf=${NEXT_SHORT_TAG}/g" dependencies.yaml
+sed_runner "s/dask-cudf=${CURRENT_SHORT_TAG}/dask-cudf=${NEXT_SHORT_TAG}/g" dependencies.yaml
```
ajschmidt8, line 39:
```suggestion
sed_runner "s/cudf=.*/cudf=${NEXT_SHORT_TAG}/g" dependencies.yaml
sed_runner "s/dask-cudf=.*/dask-cudf=${NEXT_SHORT_TAG}/g" dependencies.yaml
```

this is untested, but I usually prefer the leave the source pattern as a wildcard like this so that it doesn't break if things become out of sync (which happens occasionally).

This also matches the `UCXPY_VERSION` line above it too

.github/workflows/pr.yaml
ajschmidt8: I won't let this hold up the PR. We can update it in a later PR.

ajschmidt8: turns out these vars aren't needed. removed in 8a07c21 (details in the commit body)

.github/workflows/pr.yaml
ajschmidt8: fixed in 8e4cfbb.

.github/workflows/build.yaml
ajschmidt8: fixed in 8e4cfbb.

ci/release/update-version.sh
ajschmidt8: fixed in 7c2b129.

ajschmidt8:
I don't think the latest `pytest` failures are related to the changes in our PR.

They also seem to be appearing in https://github.com/rapidsai/dask-cuda/pull/1069.

pentschev:
> I don't think the latest `pytest` failures are related to the changes in our PR.
> 
> They also seem to be appearing in #1069.

They're not related, I opened #1070 to track that.

ci/test_python.sh
```diff
@@ -0,0 +1,88 @@
...
+  --runs 1 \
+  --backend explicit-comms
+exitcode=$?
+
+if (( ${exitcode} != 0 )); then
+    SUITEERROR=${exitcode}
+    echo "FAILED: Local benchmark with explicit comms"
+fi
+
+exit ${SUITEERROR}
```
pentschev, line 88: This will only capture the latest error, which may or may not be fine depending on how this is handled outside. Since we can't return multiple error codes on a single script, I think it might make sense to:

1. Have one script per executed test (in this particular case we would need 3, the first `pytest` and two benchmarks); or
2. Logging the error code for each failing test and returning a generic error code that the caller of this script will be able to parse (e.g., `0` for "success" and `1` for "error(s) occurred").

dependencies.yaml
```diff
@@ -0,0 +1,103 @@
...
+      - py_version
+      - test_python
+  checks:
+    output: none
+    includes:
+      - develop
+      - py_version
+channels:
+  - rapidsai
+  - rapidsai-nightly
```
pentschev, line 25: Do we really want both `rapidsai` and `rapidsai-nightly` here?

pentschev (commented):
Thanks @bdice and @ajschmidt8 for the hard work you've put on this. I've added a few comments.

conda/recipes/dask-cuda/meta.yaml
```diff
@@ -1,12 +1,12 @@
...
+# Copyright (c) 2019-2022, NVIDIA CORPORATION.
 
 # Usage:
 #   conda build -c conda-forge .
 {% set data = load_file_data("pyproject.toml") %}
 
 {% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev').lstrip('v') + environ.get('VERSION_SUFFIX', '') %}
 {% set number = environ.get('GIT_DESCRIBE_NUMBER', 0) %}
-{% set py_version = environ.get('CONDA_PY', 36) %}
+{% set py_version = environ['CONDA_PY'] %}
```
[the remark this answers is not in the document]
bdice: This is set by conda, not by RAPIDS. This should be safe -- and if it's not, we want to have a hard failure.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,63 @@
+name: build
+
+on:
+  push:
+    branches:
+      - "branch-*"
+    tags:
+      - v[0-9][0-9].[0-9][0-9].[0-9][0-9]
```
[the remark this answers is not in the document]
bdice: I'll leave this call to @ajschmidt8  but I think that at this point, we'd prefer consistency with all other RAPIDS repos, even if this is subpar. I am also not 100% sure that a general regex is appropriate here. I think it has to adhere to GitHub Actions filter patterns, which might only support a subset of regular expression syntax (this cheatsheet is fairly limited). https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

ci/test_python.sh
bdice: We have plans to revisit and simplify some of this logic in follow-up PRs, but ultimately I think we only care to distinguish zero (success) from non-zero (failure) exit codes for the purposes of GitHub Actions calling this script.

dependencies.yaml
bdice: Yes. The tests require cudf (among other RAPIDS packages), and the pinning of `cudf=23.02` requires `rapidsai-nightly`. For any dependencies on other RAPIDS packages, we want to test against nightlies so we are aware of breakage before release time.

ajschmidt8:
rerun tests

.github/workflows/build.yaml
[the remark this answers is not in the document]
ajschmidt8: Bradley is correct. GH Actions don't support full regex syntax unfortunately.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,63 @@
...
+      run:
+        shell: bash
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Build wheel
+        run: ci/build_python_pypi.sh
+      - name: Publish distribution 📦 to PyPI
+        if: inputs.build_type == 'branch'
```
ajschmidt8, line 60:
```suggestion
        if: inputs.build_type == 'nightly'
```

For parity with the Jenkins setup, we should only publish this once a day (at night).

We'll also need to add this repo to the RAPIDS nightly pipeline below

- https://github.com/rapidsai/actions/blob/main/.github/workflows/nightly-pipeline.yaml

ajschmidt8 (approved):

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - checks
+      - conda-python-build
+      - conda-python-tests
+      - wheel-build
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/pr-builder.yaml@main
+  checks:
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/checks.yaml@main
+  conda-python-build:
```
ajschmidt8, line 24:
```suggestion
  conda-python-build:
    needs: checks
```

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,47 @@
...
+    uses: rapidsai/shared-action-workflows/.github/workflows/conda-python-matrix-build.yaml@main
+    with:
+      build_type: pull-request
+  conda-python-tests:
+    needs: conda-python-build
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/conda-python-tests.yaml@main
+    with:
+      build_type: pull-request
+  wheel-build:
```
ajschmidt8, line 35:
```suggestion
  wheel-build:
      needs: checks
```

PR #1065, opened 2022-12-08
Add strings udf to build setup

quasiben:
We recently ran into trouble with strings-udf inadvertently calling cuInit: https://github.com/rapidsai/cudf/issues/12340 .  By adding `strings-udf` to the build env for dask-cuda we can properly test in CI and catch any more errors around cuda context creation/cuInit calls

cc @pentschev @wence-

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -65,6 +65,7 @@ conda list --show-channel-urls
 # Installing cucim in order to test GDS spilling
 gpuci_mamba_retry install "cudf=${MINOR_VERSION}" \
               "dask-cudf=${MINOR_VERSION}" \
+              "strings_udf=${MINOR_VERSION}" \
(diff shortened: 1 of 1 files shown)


wence-:
This looks fine, but I note that CI passes with this in. So it's not the obvious canary we might want? How do we observe the failure programmatically so that we'll spot when something bad creeps in in an upstream dep?

quasiben:
@wence- you're right.  We don't do multi-gpu testing in CI so I'll need to think of something else

wence-:
I think I can write a test in cudf that (ab)uses LD_PRELOAD, let me try...

ajschmidt8 (approved):

quasiben:
closing in favor of https://github.com/rapidsai/cudf/pull/12361

PR #1068, opened 2022-12-21
Shuffle by partition to reduce memory usage significantly

madsbk:
In order to reduce peak memory usage, this PR implements _rounds_ in explicit-comms shuffle.
The idea is that each worker handles a number of dataframe partitions in each round instead of doing everything at once.
The number of partitions handled in each round can be controlled by setting `DASK_XCOMM_BATCHSIZE` or directly when calling `shuffle()`. 

By default, each worker handles one partition per round. 
Set `DASK_XCOMM_BATCHSIZE=-1`, to handle all partitions in a single round (the previous behavior).

### DIFF
--- a/dask_cuda/explicit_comms/dataframe/shuffle.py
+++ b/dask_cuda/explicit_comms/dataframe/shuffle.py
@@ -4,10 +4,9 @@ import asyncio
 import functools
 import inspect
 from collections import defaultdict
+from math import ceil
 from operator import getitem
-from typing import Any, Callable, Dict, List, Optional, Set, TypeVar
-
-import numpy
+from typing import Any, Dict, List, Optional, Protocol, Set, TypeVar
 
 import dask
 import dask.dataframe
@@ -23,9 +22,26 @@ from .. import comms
 T = TypeVar("T")
 
 
+class Proxify(Protocol):
+    """Proxify type hint"""
+
+    def __call__(self, obj: T) -> T:
+        ...
+
+
+def get_proxify(worker: Worker) -> Proxify:
+    """Get function to proxify objects"""
+    from dask_cuda.proxify_host_file import ProxifyHostFile
+
+    if isinstance(worker.data, ProxifyHostFile):
+        data = worker.data
+        return lambda x: data.manager.proxify(x)[0]
+    return lambda x: x  # no-op
+
+
 async def send(
     eps,
-    myrank,
+    myrank: int,
     rank_to_out_part_ids: Dict[int, Set[int]],
     out_part_id_to_dataframe: Dict[int, DataFrame],
 ) -> None:
@@ -43,10 +59,10 @@ async def send(
 
 async def recv(
     eps,
-    myrank,
+    myrank: int,
     rank_to_out_part_ids: Dict[int, Set[int]],
     out_part_id_to_dataframe_list: Dict[int, List[DataFrame]],
-    proxify,
+    proxify: Proxify,
 ) -> None:
     """Notice, received items are appended to `out_parts_list`"""
 
@@ -60,17 +76,9 @@ async def recv(
     )
 
 
-def get_proxify(worker: Worker) -> Callable[[T], T]:
-    """Get function to proxify objects"""
-    from dask_cuda.proxify_host_file import ProxifyHostFile
-
-    if isinstance(worker.data, ProxifyHostFile):
-        data = worker.data
-        return lambda x: data.manager.proxify(x)[0]
-    return lambda x: x  # no-op
-
-
-def compute_map_index(df: Any, column_names, npartitions) -> Series:
+def compute_map_index(
+    df: DataFrame, column_names: List[str], npartitions: int
+) -> Series:
     """Return a Series that maps each row `df` to a partition ID
 
     The partitions are determined by hashing the columns given by column_names
@@ -79,17 +87,17 @@ def compute_map_index(df: Any, column_names, npartitions) -> Series:
 
     Parameters
     ----------
-    df: DataFrame
-    column_names: list of strings
+    df
+        The dataframe.
+    column_names
         List of column names on which we want to split.
-    npartitions: int or None
+    npartitions
         The desired number of output partitions.
 
     Returns
     -------
-    out: Dict[int, DataFrame]
-        A dictionary mapping integers in {0..k} to dataframes such that the
-        hash values of `df[col]` are well partitioned.
+    Series
+        Series that maps each row `df` to a partition ID
     """
 
     if column_names[0] == "_partitions":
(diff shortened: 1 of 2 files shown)


dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -324,6 +414,15 @@ def shuffle(
     rank_to_inkeys = c.stage_keys(name=name, keys=df.__dask_keys__())
     c.client.cancel(df)
 
+    # Get batchsize
+    max_num_inkeys = max(len(k) for k in rank_to_inkeys.values())
+    batchsize = batchsize or dask.config.get("xcomm-batchsize", 1)
+    if batchsize == -1:
+        batchsize = max_num_inkeys
```
rjzamora, line 421: My understanding is that the effective default before this PR was `batchsize=-1`. Now it will be `batchsize=1`. Is the idea to make the default as stable as possible? (seems fine to me, just asking for my own understanding)

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -165,24 +194,76 @@ def multi_shuffle_group(
...
     for i in range(npartitions):  # Iterate over all possible output partition IDs
         t = [df_grouped[i] for df_grouped in dfs_grouped]
+        assert len(t) > 0
         if len(t) == 1:
-            ret[i] = t[0]
+            ret[i] = proxify(t[0])
         elif len(t) > 1:
             ret[i] = proxify(dd_concat(t, ignore_index=ignore_index))
-        else:
-            ret[i] = df_meta  # Empty dataframe
```
rjzamora, line 173: Can you explain why we don't need to cover this case anymore? Is it just because we now have no entry at all for an empty DataFrame? Note that I'm happy that we are no longer passing around the _meta if we don't need to be.

Also, is the comment above still true? ("If the partition is empty, an empty dataframe is used.")

rjzamora (commented):
This is exciting to see. Thanks for working on this @madsbk !

Would it be possible to share some empirical benchmark data to show how different `batchsize` choices affect performance (and perhaps the system size you are able to handle on a single DGX)?

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -79,17 +87,17 @@ def compute_map_index(df: Any, column_names, npartitions) -> Series:
 
     Parameters
     ----------
-    df: DataFrame
-    column_names: list of strings
+    df
+        The dataframe.
+    column_names
```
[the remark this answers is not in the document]
madsbk: Yes, no colon if the type is omitted: https://numpydoc.readthedocs.io/en/latest/format.html#parameters

dask_cuda/explicit_comms/dataframe/shuffle.py
madsbk: `group_split_dispatch()` and thus `partition_dataframe()` will always return a dataframe per partition even if some of them becomes empty. 
I cannot remember why the meta was needed in the first place. I have added tests of the different batchsizes and everything seems to work: 268e63fcbb8b6719b8401b02439e3c92834a3a30

pentschev:
rerun tests

madsbk:
Thanks for the review @rjzamora, I think I have addressed all of your suggestions.

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -23,9 +22,72 @@
 T = TypeVar("T")
 
 
+class Proxify(Protocol):
+    """Proxify type hint"""
+
+    def __call__(self, obj: T) -> T:
+        ...
```
wence-, line 29:
```
from typing import Callable
...
Proxify = Callable[[T], T]
```
?

I see this is what you used to have, so there's presumably some reason it's no longer sufficient?

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -23,9 +22,72 @@
...
+
+    # Check that we are shuffling a cudf dataframe
+    try:
+        import cudf
+    except ImportError:
+        return lambda x: x
+    if not stage or not isinstance(next(iter(stage.values())), cudf.DataFrame):
+        return lambda x: x
+
+    # Deep coping a cuDF dataframe doesn't deep copy its index hence
```
wence-, line 80:
```suggestion
    # Deep copying a cuDF dataframe doesn't deep copy its index hence
```

Q: is this a cudf bug?

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -324,6 +471,15 @@ def shuffle(
     rank_to_inkeys = c.stage_keys(name=name, keys=df.__dask_keys__())
     c.client.cancel(df)
 
+    # Get batchsize
+    max_num_inkeys = max(len(k) for k in rank_to_inkeys.values())
+    batchsize = batchsize or dask.config.get("xcomm-batchsize", 1)
```
wence-, line 476: If we're going to introduce some (or more?) configuration parameters for explicit comms, can we think about the naming hierarchy a bit? E.g. we now have `DASK_EXPLICIT_COMMS` to turn on use of explicit comms for the shuffle, but now `DASK_XCOMM_BATCHSIZE`. How are we to know these two are related?

Would prefer something hierarchical like:

```
dask.explicit-comms.enabled = True
dask.explicit-comms.batchsize = ...
```

And however that translates to environment variables.

wence- (commented):
Mostly minor comments.

dask_cuda/explicit_comms/dataframe/shuffle.py
madsbk: It was mypy that compiled at some point but it seems to accept `Proxify = Callable[[T], T]` now. I must have changed something :)

Anyways, I think we should begin using mypy in CI: https://github.com/rapidsai/dask-cuda/issues/1077

dask_cuda/explicit_comms/dataframe/shuffle.py
madsbk: Good point, what about ?
```
dask.explicit-comms = True
dask.explicit-comms-batchsize = ...
```

dask_cuda/explicit_comms/dataframe/shuffle.py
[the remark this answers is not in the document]
rjzamora: >The colon must be preceded by a space, or omitted if the type is absent.

Ah-ha - There it is. Thanks for the link!

dask_cuda/explicit_comms/dataframe/shuffle.py
rjzamora: >group_split_dispatch() and thus partition_dataframe() will always return a dataframe per partition even if some of them becomes empty.

Ah - Sorry. I didn't see the `assert len(t) > 0` line above this. It makes sense to remove this if the zero-length case is explicitly forbidden anyway.

dask_cuda/explicit_comms/dataframe/shuffle.py
rjzamora: @wence- Perhaps we should consider how this choice would potentially interact with your proposal to make it possible to register the shuffle algorithm externally (e.g. https://github.com/dask/dask/pull/9521).  I don't have a clear design/pattern in my mind to suggest, but I thought it was worth mentioning.

rjzamora (approved):
Thanks @madsbk !  As long as your local benchmark results make you confident that this PR provides a good balance between stability and performance, the changes seem good to me.

dask_cuda/explicit_comms/dataframe/shuffle.py
wence-: @rjzamora I think there are two slightly orthogonal things here:

1. How to enable the explicit comms shuffle
2. How to configure aspects of the explicit comms shuffle

Right now, we basically enable through dask config variables, and this PR introduces some configuration slot.

I agree that it would be nice to be able to register the shuffle externally, but I think that is slightly different.

Here, I'm just wondering about how we minimise the proliferation of different naming hierarchies in the dask configuration dictionary (and potentially make them more discoverable).

We could (not in this PR) collect all of the dask-cuda related configuration slots into a `dask-cuda.yaml` that we load and insert into the dask config (as `distributed` and `dask` itself do). This would also allow us to maintain the defaults in one place for all configuration options.

wence-: > Good point, what about ?
> 
> ```
> dask.explicit-comms = True
> dask.explicit-comms-batchsize = ...
> ```

I think I am +epsilon in favour of hierarchical separation so `explicit-comms.batchsize`, `explicit-comms.enabled`, but I am not that fussed.

rjzamora: Yeah, I totally agree that the two goals are (mostly) orthogonal. However, if we were to deprecate something like `dask.explicit-comms = True`, it seems reasonable to consider if the change would also make sense in the future when external registration is hopefully a thing. (I don't think it matters much, but wasn't sure)

madsbk: Agree, if https://github.com/dask/dask/pull/9521 gets in, it make sense to have common config hierarchy for all the shuffle backends.

But I think we should keep `dask.explicit-comms = True` for now.

wence- (approved):
Thanks!

PR #1074, opened 2023-01-06
Improve shuffle-benchmark

madsbk:
Adding `--ignore-index` and balance the partition distribution between workers.

This should make the runs more consist and improve the data creation significantly.

### DIFF
--- a/dask_cuda/benchmarks/local_cudf_shuffle.py
+++ b/dask_cuda/benchmarks/local_cudf_shuffle.py
@@ -57,7 +57,8 @@ def create_df(nelem, df_type):
 def create_data(
     client: Client, args, name="balanced-df"
 ) -> Tuple[int, dask.dataframe.DataFrame]:
-    """Create evenly distributed a dask dataframe"""
+    """Create an evenly distributed dask dataframe"""
+
     workers = list(client.scheduler_info()["workers"].keys())
     assert len(workers) > 0
 
--- a/dask_cuda/benchmarks/local_cudf_shuffle.py
+++ b/dask_cuda/benchmarks/local_cudf_shuffle.py
@@ -57,7 +57,11 @@ def create_df(nelem, df_type):
 def create_data(
     client: Client, args, name="balanced-df"
 ) -> Tuple[int, dask.dataframe.DataFrame]:
-    """Create an evenly distributed dask dataframe"""
+    """Create an evenly distributed dask dataframe
+
+    The partitions are perfectly distributed across workers, if the number of
+    requested partitions is evenly divisible by the number of workers.
+    """
 
     workers = list(client.scheduler_info()["workers"].keys())
     assert len(workers) > 0
--- a/dask_cuda/benchmarks/local_cudf_shuffle.py
+++ b/dask_cuda/benchmarks/local_cudf_shuffle.py
@@ -57,8 +57,8 @@ def create_df(nelem, df_type):
 def create_data(
     client: Client, args, name="balanced-df"
 ) -> Tuple[int, dask.dataframe.DataFrame]:
-    """Create a dask dataframe that are distributed evenly"""
-    workers = list(client.run(lambda: 42).keys())
+    """Create evenly distributed a dask dataframe"""
+    workers = list(client.scheduler_info()["workers"].keys())
     assert len(workers) > 0
 
     chunksize = args.partition_size // np.float64().nbytes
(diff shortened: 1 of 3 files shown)


dask_cuda/benchmarks/local_cudf_shuffle.py
```diff
@@ -21,41 +24,69 @@
...
+    import cupy  # isort:skip
+    import cudf  # isort:skip
+
+    return cudf.DataFrame({"data": cupy.random.random(nelem)})
+
+
+def create_data(
+    client: Client, args, name="balanced-df"
+) -> Tuple[int, dask.dataframe.DataFrame]:
+    """Create a dask dataframe that are distributed evenly"""
```
pentschev, line 60:
```suggestion
    """Create evenly distributed a dask dataframe"""
```

dask_cuda/benchmarks/local_cudf_shuffle.py
```diff
@@ -21,41 +24,69 @@
...
+    workers = list(client.run(lambda: 42).keys())
+    assert len(workers) > 0
```
pentschev, line 62: Do we expect not to have at least one worker in practice? If so, then it probably makes sense to check this in https://github.com/rapidsai/dask-cuda/blob/74b4557df64fbf42461060b4bb536a6b5249202e/dask_cuda/benchmarks/common.py#L114

dask_cuda/benchmarks/local_cudf_shuffle.py
```diff
@@ -21,41 +24,69 @@
...
+    dsk = {}
+    for i in range(args.in_parts):
+        worker = workers[i % len(workers)]  # Round robin
+        dsk[(name, i)] = client.submit(
+            create_df, chunksize, args.type, workers=[worker], pure=False
+        )
+    wait(dsk.values())
+
+    divs = [None] * (len(dsk) + 1)
+    ret = new_dd_object(dsk, name, create_df(0, "cpu"), divs).persist()
```
pentschev, line 77: Is forcing `"cpu"` intentional here or would we want it to match `args.type`?

pentschev (commented):
Thanks @madsbk , left a few suggestions/questions.

dask_cuda/benchmarks/local_cudf_shuffle.py
wence-, line 60: Create a dask dataframe attempting to give an equal number of rows to each worker.

The partitions are perfectly distributed across workers, if the number of requested partitions is evenly divisible by the number of workers.

??

wence- (changes requested):
Minor quibbles (mostly around wording), but otherwise looks good.

dask_cuda/benchmarks/local_cudf_shuffle.py
```diff
@@ -21,41 +24,69 @@
...
-    totalsize = chunksize * nchunks
-    x = da.random.random((totalsize,), chunks=(chunksize,))
-    df = dask.dataframe.from_dask_array(x, columns="data").to_frame()
+def create_df(nelem, df_type):
+    if df_type == "cpu":
+        return pd.DataFrame({"data": np.random.random(nelem)})
+
+    assert df_type == "gpu"
+    import cupy  # isort:skip
+    import cudf  # isort:skip
```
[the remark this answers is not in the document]
madsbk: It is also because of an annoying newline between the two imports. 
Using `else:` now keeps the `# isort:skip`

dask_cuda/benchmarks/local_cudf_shuffle.py
```diff
@@ -21,41 +24,69 @@
...
+    workers = list(client.run(lambda: 42).keys())
+    assert len(workers) > 0
+
+    chunksize = args.partition_size // np.float64().nbytes
+    # Distribute the new partitions between workers by round robin.
+    # We use `client.submit` to control the distribution exactly.
+    # TODO: support unbalanced partition distribution
+    dsk = {}
+    for i in range(args.in_parts):
+        worker = workers[i % len(workers)]  # Round robin
```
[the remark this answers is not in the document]
madsbk: True

dask_cuda/benchmarks/local_cudf_shuffle.py
madsbk: Good catch, fixed

dask_cuda/benchmarks/local_cudf_shuffle.py
madsbk: Done

madsbk:
Thanks for the reviews guys, anything else?

pentschev (approved):
LGTM, thanks @madsbk . Will leave it open for @wence- final approval as well.

wence- (approved):
Thanks!

PR #1082, opened 2023-01-13
Add timeout to `pytest` command

ajschmidt8:
There were two instances recently (below) where some Python test errors caused the `conda-python-tests` job to run/hang for ~4 hours.

- https://github.com/rapidsai/dask-cuda/pull/981#issuecomment-1382289752
- https://github.com/rapidsai/dask-cuda/pull/1081#issuecomment-1382288016

To prevent this from happening again in the future, I've added a reasonable timeout of ~~45 minutes to that particular job~~ 30 minutes to the `pytest` command.

The job usually takes ~25 minutes to complete entirely, so 30 minutes just for `pytest` should be plenty.

This timeout will help prevent jobs from hanging and thus help preserve our finite GPU capacity for CI (particularly for `arm` nodes).

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -28,6 +28,7 @@ jobs:
     with:
       build_type: pull-request
   conda-python-tests:
+    timeout-minutes: 45
     needs: conda-python-build
     secrets: inherit
     # TODO: Switch this testing branch to "cuda-118" after `cudf` `3.10` builds are out.
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -28,7 +28,6 @@ jobs:
     with:
       build_type: pull-request
   conda-python-tests:
-    timeout-minutes: 45
     needs: conda-python-build
     secrets: inherit
     # TODO: Switch this testing branch to "cuda-118" after `cudf` `3.10` builds are out.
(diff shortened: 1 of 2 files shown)


AyodeAwe (approved):

ajschmidt8:
the `timeout-minutes` key doesn't seem to be supported by reusable workflows.

It throws the error here: https://github.com/rapidsai/dask-cuda/actions/runs/3914358886

That's very unfortunate.

pentschev:
I thought there always was a timeout, wasn't that the case before?

ajschmidt8:
> I thought there always was a timeout, wasn't that the case before?

There is a default timeout of 6 hours for GH Action workflows.

Generally you can override this at the job level, but it doesn't seem to be supported for reusable workflows (e.g. [these](https://github.com/rapidsai/dask-cuda/blob/b42151d8bfe9c28be46d1ace7e0e2be26a4de06d/.github/workflows/pr.yaml#L36) reusable workflows).

I could add a timeout to the reusable workflow itself, but that wouldn't be ideal since all of our repositories use these reusable workflows and their runtimes may be different. I was hoping to be able to set it at the repo level.

I think an adequate workaround is to just prefix `pytest` with `timeout 45m`.

ajschmidt8:
admin merging w/ permission from @charlesbluca while they investigate the underlying issue of the tests hanging.

PR #1084, opened 2023-01-16
Make proxy tests with `LocalCUDACluster` asynchronous

pentschev:
After https://github.com/dask/distributed/pull/7429 was merged, some of those tests started hanging and I could confirm there were two threads concurrently attempting to take the UCX spinlock and the GIL, which led to such deadlock. UCX-Py is currently not thread-safe, and indeed can cause problems like this should two or more threads attempt to call communication routines that will required the UCX spinlock. My theory is that the synchronous cluster will indeed cause communication on the main thread (in this case, the `pytest` thread) upon attempting to shutdown the cluster, instead of only within the Distributed communication thread, likely being the reason behind the test hanging.

Asynchronous Distributed clusters seem not to cause any communication from the main thread, but only in the communication thread as expected, thus making the tests asynchronous suffice to resolve such issues. In practice, it's unlikely that people will use sync Distributed clusters from the same process (as pytest does), and thus it's improbable to happen in real use-cases.

### DIFF
--- a/dask_cuda/benchmarks/local_cudf_groupby.py
+++ b/dask_cuda/benchmarks/local_cudf_groupby.py
@@ -107,6 +107,7 @@ def bench_once(client, args, write_profile=None):
         t1 = clock()
         agg = apply_groupby(
             df,
+            backend=args.backend,
             sort=args.sort,
             split_out=args.split_out,
             split_every=args.split_every,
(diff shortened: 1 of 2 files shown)


madsbk (approved):
Awesome, thanks @pentschev !

pentschev:
Thanks @madsbk for the review/approval!

PR #1085, opened 2023-01-17
Use `pkgutil.iter_modules` to get un-imported module for `test_pre_import`

charlesbluca:
Changed this because IIUC `pkg_resources.working_set` is listing the installed distributions and not necessarily the importable modules; this becomes an issue if the distribution and module names aren't the same (e.g. one would `conda install pillow` and then `import PIL`), which was causing some failures in CI that seem unrelated to the changes here.

_Originally posted by @charlesbluca in https://github.com/rapidsai/dask-cuda/pull/981#discussion_r1072650294_

### DIFF
--- a/dask_cuda/tests/test_dask_cuda_worker.py
+++ b/dask_cuda/tests/test_dask_cuda_worker.py
@@ -1,11 +1,11 @@
 from __future__ import absolute_import, division, print_function
 
 import os
+import pkgutil
 import subprocess
 import sys
 from unittest.mock import patch
 
-import pkg_resources
 import pytest
 
 from distributed import Client, wait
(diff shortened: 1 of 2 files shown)


jakirkham:
Wonder if we can do this with [`importlib.metadata`]( https://docs.python.org/3/library/importlib.metadata.html ) instead

Note that there is a backport package [`importlib-metadata`]( https://importlib-metadata.readthedocs.io/en/latest/api.html ) for anything that is not yet included in the Python standard library of the Python versions we support

pentschev (approved):
I think you are right indeed @charlesbluca , good catch and thanks for the investigation/fixing work! 😄

charlesbluca:
Converted to draft to try out @jakirkham's suggestion of using `importlib.metadata`/`importlib_metadata` instead

charlesbluca:
After doing some digging into `importlib.metadata`'s API, I wasn't able to find a suitable method that could grab a list of all working import packages; the closest thing I could find was `packages_distributions()`, but at least locally this seems to list some non-importable package names:

```python
>>> importlib_metadata.packages_distributions()
{'... 'docker': ['ucx-py'], ...}
>>> import docker
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'docker'
```

Since this issue is blocking CI at the moment, and my local testing of `pkgutil.iter_modules()` seems to show that it doesn't have this issue, I'm inclined to go ahead and merge this and revisit if it becomes a problem again. Though if you have a specific API in mind here would be happy to open up / review a follow up 🙂

PR #1080, opened 2023-01-11
Ensure tests run for Python `3.10`

ajschmidt8:
Previously we had disabled `cucim` testing for Python `3.10` because the tests depended on `3.10` packages of `cudf`, which weren't previously available.

Now that `3.10` packages of `cudf` are available, we can enable `3.10` testing for `cucim`.

PR #1086, opened 2023-01-18
Update tests for Python 3.10

pentschev:
Because in Python 3.10 `asyncio.get_event_loop()` does not create an event loop anymore, using synchronous `LocalCluster` raises `DeprecationWarning`s in `tornado.ioloop.IOLoop`. Ideally we should update all tests to `async`, the changes here are the minimum necessary to unblock Python 3.10.


ajschmidt8:
xref: https://github.com/rapidsai/dask-cuda/pull/1080

quasiben (approved):

pentschev:
Thanks @quasiben !

PR #1088, opened 2023-01-19
Remove `--get-cluster-configuration` option, check for scheduler in `dask cuda config`

charlesbluca:
As @pentschev brought up in https://github.com/rapidsai/dask-cuda/pull/981#discussion_r1069887904, we shouldn't need the `--get-cluster-configuration` option for `dask cuda config` since it only enables/disables printing the cluster configuration.

Also added a check to ensure that a scheduler address or scheduler file has been specified, as otherwise IIUC running `dask cuda config` would just end up starting up and querying a local cluster on CPU.

### DIFF
--- a/dask_cuda/cli.py
+++ b/dask_cuda/cli.py
@@ -430,22 +430,12 @@ def worker(
 @scheduler
 @preload_argv
 @scheduler_file
-@click.option(
-    "--get-cluster-configuration",
-    "get_cluster_conf",
-    default=False,
-    is_flag=True,
-    required=False,
-    show_default=True,
-    help="""Print a table of the current cluster configuration""",
-)
 @tls_ca_file
 @tls_cert
 @tls_key
 def config(
     scheduler,
     scheduler_file,
-    get_cluster_conf,
     tls_ca_file,
     tls_cert,
     tls_key,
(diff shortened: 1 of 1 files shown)


dask_cuda/cuda_worker.py
```diff
@@ -111,13 +111,16 @@ def del_pid_file():
         kwargs = {"worker_port": None, "listen_address": None, **kwargs}
 
         if (
-            not scheduler
-            and not scheduler_file
+            scheduler is None
+            and scheduler_file is None
```
pentschev, line 115: Just curious, is this change only to be explicit, or was there any case where the previous code would actually fail in practice?

pentschev (approved):
LGTM, thanks @charlesbluca .

dask_cuda/cuda_worker.py
charlesbluca: Just to be explicit, though I suppose if a user were to use the `CUDAWorker` constructor directly this could've led to some cryptic error messages for falsey `scheduler` or `scheduler_file` values? In practice wouldn't expect this to be a common issue

pentschev: Yeah, that makes sense. I was not considering where users could construct `CUDAWorker` directly, although I don't think it exists in practice today. Anyway, was just asking to know if you had encountered something already, seems like we're on the same page. 🙂

PR #1089, opened 2023-01-19
add initial docs build

AjayThorve:
The PR adds a docs_build process to the PR and Build workflows for this repository. The generated docs are synced to s3 for only the build workflows.

cc @ajschmidt8


pentschev (commented):
Sorry if I missed some previous conversation on this subject, but does this mean we won't use RTD anymore for documentation?

ajschmidt8:
> Sorry if I missed some previous conversation on this subject, but does this mean we won't use RTD anymore for documentation?

@pentschev, this PR adds documentation building to our GitHub Action workflows.

Previously, the doc build jobs lived in Jenkins, but we need to get them migrated to GHAs since Jenkins is going away.

The changes in this PR don't change any themes or frameworks used by `dask-cuda`, they only add the build process to GHAs.

pentschev:
Thanks for clarifying @ajschmidt8 .

ajschmidt8 (approved):

Issue #1077, opened 2023-01-09
CI: mypy check


madsbk:
Let's work towards passing a mypy check

PR #1091, opened 2023-01-24
pre-commit: spell, whitespace, and mypy check

madsbk:
close https://github.com/rapidsai/dask-cuda/issues/1077


pentschev (approved):
LGTM, thanks @madsbk . What's the idea moving forward, make mypy strict?

madsbk:
> LGTM, thanks @madsbk . What's the idea moving forward, make mypy strict?

I think the first step is to install all the library stubs in CI. Currently, all lot of the code is ignored.

wence-:
> > LGTM, thanks @madsbk . What's the idea moving forward, make mypy strict?
> 
> I think the first step is to install all the library stubs in CI. Currently, a lot of the code is ignored.

If we're using `pre-commit`, then we need to specify the extra packages in the pre-commit environment, so they all need to be pip installable (which might be problematic?)

PR #1092, opened 2023-01-25
Fix whitespace & add URLs in `pyproject.toml`

jakirkham:


pentschev (approved):
LGTM John. Are those just formatting fixes or do they get raised as errors and do not allow building in some specific scenario? In the latter case, should we look for a pre-commit hook to correctly parse that file?

jakirkham:
No issues that I'm aware of. Just happened to spot them

pentschev:
Thanks @jakirkham !

jakirkham:
Thanks Peter! 🙏

PR #1098, opened 2023-01-30
pre-commit: Update isort version to 5.12.0

wence-:
poetry version 1.5.0 broke installs of isort prior to 5.11.5 (see pycqa/isort#2077 and pycqa/isort#2078), so we need to upgrade.


pentschev (approved):
This looks good to me, but I think we also need to fix it in https://github.com/rapidsai/integration/blob/54b31400f570c40fdbc04d93747b298c4956716d/conda/recipes/versions.yaml#L89-L90 . Could you do that as well @wence- ?

PR #1113, opened 2023-02-08
Reduce error handling verbosity in CI tests scripts

AjayThorve:
This PR adds a less verbose [trap method](https://github.com/rapidsai/cugraph/blob/f2b081075704aabc789603e14ce552eac3fbe692/ci/test.sh#L19), for error handling to help ensure that we capture all potential error codes in our test scripts, and works as follows:

- setting an environment variable, EXITCODE, with a default value of 0
- setting a trap statement triggered by ERR signals which will set EXITCODE=1 when any commands return a non-zero exit code
cc @ajschmidt8

### DIFF
--- a/ci/test_python.sh
+++ b/ci/test_python.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
 
 set -euo pipefail
 
@@ -24,17 +24,18 @@ PYTHON_CHANNEL=$(rapids-download-conda-from-s3 python)
 RAPIDS_TESTS_DIR=${RAPIDS_TESTS_DIR:-"${PWD}/test-results"}
 RAPIDS_COVERAGE_DIR=${RAPIDS_COVERAGE_DIR:-"${PWD}/coverage-results"}
 mkdir -p "${RAPIDS_TESTS_DIR}" "${RAPIDS_COVERAGE_DIR}"
-SUITEERROR=0
 
 rapids-print-env
 
 rapids-mamba-retry install \
-  -c "${PYTHON_CHANNEL}" \
+  --channel "${PYTHON_CHANNEL}" \
   dask-cuda
 
 rapids-logger "Check GPU usage"
 nvidia-smi
 
+EXITCODE=0
+trap "EXITCODE=1" ERR
 set +e
 
 rapids-logger "pytest dask-cuda"
@@ -53,12 +54,6 @@ timeout 30m pytest \
   --cov-report=xml:"${RAPIDS_COVERAGE_DIR}/dask-cuda-coverage.xml" \
   --cov-report=term \
   tests
-exitcode=$?
-
-if (( ${exitcode} != 0 )); then
-    SUITEERROR=${exitcode}
-    echo "FAILED: 1 or more tests in dask-cuda"
-fi
 popd
 
 rapids-logger "Run local benchmark"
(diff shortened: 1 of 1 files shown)


pentschev (commented):
Unless I'm missing some detail, I must say I don't like this approach. The problem is, and correct me if I'm wrong, the script will terminate upon the first error and thus we would miss running all the other tests, which IMO is not something we want, if there's a problem we would like to see everything that is failing, not just the first failure.

AjayThorve:
> Unless I'm missing some detail, I must say I don't like this approach. The problem is, and correct me if I'm wrong, the script will terminate upon the first error and thus we would miss running all the other tests, which IMO is not something we want, if there's a problem we would like to see everything that is failing, not just the first failure.

@pentschev from what I understand the [`set -e`](https://github.com/rapidsai/dask-cuda/pull/1113/files#diff-ce857ae7ff36078ab0d5173765bf2bc6df4fbac666cb5c4da6a2c257250c097fR39) command makes sure we do not abort the script on error. This allows all tests to run regardless of pass/fail, but relies on the ERR trap above to manage the EXITCODE for the script.

So in essence, the behavior of the tests scripts should not change by these changes, just a less verbose method of doing the same thing. @ajschmidt8 correct me if I am wrong.

ajschmidt8:
@pentschev, I'd like to keep this PR as it is for consistency with other projects.

To your point about easily discerning test failures: our shared workflow for Python tests makes use of the [test-summary/action](https://github.com/test-summary/action) (see [here](https://github.com/rapidsai/shared-action-workflows/blob/69dcbaf9d33b643f405835df322245ae6180a61c/.github/workflows/conda-python-tests.yaml#L110-L114) for where we implement it).

This means that `pytest` job summaries can be viewed on the GitHub Action workflow [summary cards](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/) (in addition to the job's raw log output).

[Here's a quick example](https://github.com/rapidsai/cuml/actions/runs/4135061636/attempts/1#summary-11228316269) of what that looks like for a failing pytest for cuml.

ajschmidt8 (approved):

ajschmidt8:
@pentschev, I'll wait for your reply to merge.

Or feel free to merge this yourself if you're okay with the current changes.

pentschev:
> This means that `pytest` job summaries can be viewed on the GitHub Action workflow [summary cards](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/) (in addition to the job's raw log output).

In our CI, not everything is a `pytest`, we also have some benchmarks that are just CLI tools, for example [this](https://github.com/rapidsai/dask-cuda/pull/1113/files#diff-ce857ae7ff36078ab0d5173765bf2bc6df4fbac666cb5c4da6a2c257250c097fR60-R64) and [this](https://github.com/rapidsai/dask-cuda/pull/1113/files#diff-ce857ae7ff36078ab0d5173765bf2bc6df4fbac666cb5c4da6a2c257250c097fR66-R70). It's not a big deal currently, but I'd be interested in knowing if there's anything we can also add for non-pytest executables, do you know if there's anything we can report based on their outputs, or just on their exit codes?

pentschev:
> [GitHub Summary Cards](https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/) are markdown files that you can redirect output to.

This looks super cool, I think we could print a summary of our results there, but probably have to rethink how we run those benchmarks, and if it makes sense to run them here. In any case, I think this is a little beyond this PR, so let's merge it as is for now and I'll try to rethink this a bit later. Thanks so much for the suggestion, I was not aware of that!

PR #1118, opened 2023-02-15
Fix for bytes/str discrepancy after PyNVML update

pentschev:
A PyNVML update has changed how some objects were previously returned bytes but now return str, which is now handled appropriately with this change.


pentschev:
Requires https://github.com/dask/distributed/pull/7544 to be merged first.

wence- (approved):
Looks sane, thanks! Will keep an eye and merge when ready.

PR #1126, opened 2023-02-23
Pull `requirements/host` from `pyproject.toml`

jakirkham:
In addition to pulling the Conda recipe's `requirements/run` from `pyproject.toml`, pull `requirements/host` from `pyproject.toml`. This should help solidify around `pyproject.toml` being the single source of truth.

### DIFF
--- a/conda/recipes/dask-cuda/meta.yaml
+++ b/conda/recipes/dask-cuda/meta.yaml
@@ -29,8 +29,9 @@ requirements:
   host:
     - python
     - pip
-    - tomli
-    - versioneer >=0.24
+    {% for r in data.get("build-system", {}).get("requires", []) %}
+    - {{ r }}
+    {% endfor %}
   run:
     - python
     {% for r in data.get("project", {}).get("dependencies", []) %}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,9 +1,9 @@
 [build-system]
 build-backend = "setuptools.build_meta"
 requires = [
-    "setuptools>=64.0.0",
-    "tomli  ; python_version < '3.11'",
-    "versioneer>=0.24",
+    "setuptools >=64.0.0",
+    "tomli",                # Unneeded for Python 3.11+
+    "versioneer >=0.24",
 ]
 
 [project]


jakirkham:
Toggling for CI

pentschev:
Failures are unrelated, tracking that in https://github.com/rapidsai/dask-cuda/issues/1134 .

pentschev (approved):
LGTM, thanks @jakirkham !

ajschmidt8 (commented):
@jakirkham, is this a pattern we plan on using for other RAPIDS libs?

I think `dask-cuda`'s requirements are a bit simpler than other libs (e.g. no `pin_compatible` or `pin_subpackage`).

So I'm wondering if this would work for more complex cases

jakirkham:
Yeah this is something I'm looking into. Please see PR ( https://github.com/rapidsai/rmm/pull/1220 ) for example

raydouglass (approved):

ajschmidt8 (changes requested):
Temporarily requesting changes to block this PR from being merged.

I think we need to talk more about dependency management based on @vyasr's comment below.

- https://github.com/rapidsai/rmm/pull/1220#issuecomment-1442485894

Seems like we have conflicting efforts in motion so we should probably sync on the best way to handle this so that we're consistent across all repositories.

jakirkham:
Generally agree we should discuss how we adopt this in RAPIDS more generally

However with Dask-CUDA would push back on this slightly since there is a longer history here of wanting requirements in one place that is pulled into other sources ( https://github.com/rapidsai/dask-cuda/issues/569 ) ( https://github.com/rapidsai/dask-cuda/pull/893 ). Additionally Dask-CUDA being a pure Python package has generally been more aligned with requirements in Conda & PyPI packages. So in the Dask-CUDA case this is more a continuation of how things have been as opposed to a new approach

vyasr:
The main question from the rmm PR that IMO needs to be answered before we move forward with changes like this even in dask-cuda is how we want to handle the interaction between rapids-dependency-file-generator and meta.yaml reading from pyproject.toml. Those are two different approaches for single-sourcing dependencies, and I would prefer not adding two levels of indirection (dependencies.yaml->pyproject.toml->meta.yaml) where only one might suffice if we could directly write meta.yaml from dependencies.yaml. The open questions in the rmm PR would help us understand whether there is a strong argument in favor of the two pass approach (maybe because we simply decide that supporting meta.yaml is too much trouble for dfg).

PR #1129, opened 2023-02-24
Adapt to rapidsai/rmm#1221 which moves allocator callbacks

wence-:
The allocator callbacks now live in their own submodules (so that RMM does not, for example, import pytorch unless required) and so must be explicitly imported.


pentschev (approved):
LGTM, thanks Lawrence!

PR #1140, opened 2023-03-13
Reinstate `--death-timeout` CLI option

### LINKED ISSUE #1017: Reinstate --death-timeout
It looks like the `--death-timeout` CLI option was removed in #563.

We just had a bug report on `dask-cloudprovider` about this as we use it there. I'm surprised this hasn't come up in the last 18 months, perhaps folks are pinning to older versions or not using GPUs with `dask-cloudprovider` and ECS much.

https://github.com/dask/dask-cloudprovider/issues/384

@charlesbluca asked in https://github.com/dask/distributed/issues/4685 about removing this as it appeared to do nothing but James shows an example that suggests it does still function as expected. So should this be reinstated?

charlesbluca:
Add back in the `--death-timeout` option removed in #563, along with some tests to verify it's working as expected.

Closes #1017

### DIFF
--- a/dask_cuda/cli.py
+++ b/dask_cuda/cli.py
@@ -244,6 +244,12 @@ def cuda():
     ``"/path/to/foo.py"``.""",
 )
 @click.option(
+    "--death-timeout",
+    type=str,
+    default=None,
+    help="Seconds to wait for a scheduler before closing",
+)
+@click.option(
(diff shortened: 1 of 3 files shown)


pentschev (approved):
LGTM, thanks @charlesbluca !

PR #1141, opened 2023-03-21
Update usage of `get_worker()` in tests

pentschev:
In https://github.com/dask/distributed/pull/7580/ `get_worker` was modified to return the worker of a task, thus it cannot be used by `client.run`, and we must now use `dask_worker` as the first argument to `client.run` to obtain the worker.

### DIFF
--- a/dask_cuda/tests/test_explicit_comms.py
+++ b/dask_cuda/tests/test_explicit_comms.py
@@ -11,7 +11,7 @@ import dask
 from dask import dataframe as dd
 from dask.dataframe.shuffle import partitioning_index
 from dask.dataframe.utils import assert_eq
-from distributed import Client, get_worker
+from distributed import Client
 from distributed.deploy.local import LocalCluster
 
 import dask_cuda
@@ -314,8 +314,8 @@ def test_jit_unspill(protocol):
 
 
 def _test_lock_workers(scheduler_address, ranks):
-    async def f(_):
-        worker = get_worker()
+    async def f(info):
+        worker = info["worker"]
         if hasattr(worker, "running"):
             assert not worker.running
         worker.running = True
--- a/dask_cuda/tests/test_local_cuda_cluster.py
+++ b/dask_cuda/tests/test_local_cuda_cluster.py
@@ -9,7 +9,6 @@ import pytest
 from dask.distributed import Client
 from distributed.system import MEMORY_LIMIT
 from distributed.utils_test import gen_test, raises_with_cause
-from distributed.worker import get_worker
 
 from dask_cuda import CUDAWorker, LocalCUDACluster, utils
 from dask_cuda.initialize import initialize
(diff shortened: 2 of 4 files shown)


jacobtomlinson (commented):
Looks like this test also needs updating

https://github.com/rapidsai/dask-cuda/blob/ec2de7868e191d092f7215ed4bca9f2453b78e9a/dask_cuda/tests/test_proxify_host_file.py#L418

wence-:
@madsbk there are usages of `get_worker` inside the explicit comms implementation. Are these likely to be safe because they're run in a context where there is a task?

pentschev:
> Are these likely to be safe because they're run in a context where there is a task?

I guess you mean https://github.com/rapidsai/dask-cuda/blob/ec2de7868e191d092f7215ed4bca9f2453b78e9a/dask_cuda/explicit_comms/comms.py#L77 which is called from https://github.com/rapidsai/dask-cuda/blob/ec2de7868e191d092f7215ed4bca9f2453b78e9a/dask_cuda/explicit_comms/comms.py#L92-L93 and submitted from https://github.com/rapidsai/dask-cuda/blob/ec2de7868e191d092f7215ed4bca9f2453b78e9a/dask_cuda/explicit_comms/comms.py#L228-L235

The above should be fine as `_run_coroutine_on_worker` and is executed by `client.submit`.

madsbk (approved):
LGTM, thanks Peter

wence-:
> The above should be fine as `_run_coroutine_on_worker` and is executed by `client.submit`.

884a595 says "Using `client.submit` seems not allowed to resolve `get_worker`, but `client.run` does."

So this seems in conflict?

pentschev:
> [884a595](https://github.com/rapidsai/dask-cuda/commit/884a59563c0daee3a13e8c8f1520b7928366c083) says "Using `client.submit` seems not allowed to resolve `get_worker`, but `client.run` does."
> 
> So this seems in conflict?

I meant that for `async` functions only, it does work with non-`async`. I've also commented about that behavior in https://github.com/dask/distributed/issues/7696#issuecomment-1479689775 , but I currently don't know if it's possible to resolve, although it shouldn't be a problem for us currently, as we don't use it in actual code, just in that one test.

wence-:
Ah, ok, thanks for the explanation and digging!

pentschev:
Thanks everyone for reviews!

pentschev:
For completeness in this thread, regarding the substitution of `client.submit` by `client.run`, this is what Distributed devs answered in https://github.com/dask/distributed/issues/7696#issuecomment-1479965546: 

> I looked at https://github.com/rapidsai/dask-cuda/commit/884a59563c0daee3a13e8c8f1520b7928366c083 and argue this "workaround" is how it is supposed to be. client.submit is intended to run user functions (on a thread) while client.run is intended to run a function/coroutine on the worker event loop, in the main thread, outside of our task scheduling. If you are actually inspecting the worker in your test, that's how you should do it

PR #1145, opened 2023-03-28
Add argument to enable RMM alloaction tracking in benchmarks

pentschev:
Tracking RMM allocation will be useful together with https://github.com/dask/distributed/pull/5740 , and will help with the analysis of memory fragmentation when comparing regular pool and the async memory allocator.


quasiben (approved):

pentschev:
Thanks @quasiben !

PR #1147, opened 2023-03-29
Allow specifying dashboard address in benchmarks

pentschev:
This is useful for shared machines where the user may not have control of the default port `8787`.


wence- (approved):
Thanks!

pentschev:
Thanks @wence- @jacobtomlinson !

PR #1151, opened 2023-04-03
Work-around for dask/distributed#7726 and dask-cuda#1148

wence-:
The last atexit finalizer in the chain of finalizers that distributed registers needs to be one in the weakref module. Otherwise we observe hangs in the shutdown of UCX-Py where one thread has the gil but not the UCX spinlock and vice-versa.

To ensure this, unregister the known shutdown finalizers in distributed, add our weakref finalizer, and then re-register them. Along with a final re-implementation of the "python shutting down" handler which must run before other handlers in distributed.

### DIFF
--- a/dask_cuda/local_cuda_cluster.py
+++ b/dask_cuda/local_cuda_cluster.py
@@ -23,6 +23,48 @@ from .utils import (
 )
 
 
+def workaround_distributed_7726():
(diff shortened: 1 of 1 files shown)


wence-:
If @randerzander reports success, I think we could try this workaround? (cc @quasiben, @pentschev)

wence-:
Closing for now since it seems we don't have the necessity for this.

PR #1153, opened 2023-04-05
[REVIEW] Pin `dask` and `distributed` for release

galipremsagar:
This PR pins `dask` and `distributed` to `2023.3.2` and `2023.3.2.1` respectively for `23.04` release.

xref: https://github.com/rapidsai/cudf/pull/13070

PR #1155, opened 2023-04-05
Rectify `dask-core` pinning in pip requirements

galipremsagar:
As part of https://github.com/rapidsai/dask-cuda/pull/1153 `dask-core` has been added to pip requirements, which is incorrect. This PR rectifies this issue.


galipremsagar:
The issue was not caught in https://github.com/rapidsai/dask-cuda/pull/1153, because of missing wheels build & test jobs.

raydouglass (approved):

Issue #1152, opened 2023-04-04
Is the use of `__main__` necessary for `LocalCUDACluster`?


trivialfis:
Hi, I just read the new quickstart here: https://docs.rapids.ai/api/dask-cuda/stable/quickstart.html#localcudacluster thanks to https://github.com/dmlc/xgboost/pull/9008 .

My question is, in the Python script environment instead of a notebook, is wrapping code under:
``` python
if __name__ == "__main__":
    ...
```
necessary for CUDA cluster?

trivialfis:
Thank you for confirming! Would you like me to open a PR for document to reflect this requirement?

wence-:
Oh, yes please! Please target the 23.06 branch.

PR #1157, opened 2023-04-06
[doc] Add document about main guard.

trivialfis:
Close https://github.com/rapidsai/dask-cuda/issues/1152 .

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -13,13 +13,8 @@ To create a Dask-CUDA cluster using all available GPUs and connect a Dask.distri
     from dask_cuda import LocalCUDACluster
     from dask.distributed import Client
 
-
-    if __name__ == "__main__":
-        with LocalCUDACluster() as cluster:
-            with Client(cluster) as client:
-	        pass
-
-The `__main__` guard is required when running in Python script mode (instead of a notebook environment).
+    cluster = LocalCUDACluster()
+    client = Client(cluster)
 
 ``dask cuda worker``
 --------------------
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -13,8 +13,13 @@ To create a Dask-CUDA cluster using all available GPUs and connect a Dask.distri
     from dask_cuda import LocalCUDACluster
     from dask.distributed import Client
 
-    cluster = LocalCUDACluster()
-    client = Client(cluster)
+
+    if __name__ == "__main__":
+        with LocalCUDACluster() as cluster:
+            with Client(cluster) as client:
+	        pass
+
+The `__main__` guard is required when running in Python script mode (instead of a notebook environment).
 
 ``dask cuda worker``
 --------------------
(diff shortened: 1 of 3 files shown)


docs/source/quickstart.rst
```diff
@@ -13,8 +13,13 @@ To create a Dask-CUDA cluster using all available GPUs and connect a Dask.distri
...
 
-    cluster = LocalCUDACluster()
-    client = Client(cluster)
+
+    if __name__ == "__main__":
+        with LocalCUDACluster() as cluster:
+            with Client(cluster) as client:
+	        pass
+
+The `__main__` guard is required when running in Python script mode (instead of a notebook environment).
```
charlesbluca, line 22: Think my preference here would be to keep this code block as is, but make it explicit that this would be running from the context of a Python console (Distributed's [quickstart](https://distributed.dask.org/en/latest/quickstart.html#setup-dask-distributed-the-easy-way) for reference), and then include a note here linking to Distributed's upstream docs on [how to run a cluster from a Python script](https://docs.dask.org/en/stable/scheduling.html?highlight=__name__%20__main__#standalone-python-scripts).

charlesbluca (commented):
Thanks @trivialfis 😄 I notice that the Dask docs actually have a note around this cropping up a few times throughout the docs ([example](https://docs.dask.org/en/stable/scheduling.html?highlight=__name__%20__main__#dask-distributed-local)), I wonder if it would be worth doing the same here?

docs/source/quickstart.rst
trivialfis: Thank you for the suggestion, I tried to mimic the dask doc. Along with which, the following warnings are fixed:
```
/home/jiamingy/Workspace/XGBoost-dev/dask-cuda/docs/source/examples/best-practices.rst:13: WARNING: Bullet list ends without a blank line; unexpected unindent.
/home/jiamingy/Workspace/XGBoost-dev/dask-cuda/docs/source/examples/ucx.rst:125: WARNING: Title underline too short.

``dask cuda worker`` with Manual Configuration
------------------------------------------
/home/jiamingy/Workspace/XGBoost-dev/dask-cuda/docs/source/examples/ucx.rst:125: WARNING: Title underline too short.

``dask cuda worker`` with Manual Configuration
------------------------------------------
/home/jiamingy/Workspace/XGBoost-dev/dask-cuda/docs/source/ucx.rst:2: WARNING: Duplicate explicit target name: "here".
```

docs/source/quickstart.rst
```diff
@@ -16,6 +16,10 @@ To create a Dask-CUDA cluster using all available GPUs and connect a Dask.distri
     cluster = LocalCUDACluster()
     client = Client(cluster)
 
+.. tip::
+
+   Be sure to include an ``if __name__ == "__main__":`` block when using :py:class:`dask_cuda.LocalCUDACluster` in a standalone Python script. See `Standalone Python scripts <https://docs.dask.org/en/stable/scheduling.html#standalone-python-scripts>`_ for more detials.
```
wence-, line 21:
```suggestion
   Be sure to include an ``if __name__ == "__main__":`` block when using :py:class:`dask_cuda.LocalCUDACluster` in a standalone Python script. See `standalone Python scripts <https://docs.dask.org/en/stable/scheduling.html#standalone-python-scripts>`_ for more details.
```

wence- (approved):
Thanks for the fix, plus the RST sanitization!

PR #1160, opened 2023-04-13
Update to zict 3.0

pentschev:
With the release of zict 3.0 a few changes were made to resources that were used in spilling tests that are being updated here.

### DIFF
--- a/dask_cuda/tests/test_spill.py
+++ b/dask_cuda/tests/test_spill.py
@@ -30,8 +30,8 @@ def device_host_file_size_matches(
     # `dhf.disk` is only available when Worker's `memory_limit != 0`
     if dhf.disk is not None:
         file_path = [
-            os.path.join(dhf.disk.directory, dhf.disk._safe_key(k))
-            for k in dhf.disk.keys()
+            os.path.join(dhf.disk.directory, fname)
+            for fname in dhf.disk.filenames.values()
         ]
         file_size = [os.path.getsize(f) for f in file_path]
         byte_sum += sum(file_size)
--- a/dask_cuda/tests/test_spill.py
+++ b/dask_cuda/tests/test_spill.py
@@ -2,7 +2,6 @@ import os
 from time import sleep
 
 import pytest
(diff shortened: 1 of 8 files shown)


dask_cuda/tests/test_spill.py
```diff
@@ -31,7 +30,8 @@ def device_host_file_size_matches(
     # `dhf.disk` is only available when Worker's `memory_limit != 0`
     if dhf.disk is not None:
         file_path = [
-            os.path.join(dhf.disk.directory, safe_key(k)) for k in dhf.disk.keys()
+            os.path.join(dhf.disk.directory, dhf.disk._safe_key(k))
+            for k in dhf.disk.keys()
```
crusaderky, line 34:
```suggestion
            os.path.join(dhf.disk.directory, fname)
            for fname in dhf.disk.filenames.values()
```

pentschev: Checked this locally and the test pass with that change. Thanks @crusaderky !

quasiben:
rerun tests

pentschev:
rerun tests

jakirkham:
Does [this requirement]( https://github.com/rapidsai/dask-cuda/blob/eed3eb6b02951ac2311b81f546b9053184254c4b/dependencies.yaml#L104 ) need updating?

pentschev:
> Does [this requirement](https://github.com/rapidsai/dask-cuda/blob/eed3eb6b02951ac2311b81f546b9053184254c4b/dependencies.yaml#L104) need updating?

Given the only change required here is to a test, I think it's safer if we keep the pinning somewhat loose. I upgraded it to 2.0 so it doesn't stay "too loose".

quasiben:
rerun tests

quasiben:
Not sure why this was failing before.  Anything else needed here @pentschev  ?

pentschev:
> Not sure why this was failing before.

It was failing before zict 3.0 was release, but it's been passing since. 

> Anything else needed here @pentschev ?

I think we are good to merge.

PR #1166, opened 2023-04-25
Temporarily relax Python constraint

vyasr:
This PR unblocks RAPIDS CI since many places attempt to install dask-cuda from source. We can undo this change once the rest of RAPIDS has moved to Python 3.9. We will also want to discuss better strategies for handling dask-cuda in CI as part of our ongoing discussions around improving latest dask usage in CI.

### DIFF
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -2,6 +2,7 @@ import functools
 import itertools
 import logging
 import os
+import sys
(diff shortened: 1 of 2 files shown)


galipremsagar (approved):

charlesbluca (approved):
Thanks @vyasr! Sorry for the breakage here

PR #1171, opened 2023-05-03
Revert "Temporarily relax Python constraint"

vyasr:
We undid the pinning in order to unblock the Python 3.8->3.9 transition in RAPIDS, which is now complete.


pentschev (approved):
Thanks @vyasr !

PR #1181, opened 2023-05-26
Add `__main__` entrypoint to dask-cuda-worker CLI

### LINKED ISSUE #1180: Make `dask_cuda.cli runnable` as a script with `python -m`
I am working towards fixing [#dask-jobqueue/229](https://github.com/dask/dask-jobqueue/issues/229) in which one would like to provide a parameter for which type of `worker` to start when using `jobqueue`

The call site currently looks like this:

```python
        # dask-worker command line build
        dask_worker_command = "%(python)s -m distributed.cli.dask_worker" % dict(
            python=python
        )
```

IE is called through `python -m`. 

`dask_cuda.cli` starts from `worker()` rather than `main` making  it incompatible with this call site. IMO it would be nice to homogenise these interfaces. 

I propose adding a very simple:

```python

if __name__ == "__main__":
    worker()

```
to `dask_cuda.cli` so that they can be called the same way.

hmacdope:
Fixes #1180 

Making the CLI runnable with `python -m ` so that we can use the same call for both CLIs in `dask-jobqueue`

### DIFF
--- a/dask_cuda/cli.py
+++ b/dask_cuda/cli.py
@@ -499,3 +499,6 @@ def config(
     else:
         client = Client(scheduler, security=security)
     print_cluster_config(client)
+
+if __name__ == "__main__":
+    worker()
--- a/dask_cuda/cli.py
+++ b/dask_cuda/cli.py
@@ -500,5 +500,6 @@ def config(
         client = Client(scheduler, security=security)
     print_cluster_config(client)
 
+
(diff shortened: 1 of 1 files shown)


wence- (approved):
I think this makes sense, thanks.

pentschev:
Thanks @hmacdope !

pentschev (approved):

PR #1189, opened 2023-06-02
Unpin `dask` and `distributed` for development

galipremsagar:
This PR unpins `dask` and `distributed` to `>=2023.5.1` for `23.08` development.

xref: https://github.com/rapidsai/cudf/pull/13508

### DIFF
--- a/conda/recipes/dask-cuda/meta.yaml
+++ b/conda/recipes/dask-cuda/meta.yaml
@@ -33,7 +33,7 @@ requirements:
     - versioneer >=0.24
   run:
     - python
-    - dask-core ==2023.3.2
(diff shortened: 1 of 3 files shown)


pentschev:
It seems that none of the builds actually installed a newer Dask/Distributed version, they're all still on `2023.3.2.1`. I'm thinking this has to do with the fact that other test dependencies, such as `cudf`, are pinning to that version. I don't recall if we had this behavior before, but what I'm thinking here is we'll need to update the pin here in Dask-CUDA to `>=2023.3.2.1`, merge it and then unpin other projects after we have it unpinned here. @galipremsagar WDYT?

raydouglass (approved):

PR #1191, opened 2023-06-06
Remove code for Distributed<2023.5.1 compatibility

### LINKED ISSUE #1174: Error serializing CuPy==12.0.0 array

pentschev:
Closes #1174

### DIFF
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -17,7 +17,7 @@ from distributed.protocol import (
     serialize_bytelist,
 )
 from distributed.sizeof import safe_sizeof
-from distributed.utils import has_arg, nbytes
+from distributed.utils import nbytes
 
 from .is_device_object import is_device_object
 from .is_spillable_object import is_spillable_object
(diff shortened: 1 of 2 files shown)


jakirkham:
cc @madsbk

pentschev:
Thanks all!

PR #1192, opened 2023-06-06
Increase minimum timeout to wait for workers in CI

pentschev:
We have been getting timeouts waiting for workers in CI, those are not reproducible locally. The reason for that is probably some sort of congestion causing spinup to take longer in CI, therefore this change introduces a variable that can be used to control the minimum timeout and the minimum timeout is doubled in CI.


ajschmidt8 (approved):

galipremsagar (approved):

pentschev:
Thanks @ajschmidt8 and @galipremsagar !

Issue #1115, opened 2023-02-14
Dask-CUDA: CUDA 12 Conda Packages


jakirkham:
Pushing for CUDA 12 Conda package builds of libraries for RAPIDS 23.04. This issue is for tracking that work.

Requirements (please recheck):

```[tasklist]
### Developer tasks
- [x] I've confirmed the list below is correct (or updated if not)
```

```[tasklist]
### Dependencies from Conda
- [ ] https://github.com/conda-forge/pynvml-feedstock/pull/11
- [ ] https://github.com/conda-forge/numba-feedstock/issues/112
```

```[tasklist]
### RAPIDS Conda dependencies
- [ ] https://github.com/rapidsai/rmm/issues/1207
- [ ] https://github.com/rapidsai/cudf/issues/12771
- [ ] https://github.com/rapidsai/kvikio/pull/224
- [ ] https://github.com/rapidsai/ucx-py/issues/927
```

```[tasklist]
### Optional dependencies from Conda
- [ ] https://github.com/conda-forge/cupy-feedstock/issues/139
- [x] nvtx (Python)
- [ ] https://github.com/conda-forge/ucx-split-feedstock/issues/116
```

wence-:
`(via distributed/ucx-py)`

Not sure what this is supposed to mean. Do you just want to write `ucx-py (RAPIDS)` ?

Also not sure how you're specifying transitive dependencies, e.g. if you're optionally depending on cudf, do you also need to depend on rmm (which is a transitive dep of cudf)?

jakirkham:
^ @pentschev, do you have thoughts on this question?

pentschev:
> `(via distributed/ucx-py)`
> 
> Not sure what this is supposed to mean. Do you just want to write `ucx-py (RAPIDS)` ?

I just tried to be somewhat verbose, not even sure this is a necessity. What I meant to say is that the dependency is Dask-CUDA->Distributed->UCX-Py->UCX, and it is an important dependency for us, thus the verbosity. I'll defer to @jakirkham on how that can be best viewed.

> Also not sure how you're specifying transitive dependencies, e.g. if you're optionally depending on cudf, do you also need to depend on rmm (which is a transitive dep of cudf)?

[...]

PR #1201, opened 2023-06-22
CUDA 12 Support

quasiben:
Adds CUDA 12 Support to build matrix

Fixes https://github.com/rapidsai/dask-cuda/issues/1115

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -67,6 +67,7 @@ dependencies:
               cuda: "12.0"
             packages:
               - cuda-version=12.0
+              - cuda-nvcc
   develop:
     common:
       - output_types: [conda, requirements]
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -46,19 +46,23 @@ dependencies:
           - matrix:
               cuda: "11.2"
             packages:
-              - cudatoolkit=11.2
+              - cuda-version=11.2
+              - cudatoolkit
           - matrix:
               cuda: "11.4"
             packages:
-              - cudatoolkit=11.4
+              - cuda-version=11.4
+              - cudatoolkit
           - matrix:
               cuda: "11.5"
             packages:
-              - cudatoolkit=11.5
+              - cuda-version=11.5
+              - cudatoolkit
           - matrix:
               cuda: "11.8"
             packages:
-              - cudatoolkit=11.8
+              - cuda-version=11.8
+              - cudatoolkit
           - matrix:
               cuda: "12.0"
             packages:
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -62,7 +62,7 @@ dependencies:
           - matrix:
               cuda: "12.0"
             packages:
-              - cudatoolkit=12.0
+              - cuda-version=12.0
   develop:
     common:
       - output_types: [conda, requirements]
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -59,6 +59,10 @@ dependencies:
               cuda: "11.8"
             packages:
               - cudatoolkit=11.8
+          - matrix:
+              cuda: "12.0"
+            packages:
+              - cudatoolkit=12.0
   develop:
     common:
       - output_types: [conda, requirements]
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   conda-python-build:
     secrets: inherit
-    uses: rapidsai/shared-action-workflows/.github/workflows/conda-python-build.yaml@branch-23.08
+    uses: rapidsai/shared-action-workflows/.github/workflows/conda-python-build.yaml@cuda-120
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -38,7 +38,7 @@ jobs:
     if: github.ref_type == 'branch'
     needs: [conda-python-build]
     secrets: inherit
-    uses: rapidsai/shared-action-workflows/.github/workflows/custom-job.yaml@branch-23.08
+    uses: rapidsai/shared-action-workflows/.github/workflows/custom-job.yaml@cuda-120
     with:
       arch: "amd64"
       branch: ${{ inputs.branch }}
@@ -51,7 +51,7 @@ jobs:
   upload-conda:
     needs: [conda-python-build]
     secrets: inherit
-    uses: rapidsai/shared-action-workflows/.github/workflows/conda-upload-packages.yaml@branch-23.08
+    uses: rapidsai/shared-action-workflows/.github/workflows/conda-upload-packages.yaml@cuda-120
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -18,26 +18,26 @@ jobs:
       - docs-build
       - wheel-build
     secrets: inherit
-    uses: rapidsai/shared-action-workflows/.github/workflows/pr-builder.yaml@branch-23.08
+    uses: rapidsai/shared-action-workflows/.github/workflows/pr-builder.yaml@cuda-120
   checks:
     secrets: inherit
-    uses: rapidsai/shared-action-workflows/.github/workflows/checks.yaml@branch-23.08
+    uses: rapidsai/shared-action-workflows/.github/workflows/checks.yaml@cuda-120
   conda-python-build:
     needs: checks
     secrets: inherit
-    uses: rapidsai/shared-action-workflows/.github/workflows/conda-python-build.yaml@branch-23.08
+    uses: rapidsai/shared-action-workflows/.github/workflows/conda-python-build.yaml@cuda-120
     with:
       build_type: pull-request
   conda-python-tests:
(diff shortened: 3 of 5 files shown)


jakirkham:
Thanks Ben! 🙏

Would suggest looking at the RMM CUDA 12 PR ( https://github.com/rapidsai/rmm/pull/1223 ) and lining this up with that

jakirkham:
Think we may also want [these GHA changes]( https://github.com/rapidsai/rmm/pull/1223/files#diff-d0777657fa3fd81d23aaf7273e58aee453b04e67882517900c56daeef9b3e4c1 )

Not seeing any environment files here

Also [`pyproject.toml`]( https://github.com/rapidsai/dask-cuda/blob/83c64765ec6612cb7c04b077e30edf6f846d90e0/pyproject.toml ), which [the Conda recipe pulls from]( https://github.com/rapidsai/dask-cuda/blob/83c64765ec6612cb7c04b077e30edf6f846d90e0/conda/recipes/dask-cuda/meta.yaml#L5 ), doesn't appear to have any CUDA specific changes needed

So that may be it

bdice:
I agree with all @jakirkham's suggestions above. Maybe also need updates for `docs/source/install.rst`?

jakirkham:
Thanks Bradley! 🙏

Good catch. Yeah this `cudatoolkit` here should be changed to `cuda-version`

https://github.com/rapidsai/dask-cuda/blame/83c64765ec6612cb7c04b077e30edf6f846d90e0/docs/source/install.rst#L19

pentschev:
I pushed [GHA](https://github.com/rapidsai/dask-cuda/pull/1201/commits/99017faf2caf5fd916862ee7e804bd717b952fa1) and [docs](https://github.com/rapidsai/dask-cuda/pull/1201/commits/6dfedd7b251169db2b47a5f0f3c22ff2a5631bde) changes.

@jakirkham @bdice could you check that the docs is correct, is that supposed to work for both CUDA 11.x and 12.x, or the correct way to document it is CUDA 11.x requires `cudatoolkit=11.x` and CUDA >= 12.x requires `cuda-verison=12.x`?

pentschev:
Also seems like cuCIM needs to be handled before Dask-CUDA.

jakirkham:
`cuda-version` should work on both CUDA 11 & 12

Ok could you please add cuCIM to the list in this issue ( https://github.com/rapidsai/dask-cuda/issues/1115 )?

pentschev:
I'm not quite sure what's going on in the CUDA 12 tests, it seems Numba is failing to find NVVM files:

```
  File "<string>", line 7, in <module>
  File "/opt/conda/envs/test/lib/python3.10/site-packages/numba/cuda/cudadrv/runtime.py", line 111, in get_version
    self.cudaRuntimeGetVersion(ctypes.byref(rtver))
  File "/opt/conda/envs/test/lib/python3.10/site-packages/numba/cuda/cudadrv/runtime.py", line 65, in __getattr__
    self._initialize()
  File "/opt/conda/envs/test/lib/python3.10/site-packages/numba/cuda/cudadrv/runtime.py", line 51, in _initialize
    self.lib = open_cudalib('cudart')
  File "/opt/conda/envs/test/lib/python3.10/site-packages/numba/cuda/cudadrv/libs.py", line 63, in open_cudalib
    path = get_cudalib(lib)
  File "/opt/conda/envs/test/lib/python3.10/site-packages/numba/cuda/cudadrv/libs.py", line 55, in get_cudalib
    libdir = get_cuda_paths()[dir_type].info
  File "/opt/conda/envs/test/lib/python3.10/site-packages/numba/cuda/cuda_paths.py", line 223, in get_cuda_paths
    'nvvm': _get_nvvm_path(),
  File "/opt/conda/envs/test/lib/python3.10/site-packages/numba/cuda/cuda_paths.py", line 201, in _get_nvvm_path
    candidates = find_lib('nvvm', path)
  File "/opt/conda/envs/test/lib/python3.10/site-packages/numba/misc/findlib.py", line 44, in find_lib
    return find_file(regex, libdir)
  File "/opt/conda/envs/test/lib/python3.10/site-packages/numba/misc/findlib.py", line 56, in find_file
    entries = os.listdir(ldir)
FileNotFoundError: [Errno 2] No such file or directory: '/opt/conda/envs/test/nvvm/lib64'
```

@bdice @jakirkham do you know if we're missing something here? Also cc @gmarkall for Numba, in case you have suggestions.

jakirkham:
Yeah `numba` needs `cuda-nvcc` installed

This is needed to get `nvvm`, `cudart` & compilers

pentschev:
> Yeah `numba` needs `cuda-nvcc` installed
> 
> This is needed to get `nvvm`, `cudart` & compilers

Shouldn't this be provided when Numba gets installed then? Note that it happens during a test, and I don't think Dask-CUDA should be requiring `cuda-nvcc` at runtime.

dependencies.yaml
```diff
@@ -59,6 +59,10 @@ dependencies:
               cuda: "11.8"
             packages:
               - cudatoolkit=11.8
```
bdice, line 61: Can we change all these entries like the example below? This is needed to constrain the CUDA version of the RAPIDS packages used for Python testing. (It may be possible to derive the correct constraint from a pin on `cudatoolkit` but that's going to be less direct for the environment solver.)
```suggestion
              - cuda-version=11.8
              - cudatoolkit
```

bdice (changes requested):
Requesting a small change to use `cuda-version` in `dependencies.yaml`.

jakirkham:
Similar to `ucx`, which can operate without `cudatoolkit`, `numba` doesn't require `cuda-nvcc` (except when CUDA is used). So it is up to us to install this when we need it

pentschev:
> Similar to `ucx`, which can operate without `cudatoolkit`, `numba` doesn't require `cuda-nvcc` (except when CUDA is used). So it is up to us to install this when we need it
> 
> Edit: Maybe we can handle this on the cuDF side

Numba is an optional dependency, so the user will certainly need to satisfy that when using Numba+Dask-CUDA but without RAPIDS. I'm guessing we will want to update [Numba's installation instructions for CUDA 12 too](https://numba.pydata.org/numba-doc/latest/cuda/overview.html#software)?

jakirkham:
Yeah that's a good idea. Could you please raise an issue on the Numba repo and cc me?

pentschev:
@jakirkham raised the issue in https://github.com/numba/numba/issues/9045

pentschev:
> Yeah `numba` needs `cuda-nvcc` installed
> 
> This is needed to get `nvvm`, `cudart` & compilers

Adding `cuda-nvcc` raised a different error now:

```
cuda-nvcc: Please add the `compiler("c")` and `compiler("cxx")` packages to the environment.
```

It seems that this can't be added as a runtime dependency, is that the case?

jakirkham:
Thanks Peter! 🙏

This could be worked around by adding the `c-compiler` & `cxx-compiler` packages. Alternatively [`cuda-compiler` pulls both of them in]( https://github.com/conda-forge/cuda-compiler-feedstock/blob/daa71ce4a23bac73c193f3a3794c6e6c4e7261dc/recipe/meta.yaml#L14-L16 ) for us

That said, it seems reasonable to harden this requirement in `cuda-nvcc`. Submitted upstream PR ( https://github.com/conda-forge/cuda-nvcc-feedstock/pull/21 )

pentschev:
Thanks @jakirkham , after merging of https://github.com/conda-forge/cuda-nvcc-feedstock/pull/21 all tests now pass! 😄

dependencies.yaml
```diff
@@ -46,19 +46,28 @@ dependencies:
...
+              - cudatoolkit
           - matrix:
               cuda: "11.8"
             packages:
-              - cudatoolkit=11.8
+              - cuda-version=11.8
+              - cudatoolkit
+          - matrix:
+              cuda: "12.0"
+            packages:
```
bdice, line 68: Do we need cuda-nvrtc for Numba?

gmarkall: You can get away without it if you're not linking .cu files wit Python kernels or using float16. But for full functionality it should be present.

bdice (approved):

dependencies.yaml
jakirkham: Added this suggestion below

dependencies.yaml
```diff
@@ -46,19 +46,28 @@ dependencies:
...
+              - cuda-version=12.0
+              - cuda-nvcc
```
jakirkham, line 70:
```suggestion
              - cuda-nvrtc
```

jakirkham (commented):
Thanks Peter! 🙏

This is looking pretty good. Added a minor suggestion to include `cuda-nvrtc` based on the thread below

raydouglass (approved):

dependencies.yaml
pentschev: From what I understand of @gmarkall 's suggestion we don't need `cuda-nvrtc` given we're not doing any linkage whatsoever in Dask-CUDA, thus I think we should have only `cuda-nvcc` here.

jakirkham: Ok will leave this here and we can revisit as needed

jakirkham (approved):

jakirkham:
Thanks all! 🙏

pentschev:
Thanks everyone for reviews! 😄

PR #1204, opened 2023-06-26
Remove versioneer

### LINKED ISSUE #1203: Remove versioneer
dask-cuda should drop its usage of versioneer, to align with other RAPIDS projects.

For reference/examples, there's a series of PRs named "Stop using versioneer to manage versions" across various RAPIDS repos: https://github.com/search?q=org%3Arapidsai+%22Stop+using+versioneer%22&type=pullrequests

pentschev:
Closes #1203


wence- (approved):
Thanks Peter.

pyproject.toml
```diff
@@ -3,7 +3,6 @@ build-backend = "setuptools.build_meta"
 requires = [
     "setuptools>=64.0.0",
     "tomli  ; python_version < '3.11'",
-    "versioneer>=0.24",
 ]
 
 [project]
```
raydouglass, line 8: Since `version` was removed from `setup.py`, I believe it needs to be added here instead (eg https://github.com/rapidsai/cudf/blob/branch-23.08/python/dask_cudf/pyproject.toml#L12). And I think the `dynamic` field should be removed.

Also, an entry needs to be added to `update-version.sh` to update this file's version too.

You can see that no version is being used in the wheel build: https://github.com/rapidsai/dask-cuda/actions/runs/5377443612/jobs/9755962220?pr=1204#step:4:54

raydouglass (changes requested):

pyproject.toml
pentschev: Thanks @raydouglass for catching that. I've done the changes in https://github.com/rapidsai/dask-cuda/pull/1204/commits/210937e76ce282e687f83e2e79e72397878a3424 and it looks like the build version is correct now: https://github.com/rapidsai/dask-cuda/actions/runs/5379525987/jobs/9760908314#step:4:54 .

We have `RAPIDS_DATE_STRING` being passed to the package version in [`meta.yaml`](https://github.com/rapidsai/dask-cuda/blob/branch-23.08/conda/recipes/dask-cuda/meta.yaml#L20), which seems consistent with cuDF. Is there anywhere else we need to pass it?

pentschev: I don't know if this is acceptable, but I've attempted something in https://github.com/rapidsai/dask-cuda/pull/1204/commits/937919be6a757f63880b017fa460873c95d31aac . That change will essentially change `version` in `pyproject.toml` before building the wheel and later revert it. I don't know how we would identify whether we are doing a release build (and thus not modify `pyproject.toml`) or a pre-release build that we want to do that change, any ideas?

Obviously, we would also need to update `ci/release/update-version.sh` to do the necessary changes in this file if we go ahead with that solution, but want to check first whether this is how we want to move ahead.

cc @vyasr who may have insights here.

vyasr: Inside this script you have access to a bash function that you can use: `if rapids-is-release-build; do ...; done`

pentschev: Nice, you guys thought of everything! 😄

pentschev:
I think everything is correct now, including [wheel version](https://github.com/rapidsai/dask-cuda/actions/runs/5416388566/jobs/9846081266?pr=1204#step:4:56).

Would you mind taking a look @vyasr @raydouglass ?

ci/build_python_pypi.sh
```diff
@@ -8,11 +8,36 @@ python -m pip install build --user
 export GIT_DESCRIBE_TAG=$(git describe --abbrev=0 --tags)
 export GIT_DESCRIBE_NUMBER=$(git rev-list ${GIT_DESCRIBE_TAG}..HEAD --count)
 
+# Build date for PyPI pre-releases
+export RAPIDS_VERSION_NUMBER="23.08"
+export RAPIDS_FULL_VERSION_NUMBER="${RAPIDS_VERSION_NUMBER}.00"
```
raydouglass, line 13: Can't hardcode the patch version (unless you also update via `update-version.sh` but we do not do this in any other repo).

Since during a release, `update-version.sh` is run before building, you could pull the current version from `pyproject.toml`. Then only adjust it to the pre-release version for not release builds.

pentschev: The version is being updated in https://github.com/rapidsai/dask-cuda/pull/1204/files#diff-8a0188e09c831b0e6ca8c1c51ab640efd1e519b1d7f94ab6251c9f0b90a82304R51 . I have no preference for getting the version from `pyproject.toml` or updating it with `update-version.sh` (except the latter is already done, and perhaps slightly easier). Do you have a strong preference for the `pyproject.toml` option?

vyasr: I have a weak pref for pyproject.toml, not a strong one. FWIW here's how we do it for our other wheels: https://github.com/rapidsai/gha-tools/blob/main/tools/rapids-pip-wheel-version

ci/build_python_pypi.sh
```diff
@@ -8,11 +8,36 @@ python -m pip install build --user
...
+  export PACKAGE_VERSION_NUMBER="${RAPIDS_FULL_VERSION_NUMBER}a${BUILD_DATE}"
+fi
+
 # Compute/export RAPIDS_DATE_STRING
 source rapids-env-update
 
+# Inplace sed replace; workaround for Linux and Mac
+function sed_runner() {
+    sed -i.bak ''"$1"'' $2 && rm -f ${2}.bak
+}
```
vyasr, line 27: Let's avoid proliferating the `sed_runner` function. It has problems, and we don't need it here. This script is only ever run in CI on Linux machines. Let's just use plain sed for now.

vyasr (changes requested):
Couple of small things.

ci/build_python_pypi.sh
pentschev: Done in https://github.com/rapidsai/dask-cuda/pull/1204/commits/d4d887a9780664bacf077f7a5567c7a81592289a .

pentschev:
> I have a weak pref for pyproject.toml, not a strong one. FWIW here's how we do it for our other wheels: https://github.com/rapidsai/gha-tools/blob/main/tools/rapids-pip-wheel-version

Done in https://github.com/rapidsai/dask-cuda/pull/1204/commits/4117cf4f320a678e8a12ae24189751df22797b15 .

@raydouglass @vyasr would you mind taking another look?

vyasr (approved):

pentschev:
Thanks @raydouglass @vyasr @wence- !

PR #1207, opened 2023-06-30
Clarify `memory_limit` docs

pentschev:
It's not currently clear what is the purpose of `memory_limit`/`--memory-limit`. This attempts to clarify it's used to spill from host memory to disk.

### DIFF
--- a/dask_cuda/cli.py
+++ b/dask_cuda/cli.py
@@ -87,9 +87,10 @@ def cuda():
     "--memory-limit",
     default="auto",
     show_default=True,
-    help="""Bytes of memory per process that the worker can use. Can be an integer
-    (bytes), float (fraction of total system memory), string (like ``"5GB"`` or
-    ``"5000M"``), or ``"auto"`` or 0 for no memory management.""",
+    help="""Size of the host LRU cache, which is used to determine when the worker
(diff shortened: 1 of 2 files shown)


galipremsagar (approved):

dask_cuda/cli.py
```diff
@@ -87,9 +87,10 @@ def cuda():
     "--memory-limit",
     default="auto",
     show_default=True,
-    help="""Bytes of memory per process that the worker can use. Can be an integer
-    (bytes), float (fraction of total system memory), string (like ``"5GB"`` or
-    ``"5000M"``), or ``"auto"`` or 0 for no memory management.""",
+    help="""Size of the host LRU cache, which is used to determine when the worker
+    starts spilling to disk (not available if JIT-Unspill is enabled). Can be an
```
rnyak, line 91: is there any dependency between memory-limit and device-memory-limit, that should be set? like if one cannot be bigger than other etc..

madsbk: `memory-limit` and `device-memory-limit` are independent

madsbk (approved):
Thanks @pentschev

pentschev:
Thanks everyone for reviews. @rnyak if you have further suggestions to improve phrasing, please feel free to open an issue or a PR with changes.

PR #1209, opened 2023-07-10
Use minimal Numba dependencies for CUDA 12

jakirkham:
As Numba doesn't need all of `cuda-nvcc` and only needs `cuda-nvcc-impl` & `cuda-nvrtc`, simplify the dependencies in the CUDA 12 case to make sure only the needed ones are installed.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -67,7 +67,8 @@ dependencies:
               cuda: "12.0"
             packages:
               - cuda-version=12.0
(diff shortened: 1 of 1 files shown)


raydouglass (approved):

pentschev (approved):
LGTM, thanks @jakirkham !

jakirkham:
Thanks Ray & Peter! 🙏

PR #1210, opened 2023-07-18
Revert CUDA 12.0 CI workflows to branch-23.08.

bdice:
This PR reverts changes to the branch of `shared-action-workflows` used for CUDA 12 testing. Now that https://github.com/rapidsai/shared-action-workflows/pull/101 is merged, we can revert this.

PR #1211, opened 2023-07-18
Remove RTD configuration and references to RTD page

charlesbluca:
It's been some time since we started having the RTD page redirect to the RAPIDS docs, we should be good to remove the configuration files controlling this along with some outdated references to the old page.

Note that this won't actually unpublish the RTD redirect page, to do that we would need to delete the dask-cuda project from readthedocs.org - not opposed to doing this but figured it would make more sense to make less impactful changes first.

cc @bdice @pentschev


dask_cuda/cuda_worker.py
```diff
@@ -144,8 +144,8 @@ def del_pid_file():
                 warnings.warn(
                     "When using NVLink we recommend setting a "
                     "`rmm_pool_size`.  Please see: "
-                    "https://dask-cuda.readthedocs.io/en/latest/ucx.html"
-                    "#important-notes for more details"
+                    "https://docs.rapids.ai/api/dask-cuda/nightly/ucx/ "
+                    "for more details"
```
charlesbluca, line 148: Seemed like the section the original link was pointing to doesn't exist anymore, is there any specific place on the UCX integration docs we'd like to point this to?

jakirkham:
Sorry think we updated the labels at the same time (was trying to fix the label checker). Please check and make sure we have all the labels we want to include

jakirkham:
Should we also disable RTD builds of this repo?

charlesbluca:
> Should we also disable RTD builds of this repo?

It looks like the project has been pointed at `branch-21.06` for some time now, so builds effectively haven’t been running for 2 years or so; don't see a setting to explicitly disable builds but suppose we could remove the GitHub webhook integrations which should further disable things?

At that point I would lean towards deleting the project (unless we anticipate moving back to RTD some time in the future?) so that the redirect link and other associated RTD maintainer pings for the project stop.

dask_cuda/cuda_worker.py
pentschev: This links looks like the correct one to me.

pentschev:
> At that point I would lean towards deleting the project (unless we anticipate moving back to RTD some time in the future?) so that the redirect link and other associated RTD maintainer pings for the project stop.

If we delete the project, do we run into risks of having someone else taking ownership, similar to what could happen with names in PyPI? If so, I would be strongly against that, we should retain ownership.

pentschev (approved):
LGTM. Thanks @charlesbluca !

jakirkham:
Agree with Peter. Let's keep the project. It will also keep the redirect going to RAPIDS (so if there are old links to it still out there)

bdice (approved):
Thanks!

jakirkham:
Dropped one label to fix the label checker

jakirkham (approved):

jakirkham:
CI failures are the same observed in this comment ( https://github.com/rapidsai/dask-cuda/pull/1210#issuecomment-1641042208 )

As Peter noted offline, this appear to be related to upstream Distributed changes

pentschev:
rerun tests

wence-:
> CI failures are the same observed in this comment ( [#1210 (comment)](https://github.com/rapidsai/dask-cuda/pull/1210#issuecomment-1641042208) )
> 
> As Peter noted offline, this appears to be related to upstream Distributed changes

Fixed by #1212

PR #1212, opened 2023-07-19
Ensure plugin config can be passed from worker to client

wence-:
It is possible when printing a cluster configuration that attributes of a plugin are not serializable and so cannot be passed from worker to client. To guard against this, preemptively attempt to pickle the deduced configuration and just report an unknown configuration if that fails.


pentschev (approved):
LGTM, thanks @wence- !

Issue #1219, opened 2023-08-16
Cannot import `dask_cuda` with `distributed:main`


rjzamora:
A recent PR in `distributed` (https://github.com/dask/distributed/pull/8067) has removed `distributed.worker.loads_function`, which is imported by `dask_cuda`. Now I get something like the following when I try to import `dask_cuda`:

```
...
File ~/workspace/cudf-23.10/dask-cuda/dask_cuda/proxy_object.py:22
     20 from dask.sizeof import sizeof
     21 from distributed.protocol.compression import decompress
---> 22 from distributed.worker import dumps_function, loads_function
     24 from dask_cuda.disk_io import disk_read
     26 try:

ImportError: cannot import name 'loads_function' from 'distributed.worker' (/home/nfs/rzamora/workspace/cudf-23.10/distributed/distributed/worker.py)
```

PR #1220, opened 2023-08-16
Avoid importing `loads_function` from distributed

rjzamora:
Closes https://github.com/rapidsai/dask-cuda/issues/1219

Not sure if we need the caching behavior originally included in `loads_function`, or if `pickle.loads` is sufficient. Any idea @madsbk ?

### DIFF
--- a/dask_cuda/proxy_object.py
+++ b/dask_cuda/proxy_object.py
@@ -19,7 +19,7 @@ import distributed.protocol
 import distributed.utils
 from dask.sizeof import sizeof
 from distributed.protocol.compression import decompress
-from distributed.worker import dumps_function, loads_function
+from distributed.worker import dumps_function
 
 from dask_cuda.disk_io import disk_read
 
@@ -440,7 +440,7 @@ class ProxyObject:
         pxy = self._pxy_get(copy=True)
         pxy.serialize(serializers=("pickle",))
         if pxy.subclass:
-            subclass = loads_function(pxy.subclass)
+            subclass = pickle.loads(pxy.subclass)
         else:
             subclass = ProxyObject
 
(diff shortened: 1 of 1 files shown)


jakirkham:
If the `import` is causing an issue, could we just change the nature of the `import` (`import distributed` or `import distributed.worker`) or the usage of the function (like using `distributed.worker.<...>` in calls)?

jakirkham:
Am also curious why `dumps_function` doesn't have the same issue

madsbk (approved):
Looks good, `pickle.loads` should be suffice. Actually, I think we only use the subclass feature in the tests :)

For uniformity, I suggest we also replace `dumps_function` with `pickle.dumps`.

Thanks @rjzamora.

pentschev:
> For uniformity, I suggest we also replace `dumps_function` with `pickle.dumps`.

I'm +1 on this, would you mind doing that as well @rjzamora ?

Just as a note CI hasn't been failing and this wasn't addressed yet because we're still pinned to Dask/Distributed 2023.7.1, the process of unpinning Dask in RAPIDS has been delayed a bit as the work towards https://github.com/rapidsai/integration/pull/676 is still ongoing. Given this solution is independent of the Dask version we're fine to proceed with the merge when everyone's in agreement though.

rjzamora:
> >For uniformity, I suggest we also replace dumps_function with pickle.dumps.
>
>I'm +1 on this, would you mind doing that as well @rjzamora ?

Sounds good - done.

>Just as a note CI hasn't been failing and this wasn't addressed yet because we're still pinned to Dask/Distributed 2023.7.1 ... Given this solution is independent of the Dask version we're fine to proceed with the merge when everyone's in agreement though.

Right, that is what I suspected, but thanks for confirming.

pentschev (approved):
Thanks @rjzamora !

PR #1221, opened 2023-08-21
Enable maximum pool size for RMM async allocator

pentschev:
In addition to the `release_threshold`, enable as well support for the `maximum_pool_size`. The difference between the two is that `release_threshold` will attempt to bring RMM's memory usage down to that value after the next stream synchronization, whereas `maximum_pool_size` is a hard limit enforced by RMM.

Depends on https://github.com/rapidsai/rmm/pull/1327.


quasiben (approved):

pentschev:
Thanks @quasiben !

PR #1225, opened 2023-08-25
Adapt to non-string task keys in distributed

### LINKED ISSUE #1224: Tests failing after Distributed upstream changes
Changes in https://github.com/dask/distributed/pull/8083 have broken [all nightly tests](https://github.com/rapidsai/dask-cuda/actions/runs/5972198402/job/16202373197). If I didn't overlook anything, only `test_explicit_comms.py` and `test_spill.py` tests were broken.

@madsbk would you mind looking at the explicit comms tests? I have to still look in more details of what exactly changed w.r.t. spilling but I see the size of data that gets stored in `zict.Buffer` has changed, which indicates a change in how sizes are computed or that the stringify changes really changed the underlying stored data.

wence-:
Now that keys are no longer strings there are two places we must adapt here.

1. Explicit comms must no longer manually stringify task keys before staging and intersection with the on-worker data (since that data mapping doesn't use the stringified version)
2. The `zict.File`-backed slow buffer in `DeviceHostFile` needs to translate non-string keys to string keys before writing to disk, to do this, use the same implementation that distributed uses for its own spilling buffer.

- Closes #1224

### DIFF
--- a/dask_cuda/device_host_file.py
+++ b/dask_cuda/device_host_file.py
@@ -4,7 +4,7 @@ import os
 import time
 
 import numpy
-from zict import Buffer, File, Func
+from zict import Buffer, Func
 from zict.common import ZictBase
 
 import dask
@@ -17,6 +17,7 @@ from distributed.protocol import (
     serialize_bytelist,
 )
 from distributed.sizeof import safe_sizeof
+from distributed.spill import CustomFile as KeyAsStringFile
 from distributed.utils import nbytes
 
 from .is_device_object import is_device_object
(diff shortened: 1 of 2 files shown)


pentschev (approved):
Awesome, thanks so much for working on this @wence- ! 😄

charlesbluca:
Once the next release is out, we will probably want to bump our minimum Dask pinning to 2023.8.2, as this PR now makes it such that we fail to import with the latest stable release (2023.8.1)

pentschev:
> Once the next release is out, we will probably want to bump our minimum Dask pinning to 2023.8.2, as this PR now makes it such that we fail to import with the latest stable release (2023.8.1)

I wouldn't mind this but we generally agreed to only do this at release time, the reason for that is there's a heavy load to do that across all projects. Once we have the metapackage this will become much simpler.

PR #1234, opened 2023-09-21
Increase test timeouts further to reduce CI failures

pentschev:


galipremsagar (approved):

pentschev:
Merging to see if this improves CI during nightlies. I'm not totally confident though and we should probably continue discussing and investigating existing issues in https://github.com/rapidsai/dask-cuda/issues/1236 .

PR #1255, opened 2023-10-03
Small reorganization and fixes for `test_spill`

pentschev:
Do a minor reorganization on how `client.run()` is invoked and the centralize functions to check host and disk chunks.

The failures seem related to `del` not cleaning up objects in time, thus invoke garbage collection after `del` until it memory is actually released or the test times out. Local tests seem that invoking garbage collection once or twice is enough to prevent the test from failing.

### DIFF
--- a/dask_cuda/tests/test_spill.py
+++ b/dask_cuda/tests/test_spill.py
@@ -141,7 +141,7 @@ def assert_disk_chunks(spills_to_disk, dask_worker=None):
         },
     ],
 )
-@gen_test(timeout=120)
+@gen_test(timeout=30)
 async def test_cupy_cluster_device_spill(params):
     cupy = pytest.importorskip("cupy")
     with dask.config.set(
@@ -241,7 +241,7 @@ async def test_cupy_cluster_device_spill(params):
         },
     ],
 )
-@gen_test(timeout=120)
+@gen_test(timeout=30)
 async def test_cudf_cluster_device_spill(params):
     cudf = pytest.importorskip("cudf")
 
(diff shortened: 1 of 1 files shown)


dask_cuda/tests/test_spill.py
```diff
@@ -264,26 +289,35 @@ async def test_cudf_cluster_device_spill(params):
...
+                        await client.run(
+                            delayed_worker_assert,
+                            0,
+                            0,
+                            0,
+                        )
+                    except AssertionError:
+                        gc.collect()
+                    else:
+                        break
```
madsbk, line 323: Do we want a max number of tries?

```suggestion
                for _ in range(10):
                    try:
                        await client.run(
                            delayed_worker_assert,
                            0,
                            0,
                            0,
                        )
                    except AssertionError:
                        gc.collect()
                    else:
                        return
```

pentschev: I thought about this, but then I decided in favor of a timeout, as per description: "thus invoke garbage collection after del until it memory is actually released or the test times out.".

I think timing out is overall more appropriate than an arbitrarily-defined count. What do you think?

pentschev: The timeout is defined by the `gen_test` decorator, and that reminded me we can [now set it back to the 30 seconds it was before](https://github.com/rapidsai/dask-cuda/pull/1255/commits/7a12c4338686ac08cdad97fd882491729db6309c).

madsbk: Sounds good

madsbk (approved):
LGTM, thanks @pentschev

pentschev:
Thanks @madsbk !

PR #1267, opened 2023-10-27
Generate proper, consistent nightly versions for pip and conda packages

galipremsagar:
This PR changes conda Python packages and wheels to all generate a consistent version for nightlies. The nightly version is of the form YY.MM.DDaN, where N is the number of commits from the last tag. The version is embedded in both the package metadata and in the `dask_cuda.__version__` attribute. In addition the commit hash itself is embedded into the package as `dask_cuda.__git_commit__`. These changes ensure that

1. wheels are properly considered nightlies and are treated accordingly by pip (e.g. requiring --pre for installation, not conflicting with normal releases, etc)
2. wheels and conda packages are aligned on versions so that they can be easily compared if necessary.


dask_cuda/VERSION
```diff
@@ -0,0 +1 @@
+../VERSION
```
vyasr, line 1: The Github UI makes this confusing. You don't want a file containing the text `../VERSION`, this file should be a symlink to the top-level VERSION file.

ci/build_python_pypi.sh
```diff
@@ -11,17 +14,20 @@ export GIT_DESCRIBE_NUMBER=$(git rev-list ${GIT_DESCRIBE_TAG}..HEAD --count)
...
+sed -i "/^__git_commit__/ s/= .*/= \"${commit}\"/g" "${package_name}/_version.py"
+
 # Compute/export RAPIDS_DATE_STRING
 source rapids-env-update
 
 # Update pyproject.toml with pre-release build date
-if ! rapids-is-release-build; then
-  sed -i "s/^version = \""${TOML_VERSION}".*\"/version = \""${PACKAGE_VERSION_NUMBER}"\"/g" pyproject.toml
-fi
+# if ! rapids-is-release-build; then
```
vyasr, line 28: You can delete this block now.

ci/build_python_pypi.sh
```diff
@@ -30,6 +36,6 @@ python -m build \
   .
 
 # Revert pyproject.toml pre-release build date
-if ! rapids-is-release-build; then
-  sed -i "s/^version = \""${PACKAGE_VERSION_NUMBER}"\"/version = \""${TOML_VERSION}"\"/g" pyproject.toml
-fi
+# if ! rapids-is-release-build; then
```
vyasr, line 39: Delete this block.

vyasr (changes requested):
You also need to specify that the VERSION file should be included in the built distributions. Since dask-cuda uses setuptools directly, you can specify that in [`MANIFEST.in`](https://github.com/rapidsai/dask-cuda/blob/branch-23.12/MANIFEST.in).

dask_cuda/VERSION
jakirkham: This is how GH renders symlinks unfortunately, but good to confirm that this is a symlink

vyasr: You can tell that it wasn't a symlink before based on the CI failure.

galipremsagar: Thanks for the clarification @vyasr & @jakirkham ! I think I fixed it this time.

ci/build_python_pypi.sh
galipremsagar: Done

ci/build_python_pypi.sh
galipremsagar: Done

jakirkham:
Please see this comment: https://github.com/rapidsai/dask-cuda/pull/1267#pullrequestreview-1702486818

galipremsagar:
> Please see this comment: [#1267 (review)](https://github.com/rapidsai/dask-cuda/pull/1267#pullrequestreview-1702486818)

Thanks for pointing me to this John. @vyasr Apologies I missed it the first time.

vyasr (approved):

jakirkham (approved):

raydouglass (approved):

PR #1270, opened 2023-11-07
Use new `rapids-dask-dependency` metapackage for managing dask versions

galipremsagar:
Currently dask versions are pinned as part of every release cycle and then unpinned for the next development cycle across all of RAPIDS. This introduces a great deal of churn. To centralize the dependency, we have created a metapackage to manage the required dask version and this PR introduces that metapackage as a dependency of dask-cuda.

xref: https://github.com/rapidsai/cudf/pull/14364

PR #1273, opened 2023-11-14
Add missing alpha spec

vyasr:
Without this extra spec, consumers of dask-cuda nightlies won't know that dask-cuda nightlies want to use nightlies of rapids-dask-dependency.


AyodeAwe (approved):

vyasr:
Skipping a test as per https://github.com/rapidsai/dask-cuda/pull/1270#issuecomment-1806295358

galipremsagar:
Thanks @vyasr !

PR #1275, opened 2023-11-15
Fix rapids dask dependency version

vyasr:
I tried running update-version.sh and it properly updated the version, so my guess is that there was some race condition between when the script was updated and when it was run causing the version in pyproject.toml to not be properly updated.


bdice (approved):

raydouglass (approved):
Looks like `branch-24.02` was made before this change was introduced. See https://github.com/rapidsai/dask-cuda/pull/1274

PR #1276, opened 2023-11-16
Generate pyproject.toml with dfg

vyasr:
In #1275 I erroneously assumed that updating dependencies.yaml would update pyproject.toml which is not currently the case, resulting in a mismatch between the two and the incorrect rapids-dask-dependency continuing to exist. This PR updates the repo to use dfg to generate pyproject.toml from dependencies.yaml. This PR also adds the pre-commit hook so that these two are automatically kept up-to-date going forward.


raydouglass (approved):

dependencies.yaml
```diff
@@ -99,27 +132,29 @@ dependencies:
...
           - pytest-cov
+          - ucx-py==0.36.*
+      - output_types: [conda]
+        packages:
+          - distributed-ucxx==0.35.*
           - ucx-proc=*=gpu
-          - ucx-py=0.36
-          - ucxx=0.35
-          - distributed-ucxx=0.35
+          - ucxx==0.35.*
```
raydouglass, line 157: Should be `0.36`

raydouglass (changes requested):

dependencies.yaml
vyasr: Good catch. Looks like these were previously missing in update-version.sh. I've added them and also fixed all the version specifiers there to use `==` which is PEP 508 compliant.

raydouglass (approved):

PR #1289, opened 2023-12-01
Increase Nanny close timeout for `test_spilling_local_cuda_cluster`

pentschev:
Increase Nanny close timeout for `test_spilling_local_cuda_cluster` which didn't fail in the past but failed for the first time in latest nightly run.


pentschev:
For reference, the nightly run where this failed: https://github.com/rapidsai/dask-cuda/actions/runs/7053635426/job/19201061074 .

rjzamora (commented):
Should this target 24.02?

pentschev:
> Should this target 24.02?

We're still running nightlies with 23.12, thus I think it makes sense to fix this still in the 23.12 branch.

rjzamora (approved):

PR #1291, opened 2023-12-04
Start generating conda test environments

charlesbluca:
Now that `rapids-dependency-file-generator` makes it easier to update environment files through a centralized config, does it make sense to commit these generated files directly to the repo?


pentschev (approved):
To me personally they are unnecessary, but no objections from me either.

raydouglass (approved):

PR #1294, opened 2023-12-05
Publish nightly wheels to NVIDIA index instead of PyPI

pentschev:
Nightly wheels also require `rapids-dask-dependency` which is only available in NVIDIA's PyPI index and cannot be published to PyPI as it installs Dask/Distributed from GitHub, which is forbidden by PyPI. Therefore, we're switching to publishing nightlies only to NVIDIA index as it doesn't seem external projects currently rely on nightlies. Release packages will continue to be published to PyPI.


charlesbluca (changes requested):
Looks like we also need to switch to `build_wheel.sh` in the PR workflow?

https://github.com/rapidsai/dask-cuda/blob/6de222f9763c0b267cc969ccc290d9f3a3a12b8a/.github/workflows/pr.yaml#L60

pentschev:
cc @vyasr @raydouglass

pentschev:
> Looks like the wheel build is still failing - my guess is that this is because the provided nightly version of 24.02.00a22 isn't PEP440 compliant? Perhaps we need to do something like #1279 here?

You're right, but honestly I don't know what's the "right way" to fix this. The version being generated `24.02.00a22` seems in accordance to the v[ersion defined by `rapids-dask-dependency`](https://github.com/rapidsai/rapids-dask-dependency/blob/branch-23.12/pip/rapids-dask-dependency/pyproject.toml#L12) where I took these changes from. Maybe @vyasr can point us to how it should be handled here, I believe we may also need to be compliant with release versions (which won't include `a0` suffix).

charlesbluca:
On second thought, looks like maybe what we want to do here is bump the version in dask_cuda/VERSION rather than in the pyproject file, similar to what's happening here:

https://github.com/rapidsai/cudf/blob/8f7cbe69d4c2f670b97decc63e73b08e0eef7329/ci/build_wheel.sh#L25C4-L26

charlesbluca (approved):
Thanks @pentschev!

raydouglass (approved):

ci/build_wheel.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+set -euo pipefail
+
+source rapids-configure-sccache
+source rapids-date-string
+
+package_name=dask-cuda
```
vyasr, line 9: Maybe inline this var? It's only used once.

vyasr (changes requested):
Almost there.

ci/build_python_pypi.sh
```diff
@@ -1,42 +0,0 @@
...
-# For nightlies we want to ensure that we're pulling in alphas as well. The
-# easiest way to do so is to augment the spec with a constraint containing a
-# min alpha version that doesn't affect the version bounds but does allow usage
-# of alpha versions for that dependency without --pre
-alpha_spec=''
-if ! rapids-is-release-build; then
-    alpha_spec=',>=0.0.0a0'
-fi
-
-sed -r -i "s/rapids-dask-dependency==(.*)\"/rapids-dask-dependency==\1${alpha_spec}\"/g" pyproject.toml
```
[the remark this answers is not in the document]
pentschev: Done in https://github.com/rapidsai/dask-cuda/pull/1294/commits/1b9c9288551c37aed835e74cdb72a5ab1c6df845

vyasr (approved):

.github/workflows/build.yaml
```diff
@@ -58,22 +58,23 @@ jobs:
...
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-24.02
+    with:
+      build_type: ${{ inputs.build_type || 'branch' }}
+      branch: ${{ inputs.branch }}
+      sha: ${{ inputs.sha }}
+      date: ${{ inputs.date }}
+      script: ci/build_wheel.sh
+      # Package is pure Python and only ever requires one build.
+      matrix_filter: map(select(.ARCH == "amd64" and .PY_VER == "3.10" and .CUDA_VER == "12.0.1"))
```
ajschmidt8, line 70: @pentschev, you can generalize this with the following:

```yaml
matrix_filter: map(select(.ARCH=="amd64")) | sort_by(.CUDA_VER, (.PY_VER | split(".") | map(tonumber))) | .[-1]
```

That will filter the matrix to only include `amd64` machines, sort those machines by `CUDA_VER` and `PY_VER`, and then select the last element (which will have the latest CUDA and Python version).

You can optionally take this a step further to switch from a `centos7`/`rockylinux` image to an Ubuntu image with this minor amendment:

```yaml
matrix_filter: map(select(.ARCH=="amd64")) | sort_by(.CUDA_VER, (.PY_VER | split(".") | map(tonumber))) | .[-1] + {LINUX_VER: "ubuntu22.04"}
```

Though this will require the Ubuntu version to be updated periodically as RAPIDS rotates CI image versions.

ajschmidt8: Also, if you don't care about the `CUDA_VER`, then this can be simplified to:

```yaml
matrix_filter: map(select(.ARCH=="amd64")) | sort_by((.PY_VER | split(".") | map(tonumber))) | .[-1]
```

and

```yaml
matrix_filter: map(select(.ARCH=="amd64")) | sort_by((.PY_VER | split(".") | map(tonumber))) | .[-1] + {LINUX_VER: "ubuntu22.04"}
``` 

Here's one final variation that can be used too, which sorts the entries by `PY_VER`, gets the last entry, and then overrides the `ARCH` and `LINUX_VER` values:

```yaml
matrix_filter: sort_by((.PY_VER | split(".") | map(tonumber))) | .[-1] + {ARCH: "amd64", LINUX_VER: "ubuntu22.04"}
```

jakirkham: As Dask-CUDA is a pure Python package (agnostic to OS, Python version, CUDA version, etc.), think this is just selecting one build arbitrarily

Is the idea of these suggestions just to avoid hard-coding version information in the selector?

jakirkham: Filed as issue: https://github.com/rapidsai/dask-cuda/issues/1296

PR #1300, opened 2024-01-04
Update to Dask's `shuffle_method` kwarg

pentschev:
https://github.com/dask/dask/pull/10738 has deprecated the `shuffle` kwarg in favor of `shuffle_method` which now raises a `FutureWarning`. This change transitions to the new kwarg.

Requires https://github.com/rapidsai/cudf/pull/14708 .

### DIFF
--- a/dask_cuda/explicit_comms/dataframe/shuffle.py
+++ b/dask_cuda/explicit_comms/dataframe/shuffle.py
@@ -577,7 +577,7 @@ def get_rearrange_by_column_wrapper(func):
             kw = kw.arguments
             # Notice, we only overwrite the default and the "tasks" shuffle
             # algorithm. The "disk" and "p2p" algorithm, we don't touch.
-            if kw["shuffle"] in ("tasks", None):
+            if kw["shuffle_method"] in ("tasks", None):
                 col = kw["col"]
                 if isinstance(col, str):
                     col = [col]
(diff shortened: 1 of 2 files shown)


dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -577,7 +577,7 @@ def wrapper(*args, **kwargs):
             kw = kw.arguments
             # Notice, we only overwrite the default and the "tasks" shuffle
             # algorithm. The "disk" and "p2p" algorithm, we don't touch.
-            if kw["shuffle"] in ("tasks", None):
+            if kw["shuffle_method"] in ("tasks", None):
```
rjzamora, line 580: Does the wrapper still work if the user specifies `shuffle="tasks"`? (having trouble wrapping my head around this case)

rjzamora (approved):
Thanks @pentschev

dask_cuda/explicit_comms/dataframe/shuffle.py
pentschev: I _think_ we could make it work if absolutely necessary, the way things are currently written will not work if the user specifies `shuffle=`. We have not been trying to keep backwards compatibility in Dask-CUDA whatsoever.

rjzamora: Ah okay. That's what I thought.

I wouldn't consider this to be a serious problem - I was just curious. The user should still see a warning anyway, so it's not like they are "silently" loosing explicit-comms support.

pentschev: The user will see a warning but only in mainline Dask, but in Dask-cuDF the `shuffle=` kwarg ceased to exist so that will error out.

rjzamora:
Thanks @pentschev !

PR #1301, opened 2024-01-05
Prevent double UCX initialization in `test_dgx`

pentschev:
Double initialization of UCX context may raise exceptions and cause test failures, prevent that by reseting the context after doing some initial checks.


quasiben (approved):

pentschev:
Thanks @quasiben !

PR #1303, opened 2024-01-10
refactor CUDA versions in dependencies.yaml

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/7.

Proposes splitting the `cuda-version` dependency in `dependencies.yaml` out to its own thing, separate from the bits of the CUDA Toolkit this project needs.

### Benefits of this change

* prevents accidental inclusion of multiple `cuda-version` version in environments
* reduces update effort (via enabling more use of globs like `"12.*"`)
* improves the chance that errors like "`conda` recipe is missing a dependency" are caught in CI

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -87,6 +87,10 @@ dependencies:
             packages:
               - cuda-version=11.4
           - matrix:
+              cuda: "11.5"
+            packages:
+              - cuda-version=11.5
+          - matrix:
               cuda: "11.8"
             packages:
               - cuda-version=11.8
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -7,7 +7,8 @@ files:
       arch: [x86_64]
     includes:
       - build_python
-      - cudatoolkit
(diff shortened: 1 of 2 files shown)


bdice (approved):

pentschev (approved):

dependencies.yaml
```diff
@@ -75,34 +78,33 @@ dependencies:
...
         matrices:
-          - matrix:
-              cuda: "11.2"
-            packages:
-              - cuda-version=11.2
-              - cudatoolkit
           - matrix:
               cuda: "11.4"
             packages:
               - cuda-version=11.4
```
bdice, line 88: Looks like we lost CUDA 11.5 here.
```suggestion
          - matrix:
              cuda: "11.4"
            packages:
              - cuda-version=11.4
          - matrix:
              cuda: "11.5"
            packages:
              - cuda-version=11.5
```

raydouglass (approved):

PR #1305, opened 2024-01-18
Fix get_device_memory_ids

wence-:
A recent change to the way `StringColumn`s are implemented in cudf threw up that we were never correctly determining the number of device buffers belonging to cudf columns if they had children (e.g. list and struct columns) or masks (any nullable column). Handle those cases and update the test.

### DIFF
--- a/dask_cuda/tests/test_proxify_host_file.py
+++ b/dask_cuda/tests/test_proxify_host_file.py
@@ -302,13 +302,24 @@ def test_dataframes_share_dev_mem(root_dir):
 def test_cudf_get_device_memory_objects():
     cudf = pytest.importorskip("cudf")
     objects = [
-        cudf.DataFrame({"a": range(10), "b": range(10)}, index=reversed(range(10))),
+        cudf.DataFrame(
+            {"a": [0, 1, 2, 3, None, 5, 6, 7, 8, 9], "b": range(10)},
+            index=reversed(range(10)),
+        ),
         cudf.MultiIndex(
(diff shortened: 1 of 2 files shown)


dask_cuda/tests/test_proxify_host_file.py
```diff
@@ -302,13 +302,24 @@ def test_dataframes_share_dev_mem(root_dir):
...
     res = get_device_memory_ids(objects)
-    assert len(res) == 4, "We expect four buffer objects"
+    # Buffers are:
+    # 1. int data for objects[0].a
+    # 2. mask data for objects[0].a
+    # 3. int data for objects[0].b
+    # 4. int data for objects[0].index
+    # 5. int data for objects[1].levels[0]
+    # 6. char data for objects[1].levels[1]
+    # 7. offset data for objects[1].levels[1]
```
pentschev, line 321: Should we assert their types instead of listing them as a comment?

pentschev (approved):
Looks mostly good to me, left a question whether we can make the test more robust.

dask_cuda/tests/test_proxify_host_file.py
wence-: `get_device_memory_ids` just gathers `DeviceBuffer`s which are untyped. This comment is so that a future reader of the test knows where the magic number 7 comes from.

madsbk (approved):
LGTM, thanks @wence-

dask_cuda/tests/test_proxify_host_file.py
pentschev: Sounds good, thanks.

pentschev:
Thanks @wence- !

PR #1308, opened 2024-01-26
Allow using pandas 2

### LINKED ISSUE #1306: Support Pandas version 2
Pandas 2.0.0 was released in April 2023. We should spend some effort to make this project compatible with the 2.y versions. Pandas 3 also has a dev release out, so maybe we can try for that as well now.

vyasr:
dask-cuda uses pandas for some tests, but the main reason for the pinning is that it is inherited from RAPIDS libraries (mainly cudf) that do not yet support pandas 2.0 and are the primary use case for dask-cuda. However, there is no reason dask-cuda cannot be used in other contexts, so relaxing this constraint makes sense.

Resolves #1306

### DIFF
--- a/conda/environments/all_cuda-114_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-114_arch-x86_64.yaml
@@ -18,7 +18,7 @@ dependencies:
 - numba>=0.57
 - numpy>=1.21
 - numpydoc>=1.1.0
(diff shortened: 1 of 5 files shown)


galipremsagar (approved):

wence- (approved):
With a non-blocking question about whether 1.3 is the right lower bound.

raydouglass (approved):

Issue #1311, opened 2024-02-12
Support "dataframe.query-planning" config in ``dask.dataframe``


rjzamora:
This is the Dask CUDA component of https://github.com/rapidsai/cudf/issues/15027

The "legacy" `dask.dataframe` API is now deprecated, and both dask-cudf and dask-cuda need to be updated accordingly.

PR #1312, opened 2024-02-12
Filter dd deprecation

rjzamora:
Dask CUDA must use the deprecated `dask.dataframe` API until https://github.com/rapidsai/dask-cuda/issues/1311 and https://github.com/rapidsai/cudf/issues/15027 are both closed. This means that we must explicitly filter the following deprecation warning to avoid nighlty CI failures:

```
DeprecationWarning: The current Dask DataFrame implementation is deprecated. 
In a future release, Dask DataFrame will use new implementation that
contains several improvements including a logical query planning.
The user-facing DataFrame API will remain unchanged.

The new implementation is already available and can be enabled by
installing the dask-expr library:

    $ pip install dask-expr

and turning the query planning option on:

    >>> import dask
    >>> dask.config.set({'dataframe.query-planning': True})
    >>> import dask.dataframe as dd

API documentation for the new implementation is available at
https://docs.dask.org/en/stable/dask-expr-api.html

Any feedback can be reported on the Dask issue tracker
https://github.com/dask/dask/issues 

  import dask.dataframe as dd
```

This PR adds the (temporarily) necessary warning filter.

### DIFF
--- a/dask_cuda/__init__.py
+++ b/dask_cuda/__init__.py
@@ -4,7 +4,8 @@ import warnings
 if sys.platform != "linux":
     raise ImportError("Only Linux is supported by Dask-CUDA at this time")
 
-# This warning is not specific to dask-cuda
+# This warning must be filtered until dask-expr support
+# is enabled in both dask-cudf and dask-cuda.
 # See: https://github.com/rapidsai/dask-cuda/issues/1311
 warnings.filterwarnings(
     "ignore",
--- a/dask_cuda/__init__.py
+++ b/dask_cuda/__init__.py
@@ -1,16 +1,7 @@
 import sys
-import warnings
 
 if sys.platform != "linux":
     raise ImportError("Only Linux is supported by Dask-CUDA at this time")
-
-# This warning must be filtered until dask-expr support
-# is enabled in both dask-cudf and dask-cuda.
-# See: https://github.com/rapidsai/dask-cuda/issues/1311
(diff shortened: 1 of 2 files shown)


pentschev (commented):
@rjzamora this looks ok but I want to make sure of what's the intent here, do we want to filter this warning for _all users_ or only tests? If the latter it may be more appropriate to add this in the [`pyproject.toml`](https://github.com/rapidsai/dask-cuda/blob/364eba8e6948b55dc9225dd16fdc1f7b5b0d62c4/pyproject.toml#L120-L126).

rjzamora:
Good point @pentschev - I suppose it's probably best to constrain the filter to tests for now. I'll revise, thanks!

dask_cuda/__init__.py
```diff
@@ -3,7 +3,6 @@
 if sys.platform != "linux":
     raise ImportError("Only Linux is supported by Dask-CUDA at this time")
 
-
```
pentschev, line 6: I'm surprised linting didn't complain about this. It doesn't matter so let's keep it like this.

pentschev (approved):
LGTM, thanks @rjzamora !

pentschev:
Previous run passed, I merged the latest changes just in case. @rjzamora feel free to move this to "Ready for Review" and merging soon afterwards if there's nothing more in this PR. Thanks for working on it!

PR #1323, opened 2024-03-20
Update explicit-comms for dask-expr support

rjzamora:
Makes a few ~small~ changes to explicit-comms to support dask-expr.

EDIT: The changes are no longer "small".


dask_cuda/utils.py
```diff
@@ -764,3 +764,18 @@ def get_rmm_memory_resource_stack(mr) -> list:
         if isinstance(mr, rmm.mr.StatisticsResourceAdaptor):
             return mr.allocation_counts["current_bytes"]
     return None
+
```
pentschev, line 767: Linter is failing because another blank line is missing here. I suggest installing `pre-commit` and running `pre-commit install` in each repo before committing, which will automatically fix such things for you. You can still run `pre-commit run -a` to fix linting issues and committing those changes.

pentschev (approved):
Apart from the linting issues, looks good to me. I think it should be good to go as soon as CI passes. Also tagging @madsbk for awareness.

quasiben:
I gave this a try and I'm seeing errors. It's been a while since I've run this query so it could be user error: 

python local_cudf_merge.py -p ucx -d 0,1,2,3,4,5,6,7 -c 10_000_000

```
2024-03-20 12:49:49,925 - distributed.shuffle._scheduler_plugin - WARNING - Shuffle 637a8bdfdc46dc6083f254fce7023e81 initialized by task ('shuffle-transfer-637a8bdfdc46dc6083f254fce7023e81', 0) executed on worker ucx://127.0.0.1:50385
2024-03-20 12:49:50,535 - distributed.worker - WARNING - Compute Failed
Key:       ('shuffle-transfer-637a8bdfdc46dc6083f254fce7023e81', 6)
Function:  shuffle_transfer
args:      (              key  shuffle  payload  _partitions
0        60000000        1  9459992            1
1        60000001        5  5005508            5
2        60000002        3  3183193            3
3        60000003        1   536845            1
4        60000004        3  8484645            3
...           ...      ...      ...          ...
9999995  69999995        3  9745632            3
9999996  69999996        5  8371927            5
9999997  69999997        3  8066013            3
9999998  69999998        1  3762504            1
9999999  69999999        7  9411065            7

[10000000 rows x 4 columns], '637a8bdfdc46dc6083f254fce7023e81', 6, 8, '_partitions', Empty DataFrame
Columns: [key, shuffle, payload, _partitions]
Index: [], {0, 1, 2, 3, 4, 5, 6, 7}, True, True)
kwargs:    {}
Exception: "RuntimeError('P2P shuffling 637a8bdfdc46dc6083f254fce7023e81 failed during transfer phase')"
```

And with `explicit-comms`:
> python local_cudf_merge.py -p ucx -d 0,1,2,3,4,5,6,7 -c 10_000_000 -b explicit-comms

```
2024-03-20 12:51:47,067 - distributed.worker.state_machine - WARNING - Async instruction for <Task cancelled name="execute(('shuffle-transfer-637a8bdfdc46dc6083f254fce7023e81', 5))" coro=<Worker.execute() done, defined at /datasets/bzaitlen/miniconda3/envs/dask-expr-rapid
s-24.06/lib/python3.10/site-packages/distributed/worker_state_machine.py:3615>> ended with CancelledError
  File "/datasets/bzaitlen/miniconda3/envs/dask-expr-rapids-24.06/lib/python3.10/site-packages/distributed/shuffle/_core.py", line 494, in handle_transfer_errors
    yield
  File "/datasets/bzaitlen/miniconda3/envs/dask-expr-rapids-24.06/lib/python3.10/site-packages/distributed/shuffle/_shuffle.py", line 79, in shuffle_transfer
    return get_worker_plugin().add_partition(
  File "/datasets/bzaitlen/miniconda3/envs/dask-expr-rapids-24.06/lib/python3.10/site-packages/distributed/shuffle/_worker_plugin.py", line 346, in add_partition
    return shuffle_run.add_partition(
  File "/datasets/bzaitlen/miniconda3/envs/dask-expr-rapids-24.06/lib/python3.10/site-packages/distributed/shuffle/_core.py", line 343, in add_partition
    shards = self._shard_partition(data, partition_id)
  File "/datasets/bzaitlen/miniconda3/envs/dask-expr-rapids-24.06/lib/python3.10/site-packages/distributed/shuffle/_shuffle.py", line 521, in _shard_partition
    out = split_by_worker(
  File "/datasets/bzaitlen/miniconda3/envs/dask-expr-rapids-24.06/lib/python3.10/site-packages/distributed/shuffle/_shuffle.py", line 339, in split_by_worker
    t = to_pyarrow_table_dispatch(df, preserve_index=True)
  File "/datasets/bzaitlen/miniconda3/envs/dask-expr-rapids-24.06/lib/python3.10/site-packages/dask/utils.py", line 772, in __call__
    meth = self.dispatch(type(arg))
  File "/datasets/bzaitlen/miniconda3/envs/dask-expr-rapids-24.06/lib/python3.10/site-packages/dask/utils.py", line 766, in dispatch
    raise TypeError(f"No dispatch for {cls}")
TypeError: No dispatch for <class 'cudf.core.dataframe.DataFrame'>
```

dask_cuda/utils.py
rjzamora: Yeah - I usually have pre-commit installed, so just forgot I had a new environment.

rjzamora:
>I gave this a try and I'm seeing errors.

Hmm - I actually am expecting errors in this PR. (I just realized that I never addressed the fact that we literally patch/wrap a bunch of deprecated shuffling code to automatically dispatch to explicit-comms). That said, I wasn't expecting a local-cudf-merge problem, so I'll try to reproduce.

rjzamora:
It looks like explicit-comms is going to be a pretty-big headache to migrate over to dask-expr :/

I have the basic benchmarks working locally, but there is no longer a clean way to support the `"explicit-comms"` config that we used to support by monkey-patching `dask.dataframe`. Dask-expr no longer uses a monolithic function (like `rearrange_by_coumn`) to inject a shuffle operation for sorting and joining. Instead, these operations are now a composition of abstract expressions. It may be necessary to deprecate/prohibit this behavior when query-planning is enabled (for now).

pentschev:
> It looks like explicit-comms is going to be a pretty-big headache to migrate over to dask-expr :/
> 
> I have the basic benchmarks working locally, but there is no longer a clean way to support the `"explicit-comms"` config that we used to support by monkey-patching `dask.dataframe`. Dask-expr no longer uses a monolithic function (like `rearrange_by_coumn`) to inject a shuffle operation for sorting and joining. Instead, these operations are now a composition of abstract expressions. It may be necessary to deprecate/prohibit this behavior when query-planning is enabled (for now).

Thanks for debugging this Rick. I must admit I'm not of any help when it comes to either Dask Dataframe or explicit-comms, on the other hand @madsbk may have great ideas here.

madsbk:
Hmm, maybe it is time to deprecate the monkey patching of dask and tell users to call [`shuffle()`](https://github.com/rapidsai/dask-cuda/blob/branch-24.04/dask_cuda/explicit_comms/dataframe/shuffle.py#L405) directly? 

Do you guys know many are using explicit-comms?

rjzamora:
>Hmm, maybe it is time to deprecate the monkey patching of dask and tell users to call [shuffle()](https://github.com/rapidsai/dask-cuda/blob/branch-24.04/dask_cuda/explicit_comms/dataframe/shuffle.py#L405) directly? And maybe create a basic merge() function that uses [shuffle()](https://github.com/rapidsai/dask-cuda/blob/branch-24.04/dask_cuda/explicit_comms/dataframe/shuffle.py#L405) people can use?

Yes, I think this route makes the most sense for now. I have a feeling that there are clean ways to introduce an "official" explicit-comms shuffle using the new API, but I'd like to take some time to get it right.

>Do you guys know many are using explicit-comms?

The only "official" user that I personally know of is [nemo curator](https://github.com/NVIDIA/NeMo-Curator/blob/83129e2b97b92d51641267c50e0657f934868c7c/nemo_curator/gpu_deduplication/jaccard_utils/batch_shuffle_utils.py#L45). However, (1) they are not using the monkey patching approach, and (2) they do not develop against the latest RAPIDS version. cc @VibhuJawa @ayushdg

quasiben:
I'm not aware of any external usage.  It is useful to have for benchmarking and continue exploration of shuffling generally.  With that said, I agree with @rjzamora we should take our time getting it right

dask_cuda/benchmarks/local_cudf_merge.py
```diff
@@ -25,12 +24,20 @@
...
+
+def generate_chunk(input):
+    i_chunk, local_size, num_chunks, chunk_type, frac_match, gpu = input
+
     # Setting a seed that triggers max amount of comm in the two-GPU case.
     if gpu:
         import cupy as xp
 
         import cudf as xdf
+        import dask_cudf  # noqa: F401
```
rjzamora, line 40: Generating a collection in this way puts us in a strange edge case where `dask_cudf` may never get imported on the worker, and so some type-based dispatching may fail. The only way to avoid this possibility is to import `dask_cudf` before we deploy the cluster, or to import `dask_cudf` here.

dask_cuda/tests/test_explicit_comms.py
```diff
@@ -176,6 +186,11 @@ def test_dataframe_shuffle(backend, protocol, nworkers, _partitions):
     if backend == "cudf":
         pytest.importorskip("cudf")
 
+    if QUERY_PLANNING_ON:
+        # There seem to be problems with dask_expr<1.0 and dask<2024.2.1
+        pytest.importorskip("dask", minversion="2024.2.1")
+        pytest.importorskip("dask_expr", minversion="1.0")
```
rjzamora, line 192: **We should not merge this.** This is mostly a note/demonstration that the problem has to do with a bad/old dask + dask-expr combination in CI. Since dask is "unpinned", we should definitely be getting dask_expr>1.0, but the lack of nightly packages for dask_expr can produce a bad combination for us.

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -526,23 +529,32 @@ def shuffle(
...
-    ret = new_dd_object(dsk, name, df_meta, divs).persist()
+    divs = [None] * (len(futures) + 1)
+    kwargs = {"meta": df_meta, "divisions": divs, "prefix": "explicit-comms-shuffle"}
+    try:
+        ret = dd.from_delayed(futures, **kwargs).persist()
+    except TypeError:
+        # This version of dask may not support `prefix`
+        # See: https://github.com/dask/dask-expr/pull/991
+        kwargs.pop("prefix")
+        ret = dd.from_delayed(futures, **kwargs).persist()
```
rjzamora, line 553: 991 is now merged, so dask-expr>1.0.4 will support `prefix`.

rjzamora: It should be "okay" to leave out the prefix for older versions of dask-expr: When query-planning is enabled, we are skipping the tests where we assert that the collection name starts with "explicit-comms-shuffle". That is because those tests are checking that the "explicit-comms" monkey-patching logic is working - which we do not support when query-planning is enabled anyway.

rjzamora:
CI failures without the hacky test-skip logic are due to a bad/old dask-expr version being pulled (https://github.com/dask/distributed/issues/8574)

rjzamora:
@pentschev - The upstream nightly situation should be resolved now. You approved this PR a while back when it looked pretty different. Did you have time for another pass?

ci/test_python.sh
rjzamora, line 1: **Note on these changes**: For 24.06, we want to explicitly test with **both** `DASK_DATAFRAME__QUERY_PLANNING=True` and `DASK_DATAFRAME__QUERY_PLANNING=False`.

ci/test_python.sh
```diff
@@ -58,13 +59,51 @@ timeout 60m pytest \
...
-rapids-logger "Run local benchmark"
+rapids-logger "pytest explicit-comms (legacy dd)"
+pushd dask_cuda
+DASK_DATAFRAME__QUERY_PLANNING=False \
+DASK_CUDA_TEST_SINGLE_GPU=1 \
+DASK_CUDA_WAIT_WORKERS_MIN_TIMEOUT=20 \
+UCXPY_IFNAME=eth0 \
+UCX_WARN_UNUSED_ENV_VARS=n \
+UCX_MEMTYPE_CACHE=n \
+timeout 60m pytest \
```
pentschev, line 70:
```suggestion
timeout 15m pytest \
```

pentschev: Since this is running only explicit comms we can substantially decrease the timeout. I think 15m may even be too high, but let's begin with that to be safe.

Additionally, are we switching to query-planning as default immediately or at least before 24.06 release? If not, I think we should keep on testing all with query-planning disabled, as that's really what users are gonna face.

dask_cuda/benchmarks/local_cudf_merge.py
```diff
@@ -368,9 +375,27 @@ def parse_args():
...
+            None,
+        )
+        is not False
+    ):
+        raise NotImplementedError(
+            "The 'explicit-comms' config is not yet supported when "
+            "query-planning is enabled in dask. Please use the legacy "
+            "dask-dataframe API (set the 'dataframe.query-planning' "
+            "config to `False` before executing).",
+        )
```
pentschev, line 394: It seems you're not addressing this for `local_cudf_groupby` and `local_cudf_merge`, are explicit-comms still supported for those? If not, I'd suggest moving this into [`parse_benchmark_args`](https://github.com/rapidsai/dask-cuda/blob/127de5956ffd709e97d1fdda8fba71541843b4fc/dask_cuda/benchmarks/utils.py#L317-L320) just before the function returns, and as such you'd cover all three cases.

pentschev (changes requested):
Thanks @rjzamora , left a few minor requests but otherwise it looks good.

ci/test_python.sh
rjzamora: The latest version of dask uses query-planning on by default. So, unless we are using rapids-dask-dependency to override this default, we are indeed switching to query-planning as default immediately.

I'd be happy to test everything in dask-cuda with both `DASK_DATAFRAME__QUERY_PLANNING=True` and `DASK_DATAFRAME__QUERY_PLANNING=False` for `24.06`. It's possible that some downstream users of Dask  + RAPIDS will still be opting out for now. So, that's a fair point.

dask_cuda/benchmarks/local_cudf_merge.py
rjzamora: Right - sorry. I'll use the same message for `local_cudf_groupby`. This is `local_cudf_merge`, so I assume you are also asking why I am not checking for this in `local_cudf_shuffle`? If so, it's because that benchmark does not use the dask config (it uses `dask_cuda.explicit_comms.dataframe.shuffle.shuffle` directly).

ci/test_python.sh
pentschev: If we're immediately switching then I'd say we can optimistically try not to test everything twice as the entire test set is a bit long, thus we can leave it as you wrote it and only decrease the timeout as per my suggestion above.

dask_cuda/benchmarks/local_cudf_merge.py
pentschev: > Right - sorry. I'll use the same message for `local_cudf_groupby`.

I'd still suggest simply moving this to `parse_benchmarks_args`, this will cover all benchmarks and avoid code duplication, which is essentially why we wrote it that way to begin with.

> This is `local_cudf_merge`, so I assume you are also asking why I am not checking for this in `local_cudf_shuffle`?

Yes, sorry, that is what I meant. 😅 

> If so, it's because that benchmark does not use the dask config (it uses `dask_cuda.explicit_comms.dataframe.shuffle.shuffle` directly).

But would using it directly still work with query-planning enabled? If not then moving the exception above to `parse_benchmarks_args` would also cover it and avoid more boilerplate code.

rjzamora: >I'd still suggest simply moving this to parse_benchmarks_args, this will cover all benchmarks and avoid code duplication, which is essentially why we wrote it that way to begin with.

Ah - I see what you mean. Using `dask_cuda.explicit_comms.dataframe.shuffle.shuffle` directly works fine, so it would still be good to allow that option for the shuffle benchmark somehow.

AyodeAwe (approved):

dask_cuda/benchmarks/local_cudf_merge.py
pentschev: Ok, then how about still moving this to `parse_benchmarks_args` but adding another kwarg to it that would still allow running select benchmarks (e.g., `local_cudf_shuffle`) with planning-query enabled. Something like below:

```python
def parse_benchmark_args(description="Generic dask-cuda Benchmark", args_list=[], *, force_explicit_comms=False):
...

    if args.multi_node and len(args.hosts.split(",")) < 2:
        raise ValueError("--multi-node requires at least 2 hosts")

    # Raise error early if "explicit-comms" is not allowed
    if (
        args.backend == "explicit-comms"
        and dask.config.get(
            "dataframe.query-planning",
            None,
        )
        is not False
        and force_explicit_comms is False
    ):
        raise NotImplementedError(
            "The 'explicit-comms' config is not yet supported when "
            "query-planning is enabled in dask. Please use the legacy "
            "dask-dataframe API (set the 'dataframe.query-planning' "
            "config to `False` before executing).",
        )

    return args
```

Then `local_cudf_{groupby,merge}` remain as they were and for `local_cudf_shuffle` you simply add `force_explicit_comms=True` to [this line](https://github.com/rapidsai/dask-cuda/blob/127de5956ffd709e97d1fdda8fba71541843b4fc/dask_cuda/benchmarks/local_cudf_shuffle.py#L257).

WDYT?

rjzamora: Okay, sounds good. Pushed something very similar to your solution just now.

ci/test_python.sh
rjzamora: Okay, makes sense - changed.

ci/test_python.sh
```diff
@@ -58,13 +59,51 @@ timeout 60m pytest \
...
-rapids-logger "Run local benchmark"
+rapids-logger "pytest explicit-comms (legacy dd)"
+pushd dask_cuda
+DASK_DATAFRAME__QUERY_PLANNING=False \
+DASK_CUDA_TEST_SINGLE_GPU=1 \
+DASK_CUDA_WAIT_WORKERS_MIN_TIMEOUT=20 \
+UCXPY_IFNAME=eth0 \
+UCX_WARN_UNUSED_ENV_VARS=n \
+UCX_MEMTYPE_CACHE=n \
+timeout 15m pytest \
```
pentschev, line 70:
```suggestion
timeout 30m pytest \
```

pentschev (approved):
Thanks Rick, everything looks good. I'll let you hit merge if there's nothing else here, or in case someone else still wants to review this.

PR #1325, opened 2024-03-25
Prevent path conflict in builds

AyodeAwe:
This will make builds fail when there are path conflicts


jakirkham:
Thanks Jake and Joseph! 🙏

Issue #1328, opened 2024-03-27
Broken link in explicit_comms docs


yoldemir:
Hi there!

The API link on this page appears broken: https://docs.rapids.ai/api/dask-cuda/nightly/explicit_comms/

This is the specific link that's not working: https://docs.rapids.ai/api/dask-cuda/nightly/explicit_comms/api.html

pentschev:
Fixing in https://github.com/rapidsai/dask-cuda/pull/1329 .

PR #1329, opened 2024-04-02
Fix broken links in docs

pentschev:
It was brought to our attention that some links in our docs are broken, this change should fix those issues.

Closes https://github.com/rapidsai/dask-cuda/issues/1328

### DIFF
--- a/ci/release/update-version.sh
+++ b/ci/release/update-version.sh
@@ -56,3 +56,8 @@ for FILE in .github/workflows/*.yaml; do
   sed_runner "/shared-workflows/ s/@.*/@branch-${NEXT_SHORT_TAG}/g" "${FILE}"
 done
 sed_runner "s/RAPIDS_VERSION_NUMBER=\".*/RAPIDS_VERSION_NUMBER=\"${NEXT_SHORT_TAG}\"/g" ci/build_docs.sh
+
+# Docs referencing source code
+for FILE in docs/source/*.rst; do
+    sed_runner "s/branch-${CURRENT_SHORT_TAG}\//branch-${NEXT_SHORT_TAG}\//g" "${FILE}"
+done
--- a/ci/release/update-version.sh
+++ b/ci/release/update-version.sh
@@ -58,6 +58,6 @@ done
 sed_runner "s/RAPIDS_VERSION_NUMBER=\".*/RAPIDS_VERSION_NUMBER=\"${NEXT_SHORT_TAG}\"/g" ci/build_docs.sh
 
 # Docs referencing source code
-for FILE in docs/source/*.rst; do
+find docs/source/ -type f -name *.rst -print0 | while IFS= read -r -d '' filename; do
     sed_runner "s|/branch-[^/]*/|/branch-${NEXT_SHORT_TAG}/|g" "${FILE}"
 done
--- a/ci/release/update-version.sh
+++ b/ci/release/update-version.sh
@@ -59,5 +59,5 @@ sed_runner "s/RAPIDS_VERSION_NUMBER=\".*/RAPIDS_VERSION_NUMBER=\"${NEXT_SHORT_TA
 
 # Docs referencing source code
 for FILE in docs/source/*.rst; do
-    sed_runner "s/branch-${CURRENT_SHORT_TAG}\//branch-${NEXT_SHORT_TAG}\//g" "${FILE}"
+    sed_runner "s|/branch-[^/]*/|/branch-${NEXT_SHORT_TAG}/|g" "${FILE}"
 done
--- a/ci/release/update-version.sh
+++ b/ci/release/update-version.sh
@@ -59,5 +59,5 @@ sed_runner "s/RAPIDS_VERSION_NUMBER=\".*/RAPIDS_VERSION_NUMBER=\"${NEXT_SHORT_TA
 
 # Docs referencing source code
 find docs/source/ -type f -name *.rst -print0 | while IFS= read -r -d '' filename; do
-    sed_runner "s|/branch-[^/]*/|/branch-${NEXT_SHORT_TAG}/|g" "${FILE}"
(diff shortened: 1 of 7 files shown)


ci/release/update-version.sh
```diff
@@ -56,3 +56,8 @@ for FILE in .github/workflows/*.yaml; do
   sed_runner "/shared-workflows/ s/@.*/@branch-${NEXT_SHORT_TAG}/g" "${FILE}"
 done
 sed_runner "s/RAPIDS_VERSION_NUMBER=\".*/RAPIDS_VERSION_NUMBER=\"${NEXT_SHORT_TAG}\"/g" ci/build_docs.sh
+
+# Docs referencing source code
+for FILE in docs/source/*.rst; do
+    sed_runner "s/branch-${CURRENT_SHORT_TAG}\//branch-${NEXT_SHORT_TAG}\//g" "${FILE}"
```
wence-, line 62: *question*: I believe the `main` branch always points at the current stable release branch so could one point at that? Or is this to handle nightly branches too?

docs/source/explicit_comms.rst
```diff
@@ -14,4 +14,4 @@ Usage
 In order to use explicit-comms in Dask/Distributed automatically, simply define the environment variable ``DASK_EXPLICIT_COMMS=True`` or setting the ``"explicit-comms"``
 key in the `Dask configuration <https://docs.dask.org/en/latest/configuration.html>`_.
 
-It is also possible to use explicit-comms in tasks manually, see the `API <api.html#explicit-comms>`_ and our `implementation of shuffle <https://github.com/rapidsai/dask-cuda/blob/branch-0.20/dask_cuda/explicit_comms/dataframe/shuffle.py>`_ for guidance.
+It is also possible to use explicit-comms in tasks manually, see the `API <api/#explicit-comms>`_ and our `implementation of shuffle <https://github.com/rapidsai/dask-cuda/blob/branch-24.06/dask_cuda/explicit_comms/dataframe/shuffle.py>`_ for guidance.
```
wence-, line 17: *suggestion*: Use `main` rather than nightly branch? (see comment above)

wence- (approved):

ci/release/update-version.sh
wence-: No, I think this is fine, since we want both styles of links.

docs/source/explicit_comms.rst
pentschev: > I think we can leave this. However, one thing occurs to me. We now need to go through and update this branch link with every new release, yes?

Yes, which is what [this change](https://github.com/rapidsai/dask-cuda/pull/1329/files#diff-8a0188e09c831b0e6ca8c1c51ab640efd1e519b1d7f94ab6251c9f0b90a82304R61-R63) does, this script is launched when a new branch is created.

ci/release/update-version.sh
raydouglass, line 62: It's better not to use the current tag just in case things somehow deviate.

This should work:
```suggestion
    sed_runner "s|/branch-[^/]*/|/branch-${NEXT_SHORT_TAG}/|g" "${FILE}"
```

pentschev: Ah yes, that's a much better solution, thanks Ray!

ci/release/update-version.sh
```diff
@@ -56,3 +56,8 @@ for FILE in .github/workflows/*.yaml; do
   sed_runner "/shared-workflows/ s/@.*/@branch-${NEXT_SHORT_TAG}/g" "${FILE}"
 done
 sed_runner "s/RAPIDS_VERSION_NUMBER=\".*/RAPIDS_VERSION_NUMBER=\"${NEXT_SHORT_TAG}\"/g" ci/build_docs.sh
+
+# Docs referencing source code
+for FILE in docs/source/*.rst; do
```
raydouglass, line 61: This will only list the `.rst` files directly in that directory, so `docs/source/examples/ucx.rst` for example isn't included (it doesn't need to be though).

This could be updated to use `find` instead: https://github.com/rapidsai/cuml/blob/branch-24.06/ci/release/update-version.sh#L82

raydouglass: I think?
```suggestion
find .docs/source/ -type f -name '*.rst' -print0 | while IFS= read -r -d '' FILE; do
```

pentschev: I don't think it's necessary right now but it won't hurt either, so I applied that change in https://github.com/rapidsai/dask-cuda/pull/1329/commits/19522fdc42e4685c9a30c87e6d52e7e63eaedcbe . Thanks for linking that `find` line.

ci/release/update-version.sh
```diff
@@ -56,3 +56,8 @@ for FILE in .github/workflows/*.yaml; do
   sed_runner "/shared-workflows/ s/@.*/@branch-${NEXT_SHORT_TAG}/g" "${FILE}"
 done
 sed_runner "s/RAPIDS_VERSION_NUMBER=\".*/RAPIDS_VERSION_NUMBER=\"${NEXT_SHORT_TAG}\"/g" ci/build_docs.sh
+
+# Docs referencing source code
+find docs/source/ -type f -name *.rst -print0 | while IFS= read -r -d '' filename; do
+    sed_runner "s|/branch-[^/]*/|/branch-${NEXT_SHORT_TAG}/|g" "${FILE}"
```
raydouglass, line 62:
```suggestion
    sed_runner "s|/branch-[^/]*/|/branch-${NEXT_SHORT_TAG}/|g" "${filename}"
```

raydouglass (approved):

ci/release/update-version.sh
pentschev: Oops, thanks for catching!

pentschev:
Thanks @wence- and @raydouglass for the reviews!

PR #1330, opened 2024-04-03
Trap CI test errors with their original exit codes

pentschev:
Currently, each time an error occurs the CI test script will only trap the error code an switch it to a code `1` which is not representative of the underlying error that occurred. This makes it a bit difficult to identify the source of the error, particularly in cases where the test times out (exit code `124`) and there's no summary in the log.

This change captures and prints the last error code upon exit, but also logs all errors that may have occurred throughout execution where they occurred, thus facilitating debugging.


raydouglass (approved):

pentschev:
Thanks @raydouglass !

PR #1331, opened 2024-04-07
Skip TCP-only DGX tests with UCX 1.16

pentschev:
Wireup may fail in UCX 1.16 in nodes with multiple NICs if TCP is used, thus skip those tests. UCX 1.17 will resolve the issue, and alternatively `UCX_PROTO_ENABLE=n` may be used in UCX 1.16 as well.

### DIFF
--- a/dask_cuda/tests/test_dgx.py
+++ b/dask_cuda/tests/test_dgx.py
@@ -15,6 +15,10 @@ mp = mp.get_context("spawn")  # type: ignore
 psutil = pytest.importorskip("psutil")
 
 
+def _is_ucx_116(ucp):
+    return ucp.get_ucx_version()[:2] == (1, 16)
+
+
 class DGXVersion(Enum):
     DGX_1 = auto()
     DGX_2 = auto()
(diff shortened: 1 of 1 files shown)


dask_cuda/tests/test_dgx.py
```diff
@@ -102,9 +106,13 @@ def check_ucx_options():
...
 def test_tcp_over_ucx(protocol):
     if protocol == "ucx":
-        pytest.importorskip("ucp")
+        ucp = pytest.importorskip("ucp")
     elif protocol == "ucxx":
-        pytest.importorskip("ucxx")
+        ucp = pytest.importorskip("ucxx")
+    if _is_ucx_116(ucp):
+        pytest.skip(
+            "Wireup may fail in UCX 1.16 in nodes with multiple NICs if TCP is used"
```
rjzamora, line 114: Is there an Issue we can link to for the new skip logic?

pentschev: There wasn't one, I've filed https://github.com/rapidsai/ucx-py/issues/1037 to fill that gap and changed the skip message to point to that link via https://github.com/rapidsai/dask-cuda/pull/1331/commits/b0a86ab4e7d52b4d3092c8248a70ce23e2a28fb3 .

rjzamora (approved):

pentschev:
Thanks @rjzamora !

PR #1332, opened 2024-04-08
Add snappy upper bound

pentschev:


pentschev:
This should now be fixed in new conda-forge packages, closing.

PR #1337, opened 2024-04-25
Fix license name

raydouglass:
Just a small fix to the license name for our scanning tool and to [be consistent with other repositories](https://github.com/rapidsai/cudf/blob/branch-24.06/python/cudf/pyproject.toml#L23).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ readme = { file = "README.md", content-type = "text/markdown" }
 authors = [
     { name = "NVIDIA Corporation" },
 ]
-license = { text = "Apache-2.0" }
+license = { text = "Apache 2.0" }
 requires-python = ">=3.9"
 dependencies = [
     "click >=8.1",


pentschev (approved):
LGTM, thanks @raydouglass .

pyproject.toml
```diff
@@ -13,7 +13,7 @@ readme = { file = "README.md", content-type = "text/markdown" }
 authors = [
     { name = "NVIDIA Corporation" },
 ]
-license = { text = "Apache-2.0" }
+license = { text = "Apache 2.0" }
```
jakirkham, line 16: The `-` was chosen to match [SPDX's short name]( https://spdx.org/licenses/Apache-2.0.html )

pentschev: Good point @jakirkham , should other RAPIDS projects change instead @raydouglass ? Preemptively switched to "Request changes" to prevent auto-merge until we can decide what's the right way forward.

jakirkham:
Good question. Going to ask offline

pentschev (approved):

PR #1338, opened 2024-05-06
remove 'tomli' dependency

jameslamb:
Proposes removing this project's build-time dependency on `tomli`.

It appears to no longer be necessary.

```shell
git grep tomli
```

## Notes for Reviewers

I originally noticed something similar in `ucx-py` (https://github.com/rapidsai/ucx-py/pull/1042), then went searching for similar cases across RAPIDS.

I'm not sure why this project has a dependency on `tomli`, but I suspect it was related to the use of `versioneer` in this project's history. Reference: https://github.com/python-versioneer/python-versioneer/issues/338#issuecomment-1381170813

This project doesn't use `versioneer` any more (#1204). I strongly suspect that the dependency on `tomli` can be removed.


jakirkham (approved):

raydouglass (approved):

jakirkham:
Thanks all! 🙏

PR #1343, opened 2024-06-03
use rapids-build-backend

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/31
Contributes to https://github.com/rapidsai/dependency-file-generator/issues/89

Proposes introducing `rapids-build-backend` as this project's build backend, to reduce the complexity of various CI/build scripts.


dependencies.yaml
```diff
@@ -181,3 +182,23 @@ dependencies:
               arch: aarch64
             packages:
               - numactl-devel-cos7-aarch64
+      - output_types: [requirements, pyproject]
```
jameslamb, line 185: `dependencies.yaml` here is being used to populate a `[test]` extra in `pyproject.toml` that is not ever actually used (as far as I can tell).

I've documented the (I think separate) work to set up wheel tests at some point in #1344.

For purposes of this PR, I chose to:

* break this list down into CUDA-version-specific matrices (so `rapids-build-backend` will choose the suffixed ones when populating the `test` extra on the wheels it produces)
* omit `kvikio` for the CUDA-version-specific cases for `pyproject` / `requirements`, as it doesn't currently have wheels (https://github.com/rapidsai/kvikio/pull/369)

This at least leaves the `[test]` extra on CUDA-version-specific wheels produced by `rapids-build-backend` in a state where it will resolve to real wheel packages that exist, which is an improvement over the current state.

bdice (approved):
Looks good to me. I think you did the right thing re: kvikio for now.

wence- (approved):
Approving dask-cuda-python changes.

raydouglass (approved):

PR #1349, opened 2024-06-12
make conda recipe data-loading stricter

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/72

Proposes using `[]` subsetting instead of `.get()` in templating statements in the conda recipe that read data out of `pyproject.toml`. That'll ensure that we get a big loud build error if changes to `pyproject.toml` remove some sections that the conda recipe expects to exist.

### DIFF
--- a/conda/recipes/dask-cuda/meta.yaml
+++ b/conda/recipes/dask-cuda/meta.yaml
@@ -21,7 +21,7 @@ build:
   script:
     - {{ PYTHON }} -m pip install . -vv
   entry_points:
-    {% for e in data.get("project", {}).get("scripts", {}).items() %}
+    {% for e in data["project"]["scripts"].items() %}
     - {{ e|join(" = ") }}
     {% endfor %}
 
@@ -32,7 +32,7 @@ requirements:
     - rapids-build-backend>=0.3.0,<0.4.0.dev0
   run:
     - python
-    {% for r in data.get("project", {}).get("dependencies", []) %}
+    {% for r in data["project"]["dependencies"] %}
     - {{ r }}
     {% endfor %}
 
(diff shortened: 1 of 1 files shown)


jameslamb:
Moving this back to draft, there is a failure I need to investigate. Sorry for the noise.

conda/recipes/dask-cuda/meta.yaml
```diff
@@ -21,8 +21,8 @@ build:
   script:
     - {{ PYTHON }} -m pip install . -vv
   entry_points:
-    {% for e in data.get("project", {}).get("scripts", {}).items() %}
-    - {{ e|join(" = ") }}
+    {% for entrypoint in data["project"]["scripts"] %}
+    - {{ entrypoint ~ ' = ' ~ data["project"]["scripts"][entrypoint] }}
```
jameslamb, line 25: I know this one looks a little weird. I can explain.

Switching from this

```
{% for e in data.get("project", {}).get("scripts", {}).items() %}
```

to this

```text
{% for e in data["project"]["scripts"].items() %}
```

results in an error like this:

```text
  File "/__w/dask-cuda/dask-cuda/conda/recipes/dask-cuda/meta.yaml", line 24, in top-level template code
    {% for e in data["project"]["scripts"].items() %}
TypeError: 'NoneType' object is not callable
```

([build link](https://github.com/rapidsai/dask-cuda/actions/runs/9485331325/job/26137086269))

It seems that calling dictionary methods like `.items()` or `.iteritems()` isn't supported... maybe `conda-build` overrides `.get()` or `__getattr__` or `getattr()`, or maybe that's something `Jinja2` is doing.

BUT... this pattern where you `for`-iterate over **keys**, and then use those to subset the list, totally works 😁 

And it does come with a bit more safety. I tried removing the `[project.scripts]` table in `pyproject.toml` completely. On `branch-24.08`, that results in the recipe silently building "successfully" without those entrypoints defined.

On this branch, it results in an informative error.

```text
Error: Failed to render jinja template in /Users/jlamb/repos/dask-cuda/conda/recipes/dask-cuda/meta.yaml:
'dict object' has no attribute 'scripts'
```

See "How I tested this" in the PR description for details on how I tested this.

bdice: "I can explain, I promise!" is always a good thing to write on a diff. 😉 I think this is totally fine.

bdice (approved):

conda/recipes/dask-cuda/meta.yaml
jameslamb: haha yeah, thanks 😂

PR #1352, opened 2024-06-25
Allow disabling RMM in benchmarks

pentschev:
Allows disabling RMM in benchmarks via a new option `--disable-rmm`. This change makes benchmarks a little more similar to RMM setup in `LocalCUDACluster`/`dask cuda worker`, where not specifying `rmm-pool-size` or specifying `None` as its value entirely disables setting up RMM as the default allocator. Since for benchmarks it's desired that the default is having an RMM pool we cannot change the default `--rmm-pool-size` to `None` as that would make benchmarks run much slower by default, therefore `--disable-rmm` is the closest we can make this to the rest of Dask-CUDA.

Additionally add `--rmm-maximum-pool-size` for benchmarks.

### DIFF
--- a/dask_cuda/benchmarks/utils.py
+++ b/dask_cuda/benchmarks/utils.py
@@ -17,6 +17,7 @@ from dask.utils import format_bytes, format_time, parse_bytes
 from distributed.comm.addressing import get_address_host
 
 from dask_cuda.local_cuda_cluster import LocalCUDACluster
+from dask_cuda.utils import parse_device_memory_limit
 
 
 def as_noop(dsk):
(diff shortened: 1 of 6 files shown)


dask_cuda/benchmarks/utils.py
```diff
@@ -407,56 +430,96 @@ def get_worker_device():
...
+            mr = rmm.mr.CudaAsyncMemoryResource(
+                initial_pool_size=pool_size,
+                release_threshold=release_threshold,
+            )
+
+            if maximum_pool_size is not None:
+                mr = rmm.mr.LimitingResourceAdaptor(
+                    mr, allocation_limit=maximum_pool_size
+                )
+
```
madsbk, line 493: I like this, but maybe note in the docstring that the async pool might be larger than `maximum_pool_size` because of fragmentation inside the async pool.

(I know fragmentation isn't a big issue in the async pool but it is still there.)

madsbk (approved):
Looks good, I only have a minor suggestion

ayushdg (approved):
I've tested the changes specifically for the case when I'm creating a cluster externally and want to run these benchmarks without modifying the rmm pool created in the external cluster. Can confirm that it works as expected.

dask_cuda/benchmarks/utils.py
pentschev: Just to be safe, you mean that we cannot guarantee that `maximum_pool_size` is respected because of fragmentation, i.e., if you specify `--maximum-pool-size=4GiB` it _may_ still go beyond that, is that right?

madsbk:
Yes 👍

dask_cuda/benchmarks/utils.py
pentschev: Thanks for clarifying @madsbk , I've added a note in https://github.com/rapidsai/dask-cuda/pull/1352/commits/ec097540c3ade97e184bbc9ec081bcd9453f8675 .

pentschev:
Thanks everyone!

PR #1354, opened 2024-06-28
Drop `setup.py`

jakirkham:
This is just calling `setup` from `setuptools`, which should already happen with the `setuptools` backend. So go ahead and drop `setup.py`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,0 @@
-from setuptools import setup
-
-setup()


pentschev (approved):
LGTM, thanks @jakirkham .

jakirkham:
Thank you both! 🙏

Issue #1355, opened 2024-07-03
Explicit-comms shuffle produces different partitioning than `"tasks"`


rjzamora:
Similar to the Curator bug fixed in https://github.com/NVIDIA/NeMo-Curator/pull/60, Dask-CUDA's explicit comms shuffle does not include the necessary type-casting needed for the final partitioning to match the results of `DataFrame.shuffle`.

```python
import dask
dask.config.set({"dataframe.backend": "cudf"})

from dask_cuda import LocalCUDACluster as Cluster
from distributed import Client
from dask_cuda.explicit_comms.dataframe.shuffle import shuffle as ec_shuffle

if __name__ == "__main__":

    client = Client(Cluster())
    df = dask.datasets.timeseries()
    cols = ["id"]
    npartitions = 8

    # Task-based shuffle
    df2_tasks = df.shuffle(
        cols,
        npartitions=npartitions,
        shuffle_method="tasks",
    )
    result_tasks = sorted(df2_tasks.partitions[1].compute()["id"].unique())

    # P2P shuffle
    df2_p2p = df.shuffle(
        cols,
        npartitions=npartitions,
        shuffle_method="p2p",
    )
    result_p2p = sorted(df2_p2p.partitions[1].compute()["id"].unique())

    # P2P and Task-based shuffling produce consistent results
    assert result_tasks == result_p2p  # PASSES


    # Explicit-comms shuffle
    df2_ec = ec_shuffle(
        df,
        cols,
        npartitions=npartitions,
    )
    result_ec = sorted(df2_ec.partitions[1].compute()["id"].unique())

    # Explicit-comms shuffling produce different results
    assert result_tasks == result_ec  # FAILS
```

cc @ayushdg

PR #1356, opened 2024-07-03
Fix partitioning in explicit-comms shuffle

rjzamora:
Closes https://github.com/rapidsai/dask-cuda/issues/1355

Current version of the explicit-comms shuffle does not produce partitioning that is consistent with `dask.dataframe`.

### DIFF
--- a/dask_cuda/explicit_comms/dataframe/shuffle.py
+++ b/dask_cuda/explicit_comms/dataframe/shuffle.py
@@ -8,6 +8,9 @@ from math import ceil
 from operator import getitem
 from typing import Any, Callable, Dict, List, Optional, Set, TypeVar
 
+import numpy as np
+import pandas as pd
+
 import dask
 import dask.config
 import dask.dataframe
@@ -155,9 +158,16 @@ def compute_map_index(
     if column_names[0] == "_partitions":
         ind = df[column_names[0]]
     else:
-        ind = hash_object_dispatch(
-            df[column_names] if column_names else df, index=False
-        )
+        # Need to cast numerical dtypes to be consistent
+        # with `dask.dataframe.shuffle.partitioning_index`
+        dtypes = {}
+        index = df[column_names] if column_names else df
+        for col, dtype in index.dtypes.items():
+            if pd.api.types.is_numeric_dtype(dtype):
+                dtypes[col] = np.float64
+        if dtypes:
+            index = index.astype(dtypes, errors="ignore")
+        ind = hash_object_dispatch(index, index=False)
     return ind % npartitions
 
 
@@ -187,15 +197,7 @@ def partition_dataframe(
     partitions
         Dict of dataframe-partitions, mapping partition-ID to dataframe
     """
-    if column_names[0] != "_partitions" and hasattr(df, "partition_by_hash"):
-        return dict(
-            zip(
-                range(npartitions),
-                df.partition_by_hash(
-                    column_names, npartitions, keep_index=not ignore_index
-                ),
-            )
-        )
+    # TODO: Use `partition_by_hash` after dtype-casting is added
     map_index = compute_map_index(df, column_names, npartitions)
     return group_split_dispatch(df, map_index, npartitions, ignore_index=ignore_index)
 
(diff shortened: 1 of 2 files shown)


dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -529,18 +532,19 @@ def shuffle(
...
-                )
+            _futures[part_id] = c.client.submit(
+                getitem,
+                shuffle_result[rank],
+                part_id,
+                workers=[c.worker_addresses[rank]],
             )
 
+    # Make sure partitions are properly ordered
+    futures = [_futures.pop(i) for i in range(npartitions)]
```
rjzamora, line 546: @ayushdg - **FYI**: I think this means the ordering of partitions could have been "wrong" even before the dtype-casting change in dask :/

dask_cuda/tests/test_explicit_comms.py
```diff
@@ -141,15 +147,13 @@ def _test_dataframe_shuffle(backend, protocol, n_workers, _partitions):
 
             for input_nparts in range(1, 5):
                 for output_nparts in range(1, 5):
-                    ddf = dd.from_pandas(df.copy(), npartitions=input_nparts).persist(
-                        workers=all_workers
-                    )
+                    ddf1 = dd.from_pandas(df.copy(), npartitions=input_nparts)
```
rjzamora, line 150: **Strange Finding**: Even without the changes in this PR, I get a "cancelled" error when I persist `ddf` and then perform **both** an explicit comms shuffle and then a task-based shuffle. I don't understand the cause of this yet.

dask_cuda/tests/test_explicit_comms.py
```diff
@@ -141,15 +147,13 @@ def _test_dataframe_shuffle(backend, protocol, n_workers, _partitions):
...
                     # To reduce test runtime, we change the batchsizes here instead
                     # of using a test parameter.
                     for batchsize in (-1, 1, 2):
                         with dask.config.set(explicit_comms_batchsize=batchsize):
                             ddf = explicit_comms_shuffle(
-                                ddf,
+                                ddf1,
```
rjzamora, line 156: Seems a bit confusing to me that we were previously modifying the initial collection.

dask_cuda/tests/test_explicit_comms.py
pentschev: Are we ok still moving ahead with the changes even though there's a cancelation error?

dask_cuda/tests/test_explicit_comms.py
pentschev: Makes sense to avoid that. I think @madsbk may have an idea if this was an oversight or was intentional when he gets back.

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -187,15 +197,8 @@ def partition_dataframe(
...
-        return dict(
-            zip(
-                range(npartitions),
-                df.partition_by_hash(
-                    column_names, npartitions, keep_index=not ignore_index
-                ),
-            )
-        )
+    # TODO: Use `partition_by_hash` if/when dtype-casting is added
+    # (See: https://github.com/rapidsai/dask-cuda/pull/1356)
```
pentschev, line 201: The link refers to this PR, what is there exactly to see here for additional information on this TODO?

pentschev (approved):
This seems sensible to me, thanks @rjzamora . I've left a few questions, but I don't think any of them should be blockers if you're satisfied, feel free to merge it if you think there's nothing more to be done.

dask_cuda/tests/test_explicit_comms.py
rjzamora: Yes. The changes in this PR are independent of the cancellation error (it happens in `branch-24.08` without these changes).

Another important detail: I only get the error when query-planning is disabled. Therefore, I assume the problem has to do with a key-name collision that doesn't happen with dask-expr (which is much more disciplined about key names than the legacy API is).

dask_cuda/explicit_comms/dataframe/shuffle.py
rjzamora: Ah good catch! I was planning to file an issue to add a `cast_dtype` argument to `partition_by_hash` (similar to the argument added to `partitioning_index` in https://github.com/dask/dask/pull/10705). I'll update the link before this gets merged.

rjzamora, line 201:
```suggestion
    # (See: https://github.com/rapidsai/cudf/issues/16221)
```

dask_cuda/tests/test_explicit_comms.py
madsbk: I think it was a oversight :)

PR #1357, opened 2024-07-11
Build and test with CUDA 12.5.1

KyleFromNVIDIA:
This PR updates the latest CUDA build/test version 12.2.2 to 12.5.1.

Contributes to https://github.com/rapidsai/build-planning/issues/73

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   conda-python-build:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@branch-24.08
+    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@cuda-12.5.1
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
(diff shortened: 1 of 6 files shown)


jameslamb (approved):
I think we should update this too, to encourage use of newer CUDA versions:

https://github.com/rapidsai/dask-cuda/blob/fe23e45ab4ae69be193676fe98bda383c43f9e53/docs/source/install.rst#L15-L19

I've now seen a couple like that:

https://github.com/rapidsai/raft/pull/2378#pullrequestreview-2180820360

https://github.com/rapidsai/cuvs/pull/219#pullrequestreview-2180814117

None are technically "wrong", as they say "for CUDA 12.0...", but they all use only a single minor version (e.g. there's no "for CUDA 12.2...", "for CUDA 12.3...", etc.), so I think the intent is to be up to date with the latest minor version supported across RAPIDS.

It think it'd be worth adding rules like the following to the `rapids-reviser` migration.

For any `.md` or `.rst` file:

* replace `CUDA 12.[0-9]+` with `CUDA 12.{target_version}`
* replace `cuda-version=12.[0-9]+` with `cuda-version=12.{target_version}`

PR #1360, opened 2024-07-19
Use verify-alpha-spec hook

KyleFromNVIDIA:
With the deployment of rapids-build-backend, we need to make sure our dependencies have alpha specs.

Contributes to https://github.com/rapidsai/build-planning/issues/31

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,6 +32,10 @@ repos:
                 additional_dependencies: [types-cachetools]
                 args: ["--module=dask_cuda", "--ignore-missing-imports"]
                 pass_filenames: false
+      - repo: https://github.com/rapidsai/pre-commit-hooks
+        rev: v0.3.0
+        hooks:
+            - id: verify-alpha-spec
       - repo: https://github.com/rapidsai/dependency-file-generator
         rev: v1.13.11
         hooks:


jakirkham (approved):

KyleFromNVIDIA:
I'll give the merge command now, and whoever from @rapidsai/ci-codeowners reviews can trigger the actual merge with their review.

.pre-commit-config.yaml
```diff
@@ -32,6 +32,10 @@ repos:
                 additional_dependencies: [types-cachetools]
                 args: ["--module=dask_cuda", "--ignore-missing-imports"]
                 pass_filenames: false
+      - repo: https://github.com/rapidsai/pre-commit-hooks
+        rev: v0.3.0
```
vyasr, line 36: Most of RAPIDS is still using v0.2.0 IIRC. Is it important that we go through and update?

vyasr (approved):

.pre-commit-config.yaml
KyleFromNVIDIA: v0.3.0 added a `--rapids-version` flag to the `verify-alpha-spec` hook, specifically to support `ucxx` and `ucx-py`. Most repos (including this one) don't need it, but I still use the newest version anyway.

PR #1362, opened 2024-07-23
Add arguments to enable cuDF spilling and set statistics

### LINKED ISSUE #1280: Add option to use `cudf` spilling with `dask cuda`
Currently if some one wants to enable `cudf-spilling` on a dask-cuda cluster they have to do below. 

```python3
def enable_spilling():
    import cudf
    cudf.set_option("spill", True)

cluster = LocalCUDACluster()
                    
enable_spilling()
client = Client(cluster)    
client.run(enable_spilling)
```

I dont think that is a great experience, especially given that this may not work cleanly with `jit_unspill` .  We should add a `cli` to do this to make this process easier. 

**Proposed Change:**
```python3
cluster = LocalCUDACluster(cudf_spill=True)
```

pentschev:
Add arguments to enable cuDF spilling and set statistics in `dask cuda worker`/`LocalCUDACluster`. This is implemented as a Dask plugin, and does not require users anymore to rely on `client.run` to do that.

Closes #1280

### DIFF
--- a/dask_cuda/local_cuda_cluster.py
+++ b/dask_cuda/local_cuda_cluster.py
@@ -314,6 +314,12 @@ class LocalCUDACluster(LocalCluster):
             if device_memory_limit is None and memory_limit is None:
                 data = {}
             elif jit_unspill:
+                if enable_cudf_spill:
+                    warnings.warn(
+                        "Enabling cuDF spilling and JIT-Unspill together is not "
+                        "safe, consider disabling JIT-Unspill."
+                    )
+
                 data = (
                     ProxifyHostFile,
                     {
--- a/dask_cuda/local_cuda_cluster.py
+++ b/dask_cuda/local_cuda_cluster.py
@@ -10,7 +10,7 @@ from distributed.worker_memory import parse_memory_limit
 
 from .device_host_file import DeviceHostFile
 from .initialize import initialize
-from .plugins import CPUAffinity, PreImport, RMMSetup
+from .plugins import CPUAffinity, CUDFSetup, PreImport, RMMSetup
 from .proxify_host_file import ProxifyHostFile
 from .utils import (
     cuda_visible_devices,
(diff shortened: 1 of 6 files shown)


dask_cuda/local_cuda_cluster.py
```diff
@@ -73,6 +73,14 @@ class LocalCUDACluster(LocalCluster):
         starts spilling to host memory. Can be an integer (bytes), float (fraction of
         total device memory), string (like ``"5GB"`` or ``"5000M"``), or ``"auto"``, 0,
         or ``None`` to disable spilling to host (i.e. allow full device memory usage).
+    enable_cudf_spill : bool, default False
+        Enable automatic cuDF spilling.
+
+        .. warning::
+            This should NOT be used together with JIT-Unspill.
```
rjzamora, line 80: Is there any way to detect if the user is doing this and warn/raise? This doc-string is good, but I'm sure some users will never read it :)

rjzamora (approved):
Thanks for this @pentschev ! Only one minor comment/question from me.

Side thought: We have a lot of options at this point. Do you think it makes sense to deprecate anything?

pentschev:
Warning added in https://github.com/rapidsai/dask-cuda/pull/1362/commits/d7ffa0a4b5592b1a2fa4b05876da00e45ff7d778 . @madsbk do you think we should just raise a warning or raise a warning _and_ disable JIT-Unspill regardless of the user enabling both together, i.e., `--enable-cudf-spill --enable-jit-unspill` would emit a warning saying JIT-Unspill has been disabled and only cuDF spill will be active?

> Side thought: We have a lot of options at this point. Do you think it makes sense to deprecate anything?

Probably @madsbk also has a better sense of that as well, do you have thoughts on this?

madsbk:
I think we should raise a warning only. 

> Side thought: We have a lot of options at this point. Do you think it makes sense to deprecate anything?

I think we should wait a bit with deprecating until we get a good solution from the NO-OOM work.

madsbk (approved):
Thanks @pentschev, looks good

pentschev:
Thanks @rjzamora and @madsbk for reviewing!

PR #1369, opened 2024-07-26
Replace cuDF (de)serializer with cuDF spill-aware (de)serializer

### LINKED ISSUE #1363: Allow deregistering cuDF from Dask's spilling mechanism
Using both the cuDF internal spilling mechanism and `--device-memory-limit` together may lead to excessive spilling. The more appropriate solution is to disable cuDF Dask spilling when cuDF spilling (see #1362) is enabled. There is unfortunately not Dask public API to do so and we may need to monkey patch after launch.

pentschev:
Replace cuDF (de)serializer with cuDF spill-aware (de)serializer, using both together should be avoided as that will cause excessive spilling.

Additionally add:

- Missing test of cuDF internal spill mechanism with `LocalCUDACluster`;
- `dask cuda worker` warning to alert the user that cuDF spilling mechanism requires client/scheduler to enable it as well.

Closes #1363 .


pentschev:
@ayushdg @trivialfis @dantegd @quasiben people who are using cuDF spill for whatever reason should test this, the change has potential to avoid double Dask+cuDF spilling of cuDF objects and therefore _may_ improve performance. There's also a chance that this might introduce issues that are hard to predict, so please feel free to give it a try and report any issues and/or performance results, we plan to merge it in `branch-24.10` only since we're under code freeze and this is potentially risky.

pentschev:
cc @rjzamora as well

dask_cuda/__init__.py
```diff
@@ -48,3 +50,17 @@
...
+
+
+def _register_cudf_spill_aware():
+    import cudf
+
+    # Only enable Dask/cuDF spilling if cuDF spilling is disabled
+    if not cudf.get_option("spill"):
+        from cudf.comm import serialize
+
+
```
madsbk, line 62: It would be good with a comment here saying why this it needed and include a link to https://github.com/rapidsai/dask-cuda/issues/1363

madsbk (changes requested):
Looks good @pentschev, I only have one suggestion

dask_cuda/__init__.py
pentschev: Done via https://github.com/rapidsai/dask-cuda/pull/1369/commits/29751de83cb081a357d628136c7beaac2b9b9805

madsbk (approved):

pentschev:
Thanks for review and approval @madsbk !

PR #1371, opened 2024-07-30
[Benchmark] Add  parquet read benchmark

rjzamora:
Adds new benchmark for parquet read performance using a `LocalCUDACluster`. The user can pass in `--key` and `--secret` options to specify S3 credentials.

E.g.
```
$ python ./local_read_parquet.py --devs 0,1,2,3,4,5,6,7 --filesystem fsspec --type gpu --file-count 48 --aggregate-files

Parquet read benchmark
--------------------------------------------------------------------------------
Path                      | s3://dask-cudf-parquet-testing/dedup_parquet
Columns                   | None
Backend                   | cudf
Filesystem                | fsspec
Blocksize                 | 244.14 MiB
Aggregate files           | True
Row count                 | 372066
Size on disk              | 1.03 GiB
Number of workers         | 8
================================================================================
Wall clock                | Throughput
--------------------------------------------------------------------------------
36.75 s                   | 28.78 MiB/s
21.29 s                   | 49.67 MiB/s
17.91 s                   | 59.05 MiB/s
================================================================================
Throughput                | 41.77 MiB/s +/- 7.81 MiB/s
Bandwidth                 | 0 B/s +/- 0 B/s
Wall clock                | 25.32 s +/- 8.20 s
================================================================================
...
```

**Notes**:
- S3 Performance generally scales with the number of workers (multiplied the number of threads per worker)
- The example shown above was not executed from an EC2 instance
- The example shown above *should* perform better after https://github.com/rapidsai/cudf/pull/16657
- Using `--filesystem arrow` together with `--type gpu` performs well, but depends on https://github.com/rapidsai/cudf/pull/16684


pentschev:
> Performance generally scales with the number of workers (multiplied the number of threads per worker)

I'm assuming this apply to CPU-only operations, or are there CUDA kernels executed as part of this as well?

rjzamora:
>I'm assuming this apply to CPU-only operations, or are there CUDA kernels executed as part of this as well?

This benchmark is entirely IO/CPU bound. There is effectively no CUDA compute - we are just transferring remote data into host memory and moving it into device memory (when the default `--type gpu` is used). Therefore, increasing `threads_per_worker * n_workers` typically improves performance (because we have more threads making connections and sending requests to S3).

dask_cuda/benchmarks/custom/parquet.py
```diff
@@ -0,0 +1,169 @@
...
+
+        if not hasattr(worker, "_rapids_executor"):
+            fragment_parallelism = fragment_parallelism or 8
+            num_threads = min(
+                fragment_parallelism,
+                len(os.sched_getaffinity(0)),
+            )
+            worker._rapids_executor = ThreadPoolExecutor(num_threads)
+        with dask.config.set(pool=worker._rapids_executor):
+            return dask.threaded.get(dsk, chunk_name)
```
pentschev, line 104: Does this mean that only this block will execute with multiple CPU threads?

rjzamora: My intention is that the task responsible for creating a single cudf partition will use multiple CPU threads in a local thread executor to read the arrow fragments into host memory from S3. I'm not sure of the best way to accomplish this within a single task.

pentschev: I don't know what is the best way either, I'm just asking whether my understanding of this code block is correct, if it is it seems like a sensible solution though.

dask_cuda/benchmarks/custom/parquet.py
```diff
@@ -0,0 +1,169 @@
...
+from pyarrow import dataset
+
+import dask
+import dask.dataframe as dd
+from dask.base import apply, tokenize
+from dask.distributed import get_worker
+from dask.utils import parse_bytes
+
+# NOTE: The pyarrow component of this code was mostly copied
+# from dask-expr (dask_expr/io/parquet.py)
```
wence-, line 17: Should reproduce the copyright notice...

dask_cuda/benchmarks/remote_parquet.py
```diff
@@ -0,0 +1,179 @@
...
+from dask_cuda.benchmarks.utils import (
+    parse_benchmark_args,
+    print_key_value,
+    print_separator,
+    print_throughput_bandwidth,
+)
+
+DEFAULT_DATASET_PATH = "s3://dask-cudf-parquet-testing/dedup_parquet"
+DEFAULT_COLUMNS = ["text", "id"]
+DEFAULT_STORAGE_SIZE = 2_843_373_145  # Compressed byte size
```
rjzamora, line 22: TODO: The user should be able to set these values with cli args.

dask_cuda/benchmarks/local_read_parquet.py
pentschev, line 1: I'd suggest renaming to just `read_parquet.py`. The `local_` prefix is a historical leftover from when we supported running only with `LocalCUDACluster`, we should in fact get rid of that prefix.

rjzamora:
**Update**: I've generalized this benchmark. It's easy to use with S3 storage, but is also a useful benchmark for local-storage performance.

dask_cuda/benchmarks/local_read_parquet.py
```diff
@@ -0,0 +1,278 @@
...
+        },
+        {
+            "name": "--aggregate-files",
+            "default": False,
+            "action": "store_true",
+            "help": "How to set the aggregate_files option",
+        },
+        {
+            "name": "--path",
+            "default": "s3://dask-cudf-parquet-testing/dedup_parquet",
```
pentschev, line 206: Is that valid and publicly accessible? If not I'd suggest making the path a required argument without a default, this default is misleading to users.

dask_cuda/benchmarks/local_read_parquet.py
```diff
@@ -0,0 +1,278 @@
...
+            "type": str,
+            "help": "Filesystem backend",
+        },
+        {
+            "name": "--runs",
+            "default": 3,
+            "type": int,
+            "help": "Number of runs",
+        },
+        # NOTE: The following args are not relevant to this benchmark
```
pentschev, line 253: Then why don't we remove it?

dask_cuda/benchmarks/local_read_parquet.py
```diff
@@ -0,0 +1,278 @@
...
+        fs = pa_fs.FileSystem.from_uri(args.path)[0]
+        try:
+            region = {"region": fs.region}
+        except AttributeError:
+            region = {}
+        kwargs["filesystem"] = type(fs)(**region, **s3_args)
+        fsspec_fs = ArrowFSWrapper(kwargs["filesystem"])
+
+        if args.type == "gpu":
+            kwargs["blocksize"] = args.blocksize
```
pentschev, line 69: Why doesn't blocksize impact CPU?

pentschev (commented):
Thanks @rjzamora , I've left some comments.

dask_cuda/benchmarks/local_read_parquet.py
rjzamora: This is technically a "bug" in the benchmark infrastructure. Removing this block results in an error for now. It probably makes sense to fix the bug in this PR (once every thing else is ironed out).

dask_cuda/benchmarks/local_read_parquet.py
rjzamora: It is not supported for `dd.read_parquet(..., filesystem="arrow")` when the default ("parquet") backend is active. For the "cudf" backend, https://github.com/rapidsai/cudf/pull/16684 will utilize `blocksize` to control file aggregation, but the conventional splitting of large files is generally not supported when ` filesystem="arrow"`.

dask_cuda/benchmarks/local_read_parquet.py
rjzamora: Aha, I forgot that you can technically specify a non-local cluster. Thanks for the suggestion - I'll remove the `local_` prefix.

dask_cuda/benchmarks/local_read_parquet.py
rjzamora: Yeah that's fair, I was planning to point this to a public bucket before marking the PR as ready to review, but I forgot (and I'm not sure if I know of a good public dataset to use).

I suppose it would be fine to require this argument (as you suggest).

dask_cuda/benchmarks/utils.py
```diff
@@ -337,6 +337,13 @@ def parse_benchmark_args(
         "If the files already exist, new files are created with a uniquified "
         "BASENAME.",
     )
+    parser.add_argument(
+        "--ignore-size",
+        default="1 MiB",
+        metavar="nbytes",
+        type=parse_bytes,
+        help="Ignore messages smaller than this (default '1 MB')",
```
madsbk, line 345:
```suggestion
        help="Bandwidth statistics: ignore messages smaller than this (default '1 MB')",
```

madsbk (approved):
Nice @rjzamora, looks good. I only have a minor suggestion.

PR #1375, opened 2024-08-19
Remove NumPy <2 pin

seberg:
This PR removes the NumPy<2 pin which is expected to work for
RAPIDS projects once CuPy 13.3.0 is released (CuPy 13.2.0 had
some issues preventing the use with NumPy 2).

### DIFF
--- a/conda/environments/all_cuda-114_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-114_arch-x86_64.yaml
@@ -16,7 +16,7 @@ dependencies:
 - kvikio==24.10.*,>=0.0.0a0
 - numactl-devel-cos7-x86_64
 - numba>=0.57
-- numpy>=1.23,<2.0a0
+- numpy>=1.23,<3
 - numpydoc>=1.1.0
 - pandas>=1.3
 - pre-commit
--- a/conda/environments/all_cuda-114_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-114_arch-x86_64.yaml
@@ -16,7 +16,7 @@ dependencies:
 - kvikio==24.10.*,>=0.0.0a0
 - numactl-devel-cos7-x86_64
 - numba>=0.57
-- numpy>=1.23,<3
+- numpy>=1.23,<3.0a0
 - numpydoc>=1.1.0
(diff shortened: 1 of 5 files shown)


jakirkham:
Closing and reopening to rerun CI with CuPy 13.3.0 (since it was released earlier today)

jakirkham:
It is worth noting that there is a cycle in the dependency graph between [cuDF requiring Dask-CUDA in its tests]( https://github.com/rapidsai/cudf/blob/83f68c920f51f9e69f2a5bf0fddf26babac2483b/dependencies.yaml#L756-L761 ) and [Dask-CUDA requiring cuDF]( https://github.com/rapidsai/dask-cuda/blob/49ebabc5b76db2c1f1b9bbf89774689d69d129ee/dependencies.yaml#L159-L167 )

Unfortunately that means to test either one with NumPy 2, we need to rebuild both with NumPy 2 first. Since it is not possible to rebuild both and test at the same time, we need to pick one to rebuild first so we can test the other

Given Dask-CUDA's library usage of NumPy 2 is relatively simple (wrapping memory and very minor adjustments), rebuilding Dask-CUDA first seems like the safer path. We can then rebuild and test cuDF with NumPy 2 and Dask-CUDA. Once cuDF packages are out we can come back and test Dask-CUDA on NumPy 2 (using a no-op PR to run CI tests and spot checking CI logs for NumPy 2 usage with cuDF)

Leaving this note for completeness about this plan. Will also link the no-op testing PR here

jakirkham:
Should add, ran a quick sanity check using [Ruff's NumPy 2 plugin]( https://numpy.org/devdocs/numpy_2_0_migration_guide.html#ruff-plugin ) and it came back clean

```
% ruff check dask_cuda --select NPY201
All checks passed!
```

So hopefully that provides us some more confidence (in addition to repeated offline testing done as part of NumPy 2 bringup, CuPy 13.3.0 development, etc.)

jakirkham (approved):
Thanks Sebastian! 🙏

jakirkham:
Will use PR ( https://github.com/rapidsai/dask-cuda/pull/1378 ) for testing

PR #1383, opened 2024-09-10
Add notes on cudf spilling to docs

rjzamora:
Updates the dask-cuda documentation to include notes on native cuDF spilling, since it is often the best spilling approach for ETL with Dask cuDA (please feel free to correct me if I'm wrong).

### DIFF
--- a/docs/source/examples/best-practices.rst
+++ b/docs/source/examples/best-practices.rst
@@ -44,6 +44,14 @@ We also recommend allocating most, though not all, of the GPU memory space. We d
 
 Additionally, when using `Accelerated Networking`_ , we only need to register a single IPC handle for the whole pool (which is expensive, but only done once) since from the IPC point of viewer there's only a single allocation. As opposed to just using RMM without a pool where each new allocation must be registered with IPC.
 
+Spilling from Device
+~~~~~~~~~~~~~~~~~~~~
+
+Dask CUDA offers several different ways to enable automatic spilling from device memory.
+The best method often depends on the specific workflow. For classic ETL workloads with
+`Dask cuDF <https://docs.rapids.ai/api/dask-cudf/stable/>`_, cuDF spilling is usually
+the best place to start. See `spilling`_ for more details.
+
 Accelerated Networking
 ~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/source/examples/best-practices.rst
+++ b/docs/source/examples/best-practices.rst
@@ -48,8 +48,8 @@ Spilling from Device
 ~~~~~~~~~~~~~~~~~~~~
 
 Dask-CUDA offers several different ways to enable automatic spilling from device memory.
-The best method often depends on the specific workflow. For classic ETL workloads with
-`Dask-cuDF <https://docs.rapids.ai/api/dask-cudf/stable/>`_, cuDF spilling is usually the
+The best method often depends on the specific workflow. For classic ETL workloads using
+`Dask cuDF <https://docs.rapids.ai/api/dask-cudf/stable/>`_, cuDF spilling is usually the
 best place to start. See :ref:`Spilling from device <spilling-from-device>` for more details.
 
 Accelerated Networking
--- a/docs/source/examples/best-practices.rst
+++ b/docs/source/examples/best-practices.rst
@@ -49,8 +49,8 @@ Spilling from Device
 
 Dask-CUDA offers several different ways to enable automatic spilling from device memory.
 The best method often depends on the specific workflow. For classic ETL workloads with
-`Dask cuDF <https://docs.rapids.ai/api/dask-cudf/stable/>`_, cuDF spilling is usually
-the best place to start. See `spilling`_ for more details.
+`Dask-cuDF <https://docs.rapids.ai/api/dask-cudf/stable/>`_, cuDF spilling is usually the
+best place to start. See :ref:`Spilling from device <spilling-from-device>` for more details.
 
 Accelerated Networking
 ~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/source/examples/best-practices.rst
+++ b/docs/source/examples/best-practices.rst
@@ -47,7 +47,7 @@ Additionally, when using `Accelerated Networking`_ , we only need to register a 
 Spilling from Device
 ~~~~~~~~~~~~~~~~~~~~
 
-Dask CUDA offers several different ways to enable automatic spilling from device memory.
+Dask-CUDA offers several different ways to enable automatic spilling from device memory.
 The best method often depends on the specific workflow. For classic ETL workloads with
 `Dask cuDF <https://docs.rapids.ai/api/dask-cudf/stable/>`_, cuDF spilling is usually
 the best place to start. See `spilling`_ for more details.
--- a/docs/source/spilling.rst
+++ b/docs/source/spilling.rst
@@ -105,3 +105,68 @@ type checking doesn't:
 Thus, if encountering problems remember that it is always possible to use ``unproxy()``
 to access the proxied object directly, or set ``DASK_JIT_UNSPILL_COMPATIBILITY_MODE=True``
 to enable compatibility mode, which automatically calls ``unproxy()`` on all function inputs.
+
+
+cuDF Spilling
+-------------
+
+When executing a `Dask cuDF <https://docs.rapids.ai/api/dask-cudf/stable/>`_
+(i.e. Dask DataFrame) ETL workflow, it is usually best to leverage `native spilling support in
+cuDF <https://docs.rapids.ai/api/cudf/stable/developer_guide/library_design/#spilling-to-host-memory>`.
+
+Native cuDF spilling has an important advantage over the other methodologies mentioned
+above. When JIT-unspill or default spilling are used, the worker is only able to spill
+the input or output of a task. This means that any data that is created within the task
+is completely off limits until the task is done executing. When cuDF spilling is used,
+however, individual device buffers can be spilled/unspilled as needed while the task
+is executing.
+
+When deploying a ``LocalCUDACluster``, cuDF spilling can be enabled with the ``enable_cudf_spill`` argument:
+
+.. code-block::
+
+    >>> from distributed import Client​
+    >>> from dask_cuda import LocalCUDACluster​
+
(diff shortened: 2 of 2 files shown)


docs/source/examples/best-practices.rst
```diff
@@ -44,6 +44,14 @@ We also recommend allocating most, though not all, of the GPU memory space. We d
 
 Additionally, when using `Accelerated Networking`_ , we only need to register a single IPC handle for the whole pool (which is expensive, but only done once) since from the IPC point of viewer there's only a single allocation. As opposed to just using RMM without a pool where each new allocation must be registered with IPC.
 
+Spilling from Device
+~~~~~~~~~~~~~~~~~~~~
+
+Dask CUDA offers several different ways to enable automatic spilling from device memory.
+The best method often depends on the specific workflow. For classic ETL workloads with
+`Dask cuDF <https://docs.rapids.ai/api/dask-cudf/stable/>`_, cuDF spilling is usually
+the best place to start. See `spilling`_ for more details.
```
rjzamora, line 53: TODO: Fix this "spilling" link to actually point to the [spilling](https://docs.rapids.ai/api/dask-cuda/nightly/spilling/) page (wasn't sure how to do this off the top of my head).

docs/source/examples/best-practices.rst
```diff
@@ -44,6 +44,14 @@ We also recommend allocating most, though not all, of the GPU memory space. We d
 
 Additionally, when using `Accelerated Networking`_ , we only need to register a single IPC handle for the whole pool (which is expensive, but only done once) since from the IPC point of viewer there's only a single allocation. As opposed to just using RMM without a pool where each new allocation must be registered with IPC.
 
+Spilling from Device
+~~~~~~~~~~~~~~~~~~~~
+
+Dask CUDA offers several different ways to enable automatic spilling from device memory.
```
pentschev, line 50:
```suggestion
Dask-CUDA offers several different ways to enable automatic spilling from device memory.
```

docs/source/examples/best-practices.rst
```diff
@@ -44,6 +44,14 @@ We also recommend allocating most, though not all, of the GPU memory space. We d
...
+The best method often depends on the specific workflow. For classic ETL workloads with
+`Dask cuDF <https://docs.rapids.ai/api/dask-cudf/stable/>`_, cuDF spilling is usually
```
pentschev, line 52:
```suggestion
`Dask-cuDF <https://docs.rapids.ai/api/dask-cudf/stable/>`_, cuDF spilling is usually
```

docs/source/spilling.rst
```diff
@@ -105,3 +105,68 @@ type checking doesn't:
 Thus, if encountering problems remember that it is always possible to use ``unproxy()``
 to access the proxied object directly, or set ``DASK_JIT_UNSPILL_COMPATIBILITY_MODE=True``
 to enable compatibility mode, which automatically calls ``unproxy()`` on all function inputs.
+
+
+cuDF Spilling
+-------------
+
+When executing a `Dask cuDF <https://docs.rapids.ai/api/dask-cudf/stable/>`_
```
pentschev, line 113:
```suggestion
When executing a `Dask-cuDF <https://docs.rapids.ai/api/dask-cudf/stable/>`_
```

docs/source/spilling.rst
```diff
@@ -105,3 +105,68 @@ type checking doesn't:
...
+Statistics
+~~~~~~~~~~
+
+When cuDF spilling is enabled, it is also possible to have cuDF collect basic
+spill statistics. This information can be a useful way to understand the
+performance of Dask cuDF workflows with high memory utilization:
+
+.. code-block::
+
+    $ dask cuda worker --enable-cudf-spill --cudf-spill-stats 1
```
pentschev, line 153: Maybe we should add a comment with a link to https://docs.rapids.ai/api/cudf/stable/developer_guide/library_design/#statistics which defines what each spilling statistics level means.

docs/source/spilling.rst
```diff
@@ -105,3 +105,68 @@ type checking doesn't:
...
+
+To have each dask-cuda worker print spill statistics, do something like:
+
+.. code-block::
+
+    def spill_info():
+        from cudf.core.buffer.spill_manager import get_global_manager
+        print(get_global_manager().statistics)
+    client.submit(spill_info)
```
pentschev, line 162: Why writing this if the above does the same? Similarly `LocaCUDACluster(..., cudf_spill_stats=...)` does the same, so this seems redundant and I would discourage people to do that manually.

pentschev: In fact, it seems we should document `LocaCUDACluster(..., cudf_spill_stats=...)` here instead.

docs/source/spilling.rst
```diff
@@ -105,3 +105,68 @@ type checking doesn't:
...
+
+
+Limitations
+~~~~~~~~~~~
+
+Although cuDF spilling is the best option for most Dask cuDF ETL workflows,
```
pentschev, line 168:
```suggestion
Although cuDF spilling is the best option for most Dask-cuDF ETL workflows,
```

docs/source/spilling.rst
```diff
@@ -105,3 +105,68 @@ type checking doesn't:
...
+it will be much less effective if that workflow converts between ``cudf.DataFrame``
+and other data formats (e.g. ``cupy.ndarray``). Once the underlying device buffers
+are "exposed" to external memory references, they become "unspillable" by cuDF.
+In cases like this (e.g. Dask CUDA + XGBoost), JIT-Unspill is usually a better choice.
```
pentschev, line 172:
```suggestion
In cases like this (e.g., Dask-CUDA + XGBoost), JIT-Unspill is usually a better choice.
```

docs/source/examples/best-practices.rst
rjzamora: What is the "grammar" rule on hyphenating software/project names? I faintly remember discussing this with @jacobtomlinson at one point, and I though I left with the sense that the hyphen shouldn't be there. However, there is a good chance I'm remembering incorrectly.

Should we add the hyphen back across all the Dask cuDF documentation? (https://github.com/rapidsai/cudf/pull/16765)

docs/source/spilling.rst
rjzamora: How do you print the spill statistics after using `LocaCUDACluster(..., cudf_spill_stats=...)`?

This was meant to be independent from how you enable statistics collection, but happy to change this to whatever you recommend.

docs/source/examples/best-practices.rst
pentschev: I don't know about the grammar rule, but technically this is a name and I wouldn't think any grammar necessarily applies. We've been using Dask-CUDA exclusively so it matches the project/repo name `dask-cuda`, although I saw now that the most obvious place, [README.md](https://github.com/rapidsai/dask-cuda/blob/branch-24.10/README.md#dask-cuda), actually spells `Dask CUDA` but that's the only place.

For the reason stated above I'd prefer to keep `Dask-CUDA`, but if there are strong/compelling reasons to change I'd be open for that. As for `Dask cuDF` I won't be too opinionated, but to me it feels like it makes more sense to match the project name `dask-cudf`, thus with a hyphen.

docs/source/spilling.rst
pentschev: Oh, my mistake, I overlooked the "print" part. Thus I agree we should keep it, but let's also add the `LocalCUDACluster` example here as well.

docs/source/examples/best-practices.rst
rjzamora: Yeah, sorry. I probably should have left this comment on one of the `Dask cuDF -> Dask-cuDF` suggestions. I don't think there is actually a clear "grammar" rule to point to. It should be fine to just choose a name like "Dask-CUDA", and use it everywhere. It looks slightly funny to my eye, but I have no problem using it.

For `Dask cuDF`, I do have a small preference to leave out the hyphen (it's more pleasing to my eye and is more consistent with Dask DataFrame is written). I think the project name means much, since it's not really possible to use an empty space in that context anyway.

pentschev: In the case of Dask-CUDA to me it feels more natural with the hyphen as well to identify we're talking about the project itself, and not using CUDA with Dask which is perfectly doable too.

pentschev (approved):
Changes look good to me. I've seen you've renamed `Dask-cuDF`->`Dask cuDF` in https://github.com/rapidsai/cudf/pull/16765. I'd thus consider that the "official name", since we refer to it here I'd say it's best to switch from `Dask-cuDF` to `Dask cuDF` in all instances here too. Sorry for making you change this before, I didn't previously realize you were removing the hyphen in https://github.com/rapidsai/cudf/pull/16765 , I'll nevertheless leave the last choice on `Dask cuDF` naming in here up to you.

rjzamora:
>Changes look good to me. I've seen you've renamed Dask-cuDF->Dask cuDF in https://github.com/rapidsai/cudf/pull/16765. I'd thus consider that the "official name", since we refer to it here I'd say it's best to switch from Dask-cuDF to Dask cuDF in all instances here too. Sorry for making you change this before, I didn't previously realize you were removing the hyphen in https://github.com/rapidsai/cudf/pull/16765 , I'll nevertheless leave the last choice on Dask cuDF naming in here up to you.

Thanks @pentschev ! I'll admit I'm a bit torn on the hyphen issue - My eye prefers a space, but there are definitely many cases in the wild where the hyphen is present, and I do want to be as consistent as possible. I suppose I'll just start changing  Dask-cuDF to Dask cuDF whenever I come across it from now on (and Dask CUDA to Dask-CUDA).

pentschev:
> Thanks @pentschev ! I'll admit I'm a bit torn on the hyphen issue - My eye prefers a space, but there are definitely many cases in the wild where the hyphen is present, and I do want to be as consistent as possible. I suppose I'll just start changing Dask-cuDF to Dask cuDF whenever I come across it from now on (and Dask CUDA to Dask-CUDA).

That makes sense, since we're already using `Dask-CUDA` everywhere in here I'd prefer to keep it that way, but by no means I have an ultimate opinion on that, if there's strong desire from the broader group to change it then I'm fine if we make that consistent everywhere.

PR #1396, opened 2024-10-10
Reenable UCXX in CI

pentschev:
UCXX CI tests had been previously disabled due to instabilities, see https://github.com/rapidsai/dask-cuda/pull/1270#issuecomment-1806295358, it should now be much more resilient so we should reenable them in preparation for the permanent migration to UCXX.


AyodeAwe (approved):

pentschev:
Thanks @AyodeAwe !

PR #1401, opened 2024-10-23
Define "GPU" as a worker resource

pentschev:
Add "GPU" as a worker resource to each CUDA worker. This should support users in identifying whether the workers available contain a GPU resource.

### DIFF
--- a/dask_cuda/cuda_worker.py
+++ b/dask_cuda/cuda_worker.py
@@ -6,7 +6,7 @@ import logging
 import os
 import warnings
 
-from toolz import valmap
+from toolz import itemfilter, valmap
 
 import dask
 from distributed import Nanny
@@ -103,11 +103,15 @@ class CUDAWorker(Server):
             atexit.register(del_pid_file)
 
         if resources:
-            resources = resources.replace(",", " ").split()
+            resources = resources.replace(",", " ").replace("'", "").split()
             resources = dict(pair.split("=") for pair in resources)
-            resources = valmap(float, resources)
+            gpu_resources = valmap(int, itemfilter(lambda x: x != "GPU", resources))
+            resources = valmap(float, itemfilter(lambda x: x == "GPU", resources))
+            resources.update(gpu_resources)
+            if "GPU" not in resources:
+                resources["GPU"] = 1
         else:
-            resources = None
+            resources = {"GPU": 1}
 
         preload_argv = kwargs.pop("preload_argv", [])
         kwargs = {"worker_port": None, "listen_address": None, **kwargs}
--- a/dask_cuda/local_cuda_cluster.py
+++ b/dask_cuda/local_cuda_cluster.py
@@ -23,8 +23,8 @@ from .utils import (
 
 
 class LoggedWorker(Worker):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, *args, **worker_kwargs):
+        super().__init__(*args, **worker_kwargs)
 
     async def start(self):
         await super().start()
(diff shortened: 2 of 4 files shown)


dask_cuda/cuda_worker.py
```diff
@@ -103,11 +103,15 @@ def del_pid_file():
             atexit.register(del_pid_file)
 
         if resources:
-            resources = resources.replace(",", " ").split()
+            resources = resources.replace(",", " ").replace("'", "").split()
             resources = dict(pair.split("=") for pair in resources)
-            resources = valmap(float, resources)
+            gpu_resources = valmap(int, itemfilter(lambda x: x != "GPU", resources))
+            resources = valmap(float, itemfilter(lambda x: x == "GPU", resources))
```
pentschev, line 109: @jacobtomlinson you have written this section originally, I presume mapping values to `float` was done to support a definition such as `"MEMORY"` but since I don't see any tests or any other explicit mention in Dask-CUDA, could you comment if that's right and whether you can think of more robust ways for us to handle types here other than what I wrote above as _"GPU"_ or _NOT "GPU"_?

dask_cuda/local_cuda_cluster.py
```diff
@@ -399,6 +399,12 @@ def __init__(
 
         self.pre_import = pre_import
 
+        if "resources" in worker_kwargs:
+            if "GPU" not in worker_kwargs["resources"]:
+                worker_kwargs["GPU"] = 1
+        else:
+            worker_kwargs["resources"] = {"GPU": 1}
```
rjzamora, line 406: I wonder if it makes sense to use `{"cuda": 1}` (i.e. use "cuda" as the resource name for a CUDA-capable GPU)?
This would align with the convention used in DL (e.g. pytorch)

pentschev: I wouldn't mind personally, but the keyword `GPU` is documented in [Distributed worker resources docs](https://distributed.dask.org/en/stable/resources.html), so that seems like a more universal option and better documented.

rjzamora: Okay nice - That's good enough for me.

wence- (approved):

dask_cuda/cuda_worker.py
jacobtomlinson: Honestly I don't remember the reasoning. I think dask handles these values as floats, as you say it's to support values like `MEMORY` or other arbitrary quantities.

rjzamora:
@pentschev - Sorry this fell through the cracks. Do you think this is good to go for 24.12?

pentschev:
@rjzamora I actually don't know how useful this is, @jacobtomlinson point out offline that tasks ignore annotations after the dask-expr work. Doing this seems like it may resolve the original ask, but at the same time it is very misleading to users who may attempt to use those annotations for their actual purpose with Dask. I don't have a strong opinion on not going ahead, so maybe @VibhuJawa and @ryantwolf who originally requested this functionality can comment on whether it solves their issue nevertheless and whether they think it's useful to introduce this even knowing it can mislead users (ourselves included).

rjzamora:
>I don't have a strong opinion on not going ahead

Okay, fair enough. I'll just assume we are holding off on this unless the Curator team tells us otherwise.

>@jacobtomlinson point out offline that tasks ignore annotations after the dask-expr work. Doing this seems like it may resolve the original ask, but at the same time it is very misleading to users who may attempt to use those annotations for their actual purpose with Dask.

Yes, annotations are not supported by dask-expr, and I don't personally expect that they will ever be supported. With that said, I **do** think worker-resources are still useful. In the long run, it may become possible to add the resource restriction to the new [TaskSpec](https://github.com/dask/dask/blob/11c21a3b83ef27bc48068bd47a04fcfdffa3ab09/dask/_task_spec.py#L555) `Task` object. It's also possible to use a scheduler plugin to apply these resources for specific tasks (as done in https://github.com/rapidsai/dask-cuda/pull/1398).

>Doing this seems like it may resolve the original ask, but at the same time it is very misleading to users who may attempt to use those annotations for their actual purpose with Dask.

Users who try to apply annotations *should* get a warning. So, I don't think it hurts much if resources are labeled or not.

PR #1402, opened 2024-11-04
Add warmup runs and profile all iterations to benchmarks

pentschev:
Add support for initial warmup runs in benchmarks and allows profiling all iterations or just the last one.

This is technically a breaking change since `--profile` now profiles all iterations, and the new `--profile-last` option profiles only the last one as `--profile` used to behave.


madsbk (approved):
Looks good, thanks @pentschev

pentschev:
Thanks @madsbk !

PR #1406, opened 2024-11-20
Disable UCXX tests in CI

pentschev:
Temporarily disable UCXX tests in CI due to some non-deterministic failures during code freeze phase. They will be reenabled after 24.12 release.


AyodeAwe (approved):

pentschev:
Thanks @AyodeAwe !

PR #1416, opened 2024-12-13
Add wheel tests to CI

### LINKED ISSUE #1344: [CI] add wheel tests in CI
## Description

This project currently produces wheels but doesn't have any tests on those wheels in CI.

Some should be added.

## Benefits of this work

* improves release confidence in wheels
   - *(esp. for changes like https://github.com/rapidsai/build-planning/issues/31 that alter how the wheels are built)*
* reduces the effort required to fix wheel-specific issues
   - *(via finding them here, at the source, instead of via bug reports from downstream projects or users)*

## Acceptance Criteria

* `dask-cuda` wheels are tested in CI, for at least one combination of `({python}, {cuda}, {linux})` from the set of things supported by the rest of RAPIDS (and ideally more)

## Approach

Use the existing wheel-testing workflows from `shared-workflows`, e.g. https://github.com/rapidsai/shared-workflows/blob/branch-24.08/.github/workflows/wheels-test.yaml.

For a specific example testing another pure-Python package, see `cuxfilter`:

* https://github.com/rapidsai/cuxfilter/blob/7e63c04fc121b347040bb465845748983a79bc3b/.github/workflows/pr.yaml#L69
* https://github.com/rapidsai/cuxfilter/blob/7e63c04fc121b347040bb465845748983a79bc3b/.github/workflows/test.yaml#L25
* https://github.com/rapidsai/cuxfilter/blob/7e63c04fc121b347040bb465845748983a79bc3b/ci/test_wheel.sh

## Notes

As of this writing, some tests rely on `kvikio` but `kvikio` does not yet have wheels (https://github.com/rapidsai/kvikio/pull/369). If there are still not `kvikio` wheels as of when this is picked up, it could be installed in the test environment using `conda`.

> *doesn't have any tests on those wheels in CI.*

There are other RAPIDS projects that depend on `dask-cuda`, install its wheels, and use the package installed via those wheels in CI. For example, `cuml`: https://github.com/rapidsai/cuml/blob/1560886d30cfe7d4472f334d487e5b9c136a270d/python/pyproject.toml#L91. That does provide some indirect testing.

This issue is just about adding such testing **here in this repo, where the wheels are produced**.

gforsyth:
Adding this now that wheels are available

- **deps(kvikio): add kvikio to CUDA version matrices**
- **test(wheels): enable wheel tests in CI**

Resolves #1344


pentschev (approved):
LGTM, thanks @gforsyth !

pentschev:
@gforsyth I've triggered CI manually now but FYI going forward you'll need to setup [GitHub commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) to ensure CI runs automatically.

pentschev:
We don't currently use tags/keyword in the PR titles, so as a nit I just changed its title to contain only the message, this is really just a nit, hope you don't mind Gil. 🙂

.github/workflows/pr.yaml
```diff
@@ -53,3 +54,12 @@ jobs:
...
       script: "ci/build_wheel.sh"
+  wheel-tests:
+    needs: wheel-build
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-test.yaml@branch-25.02
+    with:
+      build_type: pull-request
+      script: "ci/test_wheel.sh"
+      # This selects "ARCH=amd64 + the latest supported Python + CUDA".
+      matrix_filter: map(select(.ARCH == "amd64")) | max_by([(.PY_VER|split(".")|map(tonumber)), (.CUDA_VER|split(".")|map(tonumber))]) | [.]
```
jameslamb, line 65:
```suggestion
```

I think we should cover the full RAPIDS PR CI matrix here.

Even though the `dask-cuda` wheels are pure-Python and not CPU architecture or GPU-version dependent, with tags like this:

```text
dask_cuda-25.2.0a10-py3-none-any.whl
```

... some of its dependencies do produce different variants for those different environments, and it's possible `dask-cuda` could interact with them in different ways that we'd want to catch. I'm reminded for example of https://github.com/microsoft/LightGBM/issues/6509 and https://github.com/microsoft/LightGBM/pull/6654, where I found that there was some aarch64-specific issue whose resolution was "load OpenMP as early as possible".

By the way, for PRs that matrix is only 2 jobs. You can find them here:

https://github.com/rapidsai/shared-workflows/blob/3313dce51b2cfea4601c30a986cd5980087fe2c1/.github/workflows/wheels-test.yaml#L81-L85

ci/test_wheel.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# Copyright (c) 2023, NVIDIA CORPORATION.
```
jameslamb, line 2:
```suggestion
# Copyright (c) 2024, NVIDIA CORPORATION.
```

This is a brand new file, so the copyright date should be 2024. In a follow-up PR, could you add the `verify-copyright` pre-commit hook here to catch and fix stuff like this?

ref: https://github.com/rapidsai/cuspatial/blob/ad2900ded3c4a855774fc47dcec9a22323f07c0b/.pre-commit-config.yaml#L48

ci/test_wheel.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -eou pipefail
+
+RAPIDS_PY_VERSION="312" RAPIDS_PY_WHEEL_NAME="dask-cuda" rapids-download-wheels-from-s3 ./dist
```
jameslamb, line 6:
```suggestion
RAPIDS_PY_WHEEL_NAME="dask-cuda" RAPIDS_PY_WHEEL_PURE="1" rapids-download-wheels-from-s3 python ./dist
```

Let's match the way this is spelled in other pure-Python cases like:

* `jupyterlab-nvdashboard`: https://github.com/rapidsai/jupyterlab-nvdashboard/blob/d9746aefaab945241461046fb9d87334339c13a2/ci/test_wheel.sh
* `cudf-polars`: https://github.com/rapidsai/cudf/blob/a95fbc88f94df24c3418766fbbea5b6633ff2328/ci/test_wheel_cudf_polars.sh#L9

Keeping these scripts looking similar is helpful when we do touch-all-the-repos types of migrations.

dependencies.yaml
```diff
@@ -188,21 +188,21 @@ dependencies:
...
               - dask-cudf-cu12==25.2.*,>=0.0.0a0
+              - kvikio-cu12==25.2.*,>=0.0.0a0
               - ucx-py-cu12==0.42.*,>=0.0.0a0
           - matrix:
               cuda: "11.*"
               cuda_suffixed: "true"
             packages:
               - cudf-cu11==25.2.*,>=0.0.0a0
               - dask-cudf-cu11==25.2.*,>=0.0.0a0
+              - kvikio-cu11==25.2.*,>=0.0.0a0
```
jameslamb, line 205: oh nice, thank you!

jameslamb (changes requested):
Thanks for working on this! Left some suggestions.

Can you please also add a similar entry in https://github.com/rapidsai/dask-cuda/blob/branch-25.02/.github/workflows/test.yaml? You can follow the patterns from other repos, like https://github.com/rapidsai/cudf/blob/branch-25.02/.github/workflows/test.yaml

Those `.github/workflows/test.yaml` workflow configs define which tests run in 3 scenarios:

* on the target branch (e.g. `branch-25.02`) after merges
* nightly
  - runs: https://github.com/rapidsai/workflows/actions/workflows/nightly-pipeline.yaml
  - config: https://github.com/rapidsai/workflows/blob/main/.github/workflows/nightly-pipeline.yaml
* manually triggered in the UI via workflow dispatch

`@` me if you want some helping resolving the CI failures.

gforsyth:
Hey @jameslamb -- back to this after a brief interlude.  

Could you help me understand the dependency miss in the wheel tests?  pip isn't finding the (as yet unpublished) `cudf==25.2.*,>=0.0.0a0` -- is there a pre-release staging area that it should be pointing at?  Or are there no 25.2.* wheels available for `cudf` yet?

jameslamb:
> *pip isn't finding the (as yet unpublished) cudf==25.2.*,>=0.0.0a0 *

Ah yep! I can help with this:

```text
ERROR: Could not find a version that satisfies the requirement cudf==25.2.*,>=0.0.0a0; extra == "test" (from dask-cuda[test]) (from versions: 0.6.1.post1)
ERROR: No matching distribution found for cudf==25.2.*,>=0.0.0a0; extra == "test"
```

([build link](https://github.com/rapidsai/dask-cuda/actions/runs/12635066689/job/35204427351?pr=1416#step:9:75))

When we publish `cudf` packages, their distribution names have a suffix appended with the CUDA major version. This error is happening because we don't publish a package called `cudf`... it's `"cudf-cu11"` for CUDA 11 and `"cudf-cu12"` for CUDA 12.

The problem here is that the `[test]` extra for the `dask-cuda` wheels has unsuffixed name `"cudf"` in it.

This is an interesting one ... since we only publish `"dask-cuda"` (no suffix) here, the same wheel will be pulled in regardless of CUDA version. So we can't depend on a single `[test]` extra to tell us what versions of dependencies to pull in, because some of those dependencies have to be CUDA-major-version-specific.

I can think of 2 ways to address this:

* not having any extras for testing dependencies, and just directly installing them
* having 2 extras, like `[test-cuda11]` and `[test-cuda12]`

I think we should try just directly installing them in CI. That'd mean:

1. prevent that invalid `[test]` extra from getting written into the wheel: https://github.com/rapidsai/dask-cuda/blob/d9c6996c0ef57c3b8afa8ca7922f586589fff94a/dependencies.yaml#L50-L57
2. install the testing dependencies directly in `ci/test_wheel.sh` ([follow this related example from cudf](https://github.com/rapidsai/cudf/blob/955b1f4566abccf920a022dc78a1e654acf0de16/ci/build_wheel_libcudf.sh#L11-L22))

> *is there a pre-release staging area that it should be pointing at?*

That's already happening in this job.

RAPIDS pre-release packages are published at https://pypi.anaconda.org/rapidsai-wheels-nightly/simple/, and the docker images we run wheel CI in have configuration pointing there: https://github.com/rapidsai/ci-imgs/blob/39aa63d9799dbc9b8268cbbffb2f4d0b39192e67/context/pip.conf#L5

ci/test_wheel.sh
```diff
@@ -0,0 +1,21 @@
...
+    --output requirements \
+    --file-key "test_python" \
+    --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION};cuda_suffixed=true" \
+| tee /tmp/requirements-test.txt
+
+rapids-logger "Installing test dependencies"
+python -m pip install -v --prefere-binary -r /tmp/requirements-test.txt
+
+# echo to expand wildcard before adding `[extra]` requires for pip
+python -m pip install $(echo ./dist/dask_cuda*.whl)
```
jameslamb, line 19:
```suggestion
# echo to expand wildcard before adding `[extra]` requires for pip
python -m pip install \
    -r /tmp/requirements-test.txt \
    ./dist/dask_cuda*.whl
```

I think we should group these `pip install` calls together.

That'll reduce the risk of creating an inconsistent environment, and give us a better chance of catching packaging issues.

For example, imagine if `dask-cuda` had a runtime dependency on `numpy<2` but we had a pin of `numpy>=2` in testing dependencies. With separate `pip install` calls, `pip` would happily downgrade `numpy`. With them combined, we'll get an informative, loud error saying "hey these requirements are incompatible".

jameslamb (changes requested):
Thanks! Left one suggestion for your consideration.

gforsyth:
Thanks for all the pointers, @jameslamb !  Still waiting on a few conda jobs, but the wheel tests are green, so this is probably ready for another look!

gforsyth:
Hey @jameslamb -- rebased and green

jameslamb (approved):
Looks great, thanks for working through this!

PR #1417, opened 2024-12-16
Remove legacy Dask-cuDF handling

rjzamora:
Removes testing/handling for "legacy" Dask cuDF (i.e. `DASK_DATAFRAME__QUERY_PLANNING=False`).

This PR also adds support for the `"explicit-comms"` config with query-planning enabled (we used to raise an error telling the user to disable query planning).

This should be merged **before** https://github.com/rapidsai/cudf/pull/17558 (otherwise Dask-CUDA CI will break).
This PR is marked as "breaking", because it technically breaks the `"explicit-comms"` config with the "legacy" version of Dask cuDF (which we are about to remove in 25.02 anyway).


pentschev:
The build failure is not related to the changes here, it should be fixed by #1418 .

rjzamora:
Thanks @pentschev - Just a heads up that I **do** see some pytest failures locally. My `Shuffle` patching approach seems to have issues. I spent a lot of unsuccessful time debugging yesterday - So, I may need to roll back the `"explicit-comms"` config support (for now).

dask_cuda/tests/test_explicit_comms.py
```diff
@@ -250,7 +239,7 @@ def check_shuffle():
             ):
                 dask.config.refresh()  # Trigger re-read of the environment variables
                 with pytest.raises(ValueError, match="explicit-comms-batchsize"):
-                    ddf.shuffle(on="key", npartitions=4)
+                    ddf.shuffle(on="key", npartitions=4).dask
```
rjzamora, line 242: I'm seeing some very strange behavior locally - I'm unable to get this to raise with the env-variable approach :/

rjzamora: Works fine using `with dask.config.set(...)` (as long as there is no `dask.config.refresh()` call).

pentschev: It took me a couple of hours of debugging this too, but I finally understood why this happens @rjzamora. When `dask.config.refresh()` is called `dataframe.shuffle.method` is overwritten to `None`, and [dask-expr then calls `dask.utils.get_default_shuffle_method`](https://github.com/dask/dask-expr/blob/d7577a2ff22d47b36a45a87dd5ff2857af81341f/dask_expr/_collection.py#L920), that will [return `"p2p"`](https://github.com/dask/dask/blob/2c0ac83fce0e46a15d070b6e7a8f8fe897c00a28/dask/utils.py#L2251) because `dataframe.shuffle.method is None` now. We also [patch several of Dask's `get_default_shuffle_method`](https://github.com/rapidsai/dask-cuda/blob/72b364bd47b2fcbf31b42104dc900e746af9a47e/dask_cuda/__init__.py#L37-L45), but not `dask.utils.get_default_shuffle_method`, however, even if we patch it that doesn't work and I think this has to do with the import orders, if Dask-CUDA is loaded and patches those _after_ dask-expr imports those it will not see the patched functions and thus will default to Dask's original, which makes this approach not entirely reliable.

For this specific test you can just add `"DASK_DATAFRAME__SHUFFLE__METHOD": "tasks"` to the patched dictionary in line 249. However, after this long debugging session my conclusion is that relying on `dask.config.refresh()` can get very flaky real quick and I think the best course of action here is to check environment variables in a separate test where we launch the process already with the patched dictionary instead and here move to `with dask.config.set(...)`.

rjzamora: Okay - Thanks for your help debugging this @pentschev !

I think I have an idea how to avoid the need for any `get_default_shuffle_method` patching (fingers crossed).

rjzamora: Okay, yeah. It looks like we don't need to patch `get_default_shuffle_method` if we patch `Shuffle._lower`. As far as I can tell, this is the central place where `get_default_shuffle_method` is typically used to set the shuffle method to `"p2p"` anyway.

pentschev (approved):
I think this looks good now given tests are also passing. Would be best if @madsbk could have a look too before we merge.

jameslamb (approved):
Giving you a `packaging-codeowners` approval. Thanks for removing those warnings filters from `pyproject.toml`!

rjzamora:
@pentschev @madsbk - I plan to merge this tomorrow AM if there are no other review comments.

madsbk (approved):
Thanks @rjzamora, looks good.

PR #1418, opened 2024-12-17
Add missing `setuptools` to conda recipe

pentschev:
Conda builds are failing due to missing `setuptools`, this change add the missing dependency to fix the failure.

### DIFF
--- a/conda/recipes/dask-cuda/meta.yaml
+++ b/conda/recipes/dask-cuda/meta.yaml
@@ -30,7 +30,7 @@ requirements:
     - python
     - pip
     - rapids-build-backend>=0.3.0,<0.4.0.dev0
-    - setuptools
+    - setuptools>=64.0.0
   run:
     - python
     {% for r in data["project"]["dependencies"] %}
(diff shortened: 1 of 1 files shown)


conda/recipes/dask-cuda/meta.yaml
```diff
@@ -30,6 +30,7 @@ requirements:
     - python
     - pip
     - rapids-build-backend>=0.3.0,<0.4.0.dev0
+    - setuptools
```
jameslamb, line 33:
```suggestion
    - setuptools>=64.0.0
```

Let's please match the floor in https://github.com/rapidsai/dask-cuda/blob/d70f28d758f2dcae5aa8aeaad759c6acfa65309c/dependencies.yaml#L78

Somewhere around the 64.0.0 release (don't remember exactly where) is where `setuptools` added support for `pyproject.toml`, which we rely on.

jameslamb (approved):
Thanks! I think this will fix the issue you observed in https://github.com/rapidsai/dask-cuda/actions/runs/12366711202/job/34513856325.

And is appropriate given that `setuptools` is the backend for this package:

https://github.com/rapidsai/dask-cuda/blob/d70f28d758f2dcae5aa8aeaad759c6acfa65309c/pyproject.toml#L137

I guess we must have been getting it transitively from other packages before.

pentschev:
Thanks @jameslamb !

bdice (approved):

Issue #1117, opened 2023-02-14
Moving from `pynvml` to `nvidia-ml-py`


jakirkham:
Currently this is making use of `pynvml` in a few places:

https://github.com/rapidsai/dask-cuda/blob/a0fda7666d8d4310937d1fa8ed175c42330346f0/dask_cuda/utils.py#L13

https://github.com/rapidsai/dask-cuda/blob/a0fda7666d8d4310937d1fa8ed175c42330346f0/dask_cuda/tests/test_utils.py#L186-L187

https://github.com/rapidsai/dask-cuda/blob/a0fda7666d8d4310937d1fa8ed175c42330346f0/dependencies.yaml#L103

However we would like to move to `nvidia-ml-py` in the future. Raising this issue to track this work

wence-:
I think it is sufficient to just change the dependency, since pynvml is identical in interface.

PR #1419, opened 2024-12-19
Use PyNVML 12

jakirkham:
Bump `pynvml` from `11` to `12`. This version of `pynvml` also now depends on `nvidia-ml-py` for core functionality.


jakirkham:
As cuDF is a test dependency ( https://github.com/rapidsai/cudf/pull/17627 ), we still pick up PyNVML 11 in some places

Think we will need to get one of these PRs in first (perhaps this one?) then come back and retest once cuDF is also built

jameslamb (approved):
Not strictly related to this PR's changes, but they'd help us build confidence in this and future `pynvml`-related changes... since `pynvml` is a hard runtime dependency, could we switch to using it **unconditionally** in tests, by removing stuff like this?

https://github.com/rapidsai/dask-cuda/blob/fd8a73659c89fc5d2678a6343002f7a2fc0a1a1f/dask_cuda/tests/test_utils.py#L200

https://github.com/rapidsai/dask-cuda/blob/fd8a73659c89fc5d2678a6343002f7a2fc0a1a1f/dask_cuda/tests/test_utils.py#L253

Think those skips could potentially cover up some packaging issues. I'd support trying to remove those right here in this PR, but also fine if you want to do it as a follow-up.

pentschev:
> Not strictly related to this PR's changes, but they'd help us build confidence in this and future `pynvml`-related changes... since `pynvml` is a hard runtime dependency, could we switch to using it **unconditionally** in tests, by removing stuff like this?

I agree that's sensible. @jakirkham do you want to address this in this PR or would you prefer I do that in a separate PR?

jakirkham:
As discussed offline, this is being handled in PR: https://github.com/rapidsai/dask-cuda/pull/1421

jakirkham:
Thanks James and Peter! 🙏

Looks like it is passing. Let's get this in :shipit:

jakirkham:
Fixed issue: https://github.com/rapidsai/dask-cuda/issues/1117

PR #1423, opened 2025-01-06
Add verify-copyright pre-commit hook

gforsyth:
Adds a pre-commit hook to ensure files have an up-to-date copyright notice.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,6 +36,7 @@ repos:
         rev: v0.4.0
         hooks:
             - id: verify-alpha-spec
+            - id: verify-copyright
       - repo: https://github.com/rapidsai/dependency-file-generator
         rev: v1.16.0
         hooks:


.pre-commit-config.yaml
```diff
@@ -36,6 +36,8 @@ repos:
         rev: v0.4.0
         hooks:
             - id: verify-alpha-spec
+            - id: verify-copyright
+              args: [--fix, --main-branch=branch-25.02]
```
jameslamb, line 40:
```suggestion
```

I don't think you should need `--fix` or `--main-branch` for this hook... I don't see it in the other uses of this hook across RAPIDS: https://github.com/search?q=org%3Arapidsai+%22id%3A+verify-copyright%22+AND+NOT+is%3Aarchived&type=code

`--fix` is the default already (https://github.com/rapidsai/pre-commit-hooks/blob/7dd46df494144721375eb4d7dfa986b0c1b62924/.pre-commit-hooks.yaml#L53)

And `--main-branch` is only needed for projects that don't follow the RAPIDS calendar-versioning scheme, like those whose target branch is literally called `main` ([e.g. `pynvjitlink`](https://github.com/rapidsai/pynvjitlink/blob/cab3ad1078fe592ebf6b948caf3c8101c0959fa1/.pre-commit-config.yaml#L23-L25)).

jameslamb (changes requested):
Left one suggestion.

`dask-cuda` 25.02 is past code freeze at this point, can you re-target this to `branch-25.04` please?

.pre-commit-config.yaml
gforsyth: Thanks, @jameslamb !  Retargeted, rebased, and updated.

rjzamora (approved):
Thanks @gforsyth !

jameslamb (approved):

PR #1427, opened 2025-01-22
Add shellcheck to pre-commit and fix warnings

gforsyth:
## Description                                                                                           
                                                                                                              
 shellcheck  is a fast, static analysis tool for shell scripts. It's good at                                  
flagging up unused variables, unintentional glob expansions, and other potential                              
execution and security headaches that arise from the wonders of  bash  (and other shlangs).                   
                                                                                                              
This PR adds a  pre-commit  hook to run  shellcheck  on all of the  sh-lang  files in the  ci/  directory, and
the changes requested by  shellcheck  to make the existing files pass the check.                              
                                                                                                              
xref: rapidsai/build-planning#135

### DIFF
--- a/ci/build_docs.sh
+++ b/ci/build_docs.sh
@@ -25,7 +25,8 @@ rapids-mamba-retry install \
     --channel "${PYTHON_CHANNEL}" \
     "dask-cuda=${RAPIDS_VERSION}"
 
-export RAPIDS_DOCS_DIR="$(mktemp -d)"
+RAPIDS_DOCS_DIR="$(mktemp -d)"
+export RAPIDS_DOCS_DIR
 
 rapids-logger "Build Python docs"
 pushd docs
--- a/ci/validate_wheel.sh
+++ b/ci/validate_wheel.sh
@@ -9,10 +9,10 @@ rapids-logger "validate packages with 'pydistcheck'"
 
 pydistcheck \
     --inspect \
-    "$(echo ${wheel_dir_relative_path}/*.whl)"
+    "$(echo "${wheel_dir_relative_path}"/*.whl)"
 
 rapids-logger "validate packages with 'twine'"
 
 twine check \
     --strict \
-    "$(echo ${wheel_dir_relative_path}/*.whl)"
+    "$(echo "${wheel_dir_relative_path}"/*.whl)"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,6 +41,12 @@ repos:
         hooks:
             - id: rapids-dependency-file-generator
               args: ["--clean"]
+      - repo: https://github.com/shellcheck-py/shellcheck-py
+        rev: v0.10.0.1
+        hooks:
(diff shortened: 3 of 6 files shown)


bdice (approved):

ci/build_docs.sh
```diff
@@ -25,7 +25,8 @@ rapids-mamba-retry install \
     --channel "${PYTHON_CHANNEL}" \
     "dask-cuda=${RAPIDS_VERSION}"
 
-export RAPIDS_DOCS_DIR="$(mktemp -d)"
+RAPIDS_DOCS_DIR="$(mktemp -d)"
+export RAPIDS_DOCS_DIR
```
pentschev, line 29: This looks very weird to me and unnecessarily verbose, why is that change recommended by shellcheck?

ci/validate_wheel.sh
```diff
@@ -9,10 +9,10 @@ rapids-logger "validate packages with 'pydistcheck'"
 
 pydistcheck \
     --inspect \
-    "$(echo ${wheel_dir_relative_path}/*.whl)"
+    "$(echo "${wheel_dir_relative_path}"/*.whl)"
```
pentschev, line 12: Maybe I'm mistaken, but wouldn't this mean `${wheel_dir_relative_path}` is not within double-quotes since the open double-quote starts before `$(echo ...`?

ci/validate_wheel.sh
```diff
@@ -9,10 +9,10 @@ rapids-logger "validate packages with 'pydistcheck'"
...
 
 rapids-logger "validate packages with 'twine'"
 
 twine check \
     --strict \
-    "$(echo ${wheel_dir_relative_path}/*.whl)"
+    "$(echo "${wheel_dir_relative_path}"/*.whl)"
```
pentschev, line 18: Similar to the above, doesn't this mean it's actually outside double-quotes?

pentschev (commented):
Mostly looks good, but there are a couple of entries I'm not sure about, I presume you have more experience with shellcheck so could you comment on those choices @gforsyth ?

ci/build_docs.sh
gforsyth: It's a `bash` footgun.  
If it's a single-line, then the return code of the line is the result of `export` and it swallows the return code of the captured command.

ci/validate_wheel.sh
gforsyth: Another `bash` footgun, quoting resets inside of captures

ci/build_docs.sh
TomAugspurger: And when running this locally, the errors will include a link explaining the issue: https://www.shellcheck.net/wiki/SC2155

pentschev: Ah makes sense, but it's also only problematic if you're running a command, if you're assigning a static value then exporting in the same line should have no problems. Thanks both!

pentschev (approved):
LGTM, thanks @gforsyth .

PR #1432, opened 2025-01-28
Build and test with CUDA 12.8.0

bdice:
This PR uses CUDA 12.8.0 to build and test.

xref: https://github.com/rapidsai/build-planning/issues/139

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -16,7 +16,7 @@ To install the latest version of Dask-CUDA along with CUDA Toolkit 12.5:
 
 .. code-block:: bash
 
-    conda install -c rapidsai -c conda-forge -c nvidia dask-cuda cuda-version=12.5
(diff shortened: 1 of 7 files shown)


docs/source/install.rst
```diff
@@ -16,7 +16,7 @@ To install the latest version of Dask-CUDA along with CUDA Toolkit 12.5:
 
 .. code-block:: bash
 
-    conda install -c rapidsai -c conda-forge -c nvidia dask-cuda cuda-version=12.5
+    conda install -c rapidsai -c conda-forge -c nvidia dask-cuda cuda-version=12.8
```
jameslamb, line 19: The line above this needs to modified as well. It currently reads "... along with CUDA Toolkit 12.5".

I recommend we just change that to "... along with the CUDA Toolkit", but also fine if you just want to change 12.5 to 12.8.

bdice: Improved in 7c89a67. Thanks!

PR #1440, opened 2025-02-04
Reorganize CI tests and benchmarks

pentschev:
Recently wheel tests have been failing, this is probably due to the `DASK_CUDA_WAIT_WORKERS_MIN_TIMEOUT=20` configuration that is set for conda tests but not for wheels. With this change, the tests and benchmarks are now moved to new scripts that can be called by both conda and wheels tests to provide unified configurations with sane values for CI, and therefore should also resolve the CI failures.

Additionally removed `--durations` which was previously added to help understand timeouts but has not been of much use since.

### DIFF
--- a/ci/run_benchmarks.sh
+++ b/ci/run_benchmarks.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+# Copyright (c) 2025, NVIDIA CORPORATION.
+
+set -euo pipefail
+
+# Support invoking run_benchmarks.sh outside the script directory
+cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../dask_cuda
+
+rapids-logger "Run local benchmark"
+python benchmarks/local_cudf_shuffle.py \
+  --partition-size="1 KiB" \
+  -d 0 \
+  --runs 1 \
+  --backend dask
+
+python benchmarks/local_cudf_shuffle.py \
+  --partition-size="1 KiB" \
+  -d 0 \
+  --runs 1 \
+  --backend explicit-comms
+
+python benchmarks/local_cudf_shuffle.py \
+  --disable-rmm \
+  --partition-size="1 KiB" \
+  -d 0 \
+  --runs 1 \
+  --backend explicit-comms
+
+python benchmarks/local_cudf_shuffle.py \
+  --disable-rmm-pool \
+  --partition-size="1 KiB" \
+  -d 0 \
+  --runs 1 \
+  --backend explicit-comms
+
+python benchmarks/local_cudf_shuffle.py \
+  --rmm-pool-size 2GiB \
+  --partition-size="1 KiB" \
+  -d 0 \
+  --runs 1 \
+  --backend explicit-comms
+
+python benchmarks/local_cudf_shuffle.py \
+  --rmm-pool-size 2GiB \
+  --rmm-maximum-pool-size 4GiB \
+  --partition-size="1 KiB" \
+  -d 0 \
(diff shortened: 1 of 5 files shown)


pentschev:
Note that I've targeted 25.02 because we're still running nightlies there for at least a few more days, I believe. The changes are only CI-related, no actual Dask-CUDA code changes are being made.

TomAugspurger (approved):

ci/run_benchmarks.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+# Copyright (c) 2025, NVIDIA CORPORATION.
+
+set -euo pipefail
+
+# Support invoking run_benchmarks.sh outside the script directory
+cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../dask_cuda
+
+rapids-logger "Run local benchmark"
```
jameslamb, line 9:
```suggestion
echo "Run local benchmark"
```

By convention, we try to keep any stuff from `gha-tools` like `rapids-logger` out of these `ci/run_*.sh` scripts, as those scripts are intended to be run outside of CI as well (like in DLFW builds).

Could we use `echo` here instead?

ci/run_pytest.sh
```diff
@@ -0,0 +1,21 @@
...
+# Support invoking run_pytest.sh outside the script directory
+cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../dask_cuda
+
+rapids-logger "pytest dask-cuda"
+DASK_CUDA_TEST_SINGLE_GPU=1 \
+  DASK_CUDA_WAIT_WORKERS_MIN_TIMEOUT=20 \
+  UCXPY_IFNAME=eth0 \
+  UCX_WARN_UNUSED_ENV_VARS=n \
+  UCX_MEMTYPE_CACHE=n \
+  timeout 90m pytest \
```
jameslamb, line 15:
```suggestion
  timeout 90m python -m pytest \
```

Wheel test jobs are failing like this:

```text
timeout: failed to run command 'pytest': No such file or directory
```

([build link](https://github.com/rapidsai/dask-cuda/actions/runs/13133642104/job/36644321569?pr=1440))

Suspect you need `python -m pytest` here, as it was in `ci/test_wheel.sh`.

jameslamb (approved):
I left a few small suggestions. Giving you a `ci-codeowners` approval, so you can merge without a re-review after you've addressed them.

ci/run_benchmarks.sh
pentschev: Makes sense, however, I prefer `rapids-logger` as it makes it easier to visually identify that in the logs, so I moved the calls to the `test_*.sh` files instead.

ci/run_pytest.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+# Copyright (c) 2025, NVIDIA CORPORATION.
+
+set -euo pipefail
+
+# Support invoking run_pytest.sh outside the script directory
+cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../dask_cuda
+
+rapids-logger "pytest dask-cuda"
```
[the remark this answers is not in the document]
pentschev: As above, moved `rapids-logger` to `test_*.sh` files.

pentschev:
I had to xfail one test (see https://github.com/rapidsai/dask-cuda/issues/1441), that's why we also require a daskcuda-python-codeowners review.

rjzamora (approved):

PR #1445, opened 2025-02-06
add telemetry

msarahan:
Enables telemetry during CI runs. This is done by parsing GitHub Actions run log metadata and should have no impact on build or test times.

xref https://github.com/rapidsai/build-infra/issues/139

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -18,11 +18,28 @@ jobs:
       - docs-build
       - wheel-build
       - wheel-tests
+      - telemetry-setup
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@nvks-runners
+    with:
+      needs: ${{ toJSON(needs) }}
+  telemetry-setup:
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    env:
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
Looks good to me! I looked through the logs of `telemetry-setup` and `telemetry-summarize`, didn't see any obvious issues.

And I saw the check statuses as ✅.

The build summary had a small number of these warnings:

> Warning: Failed to download action 'https://api.github.com/repos/actions/github-script/tarball/60a0d83039c74a4aee543508d2ffcb1c3799cdea'. Error: Name or service not known (codeload.github.com:443) 
Warning: Back off 26.768 seconds before retry.

logs: https://github.com/rapidsai/dask-cuda/actions/runs/13182783704/job/36798012178#step:4:165

summary: https://github.com/rapidsai/dask-cuda/actions/runs/13182783704 (scroll down to "Annotations"

Do those matter? It sort of looks to me from those logs like that's something that just failed on a first attempt and then succeeded on a retry, but just calling them out because you'd know better if that's a symptom of a problem.

msarahan:
Looks to me like it is failing to download the github-hosted `github-script` action, which smells really bad, but which isn't anything we can address. I'm glad that Github at least has an automatic retry on it.

Issue #1450, opened 2025-02-13
Explicit Comms Object Not Cleared After Cluster State Change or Restart


VibhuJawa:
## Description

The explicit_comms module in dask_cuda does not properly reset or clear up after a cluster state change or when a new cluster is started. This results in default_comms().worker_addresses retaining references to old worker addresses, even after the original cluster has been cleaned up. The communication object does not respect the lifetime of the worker or cluster objects.

```python3
import dask_cuda
from dask_cuda import LocalCUDACluster
from dask.distributed import Client
from dask_cuda.explicit_comms import comms

# Start a cluster and check the worker addresses
with LocalCUDACluster(n_workers=1) as cluster, Client(cluster) as client:
    print(client.scheduler_info()["workers"].keys())  # Expected: Current worker
    print(comms.default_comms().worker_addresses)  # Expected: Current worker
```
Both addresses align 
```python3
dict_keys(['tcp://127.0.0.1:43585'])
['tcp://127.0.0.1:43585']
```

```python3

# At this point, the cluster has been cleaned up,
# but comms.default_comms() still retains the old worker address
print(comms.default_comms().worker_addresses)  
```
```python3
['tcp://127.0.0.1:43585'] # Unexpected: Old worker address still present
```

```python3
# Start a new cluster and check worker addresses
with LocalCUDACluster(n_workers=1) as cluster, Client(cluster) as client:
    print(client.scheduler_info()["workers"].keys())  # Expected: New worker
    print(comms.default_comms().worker_addresses)  # Unexpected: Old worker address still retained

```

 Addressees do not align anymore
```python3
dict_keys(['tcp://127.0.0.1:38569']) # Expected: New worker
['tcp://127.0.0.1:43585'] # Unexpected: Old worker address still retained
```

### Observed Behavior
When the first cluster is cleaned up, `comms.default_comms().worker_addresses` still references the old worker.

After starting a new cluster, `default_comms().worker_addresses` continues to reference the worker from the previous cluster rather than updating to the new worker.

###  Expected Behavior
comms.default_comms().worker_addresses should be cleared or reset when the cluster is shut down or workers are updated.
When a new cluster is started, default_comms().worker_addresses should reflect the new worker(s) only.

## Additional Context: 
This issue led to multiple unexpected CI failures in NeMo Curator ([PR #540](https://github.com/NVIDIA/NeMo-Curator/pull/540)), which took significant effort to diagnose and debug.

[...]

PR #1451, opened 2025-02-14
Avoid stale CommContext in explicit comms

TomAugspurger:
This PR updates the CommContext caching to be keyed by some information about the cluster, rather than a single global. This prevents us from using a stale comms object after the cluster changes (add or remove workers) or is recreated entirely.

Closes https://github.com/rapidsai/dask-cuda/issues/1450

### DIFF
--- a/dask_cuda/explicit_comms/comms.py
+++ b/dask_cuda/explicit_comms/comms.py
@@ -1,15 +1,21 @@
+# Copyright (c) 2021-2025 NVIDIA CORPORATION.
 import asyncio
 import concurrent.futures
 import contextlib
 import time
 import uuid
+import weakref
 from typing import Any, Dict, Hashable, Iterable, List, Optional
 
 import distributed.comm
+from dask.tokenize import tokenize
 from distributed import Client, Worker, default_client, get_worker
 from distributed.comm.addressing import parse_address, parse_host_port, unparse_address
 
-_default_comms = None
+# Mapping tokenize(client ID, [worker addresses]) to CommsContext
+_comms_cache: weakref.WeakValueDictionary[
+    str, "CommsContext"
+] = weakref.WeakValueDictionary()
 
 
 def get_multi_lock_or_null_context(multi_lock_context, *args, **kwargs):
@@ -38,9 +44,10 @@ def get_multi_lock_or_null_context(multi_lock_context, *args, **kwargs):
 
 
 def default_comms(client: Optional[Client] = None) -> "CommsContext":
-    """Return the default comms object
+    """Return the default comms object for ``client``.
 
-    Creates a new default comms object if no one exist.
+    Creates a new default comms object if one does not already exist
+    for ``client``.
 
     Parameters
     ----------
@@ -52,11 +59,31 @@ def default_comms(client: Optional[Client] = None) -> "CommsContext":
     -------
     comms: CommsContext
         The default comms object
+
+    Notes
+    -----
+    There are some subtle points around explicit-comms and the lifecycle
+    of a Dask Cluster.
+
+    A :class:`CommsContext` establishes explicit communication channels
+    between the workers *at the time it's created*. If workers are added
+    or removed, they will not be included in the communication channels
+    with the other workers.
+
+    If you need to refresh the explicit communications channels, then
+    create a new :class:`CommsContext` object or call ``default_comms``
+    again after workers have been added to or removed from the cluster.
     """
-    global _default_comms
-    if _default_comms is None:
-        _default_comms = CommsContext(client=client)
-    return _default_comms
+    # Comms are unique to a {client, [workers]} pair, so we key our
+    # cache by the token of that.
+    client = client or default_client()
+    token = tokenize(client.id, list(client.scheduler_info()["workers"].keys()))
+    maybe_comms = _comms_cache.get(token)
+    if maybe_comms is None:
+        maybe_comms = CommsContext(client=client)
+        _comms_cache[token] = maybe_comms
+
+    return maybe_comms
 
 
 def worker_state(sessionId: Optional[int] = None) -> dict:
--- a/dask_cuda/tests/test_explicit_comms.py
+++ b/dask_cuda/tests/test_explicit_comms.py
@@ -1,3 +1,5 @@
+# Copyright (c) 2021-2025 NVIDIA CORPORATION.
+
 import asyncio
 import multiprocessing as mp
 import os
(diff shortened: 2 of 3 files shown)


dask_cuda/explicit_comms/comms.py
```diff
@@ -52,11 +59,31 @@ def default_comms(client: Optional[Client] = None) -> "CommsContext":
...
+    of a Dask Cluster.
+
+    A :class:`CommsContext` establishes explicit communication channels
+    between the workers *at the time it's created*. If workers are added
+    or removed, they will not be included in the communication channels
+    with the other workers.
+
+    If you need to refresh the explicit communications channels, then
+    create a new :class:`CommsContext` object or call ``default_comms``
+    again after workers have been added to or removed from the cluster.
```
rjzamora, line 75: Nice - Thanks for adding these notes.

dask_cuda/explicit_comms/comms.py
```diff
@@ -1,15 +1,21 @@
...
 import distributed.comm
+from dask.tokenize import tokenize
 from distributed import Client, Worker, default_client, get_worker
 from distributed.comm.addressing import parse_address, parse_host_port, unparse_address
 
-_default_comms = None
+# Mapping tokenize(client ID, [worker addresses]) to CommsContext
+_comms_cache: weakref.WeakValueDictionary[
+    str, "CommsContext"
+] = weakref.WeakValueDictionary()
```
rjzamora, line 18: I like the idea of using a `WeakValueDictionary` here.

dask_cuda/explicit_comms/comms.py
```diff
@@ -52,11 +59,31 @@ def default_comms(client: Optional[Client] = None) -> "CommsContext":
...
+    A :class:`CommsContext` establishes explicit communication channels
+    between the workers *at the time it's created*. If workers are added
+    or removed, they will not be included in the communication channels
+    with the other workers.
+
+    If you need to refresh the explicit communications channels, then
+    create a new :class:`CommsContext` object or call ``default_comms``
+    again after workers have been added to or removed from the cluster.
     """
-    global _default_comms
```
rjzamora, line 56: Python question - Do you happen to know why we needed `global` before, but don't need it anymore?

dask_cuda/tests/test_explicit_comms.py
```diff
@@ -415,3 +417,79 @@ def test_lock_workers():
...
+    # because we referenced the old cluster's addresses.
+    with LocalCluster(n_workers=1) as cluster:
+        with Client(cluster) as client:
+            scheduler_addresses_new = list(client.scheduler_info()["workers"].keys())
+            comms_addresses_new = list(comms.default_comms().worker_addresses)
+
+    assert scheduler_addresses_new == comms_addresses_new
+
+
+def test_update():
```
rjzamora, line 455: Can we add a brief description of this test, and maybe make the test name a bit more descriptive?

dask_cuda/tests/test_explicit_comms.py
```diff
@@ -415,3 +417,79 @@ def test_lock_workers():
...
+        with Client(cluster) as client:
+            context = comms.default_comms()
+            scheduler_addresses_old = list(client.scheduler_info()["workers"].keys())
+            comms_addresses_old = list(comms.default_comms().worker_addresses)
+            assert comms.default_comms() is context
+            assert len(comms._comms_cache) == 1
+
+            # Add a worker, which should have a new comms object
+            cluster.scale(2)
+            client.wait_for_workers(2)
```
rjzamora, line 435: Do we need a conservative timeout to avoid flaky CI?

rjzamora (commented):
This is really nice @TomAugspurger - Thanks!

I left a few minor comments. One additional question: Do you think we need to test an explicit-comms shuffle **after** the comms context has changed? These tests show that the comms context can be refreshed after we scale up or down workers, but I suppose we could also make sure that the refreshed context "works" as expected?

dask_cuda/explicit_comms/comms.py
TomAugspurger: I believe it's whether you're you're reassigning the object or not. Previously, we changed the name `_default_comms` from referencing `None` to some `CommsContext`. Now the object `_comms_cache` isn't being reassigned. It's always that WeakValueDictionary. We just happen to mutate it in this function.

TomAugspurger:
> Do you think we need to test an explicit-comms shuffle after the comms context has changed

Yeah, that would be good. I'll add that.

dask_cuda/explicit_comms/comms.py
TomAugspurger: Yeah, the `CommsContext holds a reference to a `Client` so we don't want to be the last one referencing it and keeping it alive.

(There's maybe a discussing to be had about whether CommContext needs to be client-specific, or whether two clients might be able to reuse the same explicit comms. Skimming the code it seems like it might be possible, but I don't want to go down that rabbit hole for something that is probably never going to get used).

TomAugspurger:
The CI failure appears to be an unrelated timeout: https://github.com/rapidsai/dask-cuda/actions/runs/13332958081/job/37241713827?pr=1451#step:10:5333

But I'm going to spend a bit of time to try to understand what's going on.

rjzamora:
>But I'm going to spend a bit of time to try to understand what's going on.

I certainly appreciate that. It's also fine to try rerunning that check given that a `Timeout` error like that is probably unrelated to the PR (and spilling tests are known to be flaky in CI :/ ).

TomAugspurger:
I didn't figure out much. This test logs stuff at various points in time:

1. cluster startup
2. Unmanaged memory warnings (I think from imports)
3. worker_assert (x2)
4. assert_host_chunks
5. assert_disk_chunks

[Success logs](https://github.com/rapidsai/dask-cuda/actions/runs/13311363881/job/37174806571#step:10:5051) from a run yesterday.

[Failure logs](https://github.com/rapidsai/dask-cuda/actions/runs/13332958081/job/37241713827?pr=1451#step:10:5267) from this run.

This table shows very roughly how long some stages took:

| stage              | failure  | success  |
| ------------------ | -------- | -------  |
| unmanaged memory   | baseline | baseline |
| worker_assert      | +6s      | +1s      |
| worker_assert2     | +8       | +0.5s    |
| assert_host_chunks | +0       | +0       |
| assert_disk_chunks | +0       | +0       |


So for whatever reason, the `worker_assert`s took a while on the failed run. That does involve an RPC between the client and the worker, but most of the time should be spent in `device_host_file_size_matches`. I'd need to look a bit more closely about where stuff could be taking time there (it does interact with the filesystem). Probably not today, so I'll restart that one failure.

PR #1453, opened 2025-02-20
ensure dask_cuda.__git_commit__ is populated

jameslamb:
Installing `dask-cuda` like this:

```shell
pip install \
  --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple/ \
  'dask-cuda==25.4.*,>=0.0.0a0'
```

The `__git_commit__` attribute on the main module isn't populated:

```shell
python -c "import dask_cuda; print(dask_cuda.__git_commit__)"
```

The way this *should* work is that `rapids-build-backend` writes a file `dask_cuda/GIT_COMMIT` which is then read by this code: 

https://github.com/rapidsai/dask-cuda/blob/412ef5891f1cca78af48c076e0922874c227b34b/dask_cuda/_version.py#L20-L28

I think that what's happening here is this:

* `rapids-build-backend` *is* writing that file
* the file is not being packaged, because this project uses `setuptools` + a `MANIFEST.in`, and that `MANIFEST.in` does not include that file

This proposes the following:

* add `GIT_COMMIT` to `MANIFEST.in`
* update RAPIDS-specific pre-commit hooks to their latest versions (not related, but might as well, while we're using a CI run anyway)

## Notes for Reviewers

Helpful reference for this... "Controlling files in the distribution" from the `setuptools` docs: https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,12 +33,12 @@ repos:
                 args: ["--module=dask_cuda", "--ignore-missing-imports"]
                 pass_filenames: false
       - repo: https://github.com/rapidsai/pre-commit-hooks
-        rev: v0.4.0
+        rev: v0.6.0
         hooks:
             - id: verify-alpha-spec
             - id: verify-copyright
       - repo: https://github.com/rapidsai/dependency-file-generator
-        rev: v1.16.0
+        rev: v1.17.1
         hooks:
             - id: rapids-dependency-file-generator
               args: ["--clean"]
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,3 @@
+include dask_cuda/GIT_COMMIT
 include dask_cuda/_version.py
 include dask_cuda/VERSION


TomAugspurger:
Here's a run from a previous PR: https://github.com/rapidsai/dask-cuda/actions/runs/13332958081/job/37252862171. There's no `GIT_COMMIT` in the build output.

The build output from this PR is at https://github.com/rapidsai/dask-cuda/actions/runs/13439129864/job/37548802662?pr=1453#step:11.

There's a warning about no files matching `GIT_COMMIT`, which is maybe understandable since it's generated during the build? But we do have `dask_cuda/VERSION` symlinked to `VERSION` https://github.com/rapidsai/dask-cuda/blob/branch-25.04/dask_cuda/VERSION.

Anyway, what we really care about is that the built wheel has it, and it looks like it does:

```
(base) toaugspurger@dgx12:/tmp/a$ curl -LO "https://downloads.rapids.ai/ci/dask-cuda/pull-request/1453/c2cd9ba/dask-cuda_wheel_python_dask-cuda.tar.gz"
...
(base) toaugspurger@dgx12:/tmp/a$ tar xvf dask-cuda_wheel_python_dask-cuda.tar.gz
./
./dask_cuda-25.4.0a23-py3-none-any.whl
(base) toaugspurger@dgx12:/tmp/a$ unzip -l dask_cuda-25.4.0a23-py3-none-any.whl | grep GIT_COMMIT
       41  2025-02-20 15:49   dask_cuda/GIT_COMMIT
```

jameslamb:
Yeah thank you for looking carefully at that! I think I can explain what's happening there.

The key point is this from the `setuptools` docs:

> _You can think about the build process as two stages: first the sdist will be created and then the wheel will be produced from that sdist._

([docs link](https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html#id2))

The first pass through, creating an sdist, is what emits these warnings:

```text
reading manifest template 'MANIFEST.in'
warning: no files found matching 'dask_cuda/GIT_COMMIT'
```

([build link](https://github.com/rapidsai/dask-cuda/actions/runs/13439129864/job/37548802662?pr=1453#step:11:123))

That happens around here in `setuptools`: https://github.com/pypa/setuptools/blob/ba243756233d0afe944db6e02ddcb70064dcd22c/setuptools/_distutils/command/sdist.py#L342

And at that point, that `GIT_COMMIT` file doesn't exist because it hasn't been created yet.

But then later on, during wheel-building, those rules in the manifest are executed one at a time.

```text
writing manifest file 'dask_cuda.egg-info/SOURCES.txt'
copying dask_cuda/GIT_COMMIT -> build/lib/dask_cuda
copying dask_cuda/VERSION -> build/lib/dask_cuda
```

([build link](https://github.com/rapidsai/dask-cuda/actions/runs/13439129864/job/37548802662?pr=1453#step:11:201))

And by there, the file exists and gets packaged.

TomAugspurger (commented):
Looks great, thanks!

msarahan (approved):

bdice (approved):

PR #1458, opened 2025-03-21
chore(conda): enable strict channel priority for conda build

gforsyth:
Enabling strict channel priority repo-by-repo


gforsyth:
The `dask/label/dev` channel breaks strict channel priority and prevents solves in the post-build tests.  Right now I'm removing that channel only in this PR in a hacky way -- checking with other projects to see if _any_ of them need it.  And if not, I'll just rip it out in `gha-tools`

jameslamb (approved):

PR #1459, opened 2025-03-25
Updated shuffle for new setattr

TomAugspurger:
The base class for Shuffle has changed such that doing self._name = value doesn't work for attributes not defined on the class when it's instantiated. Trying to monkey-patch that on later gives

```
  File "/__w/dask-upstream-testing/dask-upstream-testing/.venv/lib/python3.12/site-packages/dask/_expr.py", line 183, in __setattr__
    raise AttributeError(
AttributeError: ECShuffle object has no attribute _ec_shuffled
```

xref https://github.com/rapidsai/dask-upstream-testing/issues/37


rjzamora (approved):
Seems reasonable to me. Thanks Tom!

PR #1460, opened 2025-04-02
Port all conda recipes to `rattler-build`

gforsyth:
Port all condabuild recipes over to use `rattler-build` instead.

Contributes to rapidsai/build-planning#47

- To satisfy `rattler-build`, this changes all the licenses in the `pyproject.toml` files to the SPDX-compliant `Apache-2.0` instead of `Apache 2.0`

Is there any reason this shouldn't be a `python: noarch` package?


conda/recipes/dask-cuda/recipe.yaml
```diff
@@ -0,0 +1,59 @@
...
+package:
+  name: dask-cuda
+  version: ${{ version }}
+
+source:
+  path: ../../..
+
+build:
+  noarch: python
+  string: py${{ py_buildstring }}_${{ date_string }}_${{ head_rev }}
```
gforsyth, line 21: This seems like a `noarch` package to me, in which case we should remove the python build string from the version string, and also restrict the build to a single Python CI job

ci/build_docs.sh
```diff
@@ -4,7 +4,10 @@
 set -euo pipefail
 
 rapids-logger "Downloading artifacts from previous jobs"
-PYTHON_CHANNEL=$(rapids-download-conda-from-s3 python)
+# rapids-download-conda-from-s3 doesn't have handling for noarch packages
+# these are uncommon enough for RAPIDS that we handle this one as a special case
+PYTHON_CHANNEL="/tmp/python_channel"
+rapids-download-from-s3 "dask-cuda_conda_python_noarch.tar.gz" "${PYTHON_CHANNEL}"
```
jameslamb, line 10: > these are uncommon enough for RAPIDS

Thanks for calling this out... I haven't really thought about it before! There are several other projects across RAPIDS that are candidates for `noarch` (not dependent on CPU architecture or Python minor version):

* `cudf-polars`
* `cuxfilter`
* `jupyterlab_nvdashboard`
* `nx-cugraph`

Those are all publishing `*-py3-none-any-*` wheels, but for conda are doing what you found here... publishing 1 package per Python minor version.

* where you can see the wheels: https://pypi.anaconda.org/rapidsai-wheels-nightly/simple/nx-cugraph
* where you can see the variants of conda packages: https://anaconda.org/rapidsai-nightly/nx-cugraph/files

And I suspect some of these that are conda-only (no wheels) could also be candidates for `noarch` treatment:

* `custreamz`
* `cugraph-service-client`

I'm not sure if all of those projects producing one conda package per Python version is intentional, or if it was just convenient. I THINK it's desirable to switch them to `noarch` and that that's part of what @vyasr was documenting here: https://github.com/rapidsai/build-planning/issues/43

It's probably worth investigating that a bit before switching `dask-cuda` over to `noarch`. And if we do want to do this, that issue would be a good place to capture the work (including anything needed in `gha-tools` to avoid workarounds like what you did here).

I know it's annoying to get a comment like "maybe we should make a change across a bunch of repos" on a PR where you weren't necessarily trying to set a pattern... if you don't want to block this on that conversation, you could revert the `noarch` stuff and keep this PR more tightly focused on just changing over to `rattler-build`.

conda/recipes/dask-cuda/recipe.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - dask-cuda-config = dask_cuda.cli:config
+
+requirements:
+  host:
+    - python
+    - pip
+    - rapids-build-backend >=0.3.0,<0.4.0dev0
+    - setuptools >=64.0.0
+  run:
+    - python
```
jameslamb, line 35: I don't know the `rattler-build` recipe syntax well so can't make an exact suggestion, but if this is going to be `noarch` I think we should set a floor on `python` to the lowest version we test against, something like:

```yaml
    - python >=3.10
```

I'm not sure how you want to accomplish that in CI... maybe change the `matrix_filter` in `.github/workflows` to build on the OLDEST Python instead of the latest, then rely on environment variables in the container? Figuring out a pattern for that would be part of the work for https://github.com/rapidsai/build-planning/issues/43, I think.

jameslamb (changes requested):
I think the `noarch` part of this deserves a bit more discussion. Added some thoughts. I'm leaving a blocking review because I think those should be resolved before this is merged.

ci/build_docs.sh
jameslamb: Ok yep, works for me. Thanks for considering it.

conda/recipes/dask-cuda/recipe.yaml
gforsyth: I think for now I'm comfortable setting this equal to the value of `python-requires` in the `pyproject.toml`, so at least we're self-consistent across conda and pip

gforsyth: Misread the above comment -- we aren't `noarch` anymore so this doesn't apply at the moment.

Setting the lower bound equal to `python-requires` is a reasonable pattern, I think, although at the moment it doesn't work because the `rattler` jinja parser sees the `-` as a minus sign and fails out.

gforsyth: https://github.com/prefix-dev/rattler-build/issues/1539

jameslamb: Ah yep! Now that we're deferring `noarch` to later, nothing needs to be done for this comment. We'll get a Python 3.10 package that'll get built with a `python =3.10.*` or similar constraint, and similar for Python 3.11 and 3.12.

Thanks for opening that `rattler` issue, I subscribed there.

jameslamb (approved):
Approving now that this is a straight port, everything looks good to me. Thanks for talking through the `noarch` stuff with me.

conda/recipes/dask-cuda/recipe.yaml
gforsyth: Just to update on the `rattler-build` issue, it's possible to escape keys with `-` and other problem characters by using square brackets to look up fields, e.g.

```yaml
py_requires: ${{ load_from_file("pyproject.toml")["project"]["requires-python"] }}
```

conda/recipes/dask-cuda/meta.yaml
```diff
@@ -1,60 +0,0 @@
...
-    - python
-    - pip
-    {% for r in data["build-system"]["requires"] %}
-    - {{ r }}
-    {% endfor %}
-  run:
-    - python
-    {% for r in data["project"]["dependencies"] %}
-    - {{ r }}
-    {% endfor %}
```
jakirkham, line 39: Previously we were able to load dependencies from `pyproject.toml`

conda/recipes/dask-cuda/recipe.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - click >=8.1
+    - numba >=0.59.1,<0.61.0a0
+    - numpy >=1.23,<3.0a0
+    - pandas >=1.3
+    - pynvml >=12.0.0,<13.0.0a0
+    - rapids-dask-dependency =${{ minor_version }}
+    - zict >=2.0.0
```
jakirkham, line 44: However it looks like we have expanded this for `rattler-build`

Is there not a way to load this from `pyproject.toml` with `rattler-build`?

jakirkham (commented):
Thanks Gil! 🙏

Had a question about one of the changes below

conda/recipes/dask-cuda/recipe.yaml
gforsyth: I don't believe so, currently.  There's not support for the jinja looping constructs to handle splatting out a list of dependencies.  I think updates to the dependency generator are the long-term solution here, now that the conda recipes are in standardized yaml

PR #1462, opened 2025-04-07
Workaround hang with dask main and explicit comms

TomAugspurger:
Dask>2025.3.0 will delay materialization of the task graph for more operations, including expressions and HighLevelGraphs. The final task graph will now be built on the scheduler.

The explicit-comms shuffle relied on the low-level graph being materialized on the client. Parts of the task generation in `ECShuffle._lower` use code that, apparently, doesn't run well on the scheduler in the context of graph materialization.

This works around that issue by continuing to materialize the task graph on the client before sending it to the scheduler. We only do this when we're using explicit comms and the result to compute contains a shuffle task.

https://github.com/rapidsai/dask-upstream-testing/issues/37#issuecomment-2779798670 contains more information on the issue.

Given the fragility of this, a followup PR will deprecate using explicit comms shuffle through `df.shuffle()`. It'll still be available through `explicit_comms.dataframe.shuffle(df)`.

### DIFF
--- a/dask_cuda/explicit_comms/dataframe/shuffle.py
+++ b/dask_cuda/explicit_comms/dataframe/shuffle.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import functools
 from collections import defaultdict
 from math import ceil
 from operator import getitem
@@ -25,6 +26,7 @@ from distributed import wait
 from distributed.protocol import nested_deserialize, to_serialize
 from distributed.worker import Worker
 
+from ..._compat import DASK_2025_4_0
 from .. import comms
 
 T = TypeVar("T")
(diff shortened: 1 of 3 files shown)


dask_cuda/tests/test_explicit_comms.py
```diff
@@ -21,18 +21,16 @@
...
+    _contains_shuffle_expr,
+    shuffle as explicit_comms_shuffle,
+)
 from dask_cuda.utils_test import IncreasedCloseTimeoutNanny
 
 mp = mp.get_context("spawn")  # type: ignore
 ucp = pytest.importorskip("ucp")
 
 
-# Set default shuffle method to "tasks"
```
TomAugspurger, line 31: I'm not sure why, but the tests were hanging when this was set. It passes fine without this set.

I wasn't able to reproduce the hang outside of pytest.

TomAugspurger:
I'm doing one more run with dask `main` locally, but things should be ready for review.

dask_cuda/explicit_comms/dataframe/shuffle.py
```diff
@@ -584,6 +586,128 @@ def _use_explicit_comms() -> bool:
...
+            *args,
+            traverse=traverse,
+            optimize_graph=optimize_graph,
+            scheduler=scheduler,
+            get=get,
+            **kwargs,
+        )
+
+
+class ECShuffle(dask_expr._shuffle.TaskShuffle):
```
TomAugspurger, line 675: This is just moved to the top-level of the module, rather than defining the class inside a function. I moved this will debugging some things. I can move it back if we want a minimal diff, but I think putting it at the module level is a bit nicer.

pentschev (approved):
Thanks @TomAugspurger , please make sure to apply Mads' suggestion but otherwise LGTM.

TomAugspurger:
> If this gets too fragile, I think we should consider disabling explicit-comms in Dask 2025.4+

I'll put up a PR later today discussing this more, but I think we will need to deprecate the explicit-comms shuffle you get through `df.shuffle(...)`. The shuffle using `dask_cuda.explicit_comms.dataframe.shuffle(df)` should be OK to keep supporting.

fjetter:
FYI if the materialization on scheduler is a problem / becomes a bigger problem, it should be straight forward to put in a toggle to control this

TomAugspurger:
Thanks. I considered suggesting that, but opted against it because ideally we *wouldn't* need to materialize it. If we or others come across cases where client-side materialization is needed then I'd be happy to put something together in dask / distributed. But I don't think this case alone is sufficient.

PR #1463, opened 2025-04-09
Add missing `runs-on` to `ucxx-python-tests`

pentschev:
The `runs-on` option is required and may use `ubuntu-latest`, as it is just used to launch another workflow.


ajschmidt8 (approved):

pentschev:
Thanks @ajschmidt8 !

PR #1465, opened 2025-04-09
Move env out of with context workflow context

pentschev:
The `env` context in GHA workflows cannot be within `with` context, causing errors as below:

```
The workflow is not valid. .github/workflows/test_ucxx.yaml (Line: 26, Col: 9): A mapping was not expected
```

### DIFF
--- a/.github/workflows/test_ucxx.yaml
+++ b/.github/workflows/test_ucxx.yaml
@@ -16,12 +16,12 @@ jobs:
   ucxx-python-tests:
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/conda-python-tests.yaml@branch-25.06
+    env:
+      GH_TOKEN: ${{ github.token }}
+      GHA_DASK_UCXX_ONLY: true
     with:
       build_type: ${{ inputs.build_type }}
       branch: ${{ inputs.branch }}
       date: ${{ inputs.date }}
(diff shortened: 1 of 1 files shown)


ajschmidt8 (approved):
Approving, though I'm not entirely sure if this is valid. I don't think you can set the `env` block like this on a reusable workflow.

It might be worth testing this pattern in a personal private repository first. Otherwise you may have to open another PR to undo the change.

pentschev:
> Approving, though I'm not entirely sure if this is valid. I don't think you can set the `env` block like this on a reusable workflow.
> 
> It might be worth testing this pattern in a personal private repository first. Otherwise you may have to open another PR to undo the change.

Thanks AJ, and that's a good idea, will do that first.

TomAugspurger:
I've used tools like https://github.com/rhysd/actionlint / https://rhysd.github.io/actionlint/ in the past to validate these schema things quickly. It does look like `env` isn't valid under `jobs.<job>`.

pentschev:
Thanks @TomAugspurger , indeed that seems invalid. After some further discussion with @ajschmidt8 and some headache getting things properly setup it probably makes more sense to use a custom nightly workflow instead, doing that in https://github.com/rapidsai/dask-cuda/pull/1470 and closing this.

PR #1467, opened 2025-04-16
Reference public cuDF Python objects during dispatch registration

mroeschke:
It appears some non-public subclasses were being referenced during registration such as `cudf.BaseIndex` or `cudf.core.frame.Frame`. It would be great to instead reference the public objects in cuDF python that should more commonly appear so cuDF has liberty to modify these non-public subclasses in the future.


TomAugspurger (approved):
Thanks.

I think the primary downside is if cudf happens to add a new Index type in the future that doesn't subclass `Index`. That seems infrequent / low probability enough to be fine to handle manually.

mroeschke:
Yeah that's fair. If we're OK with dask-cuda responding to changes in cuDF then we can keep this as is. Closing this PR then.

PR #1468, opened 2025-04-17
Register OwnerWithCAI with dask.utils.sizeof

TomAugspurger:
Fixes the failing proxy test seen in https://github.com/rapidsai/dask-cuda/pull/1467.

---

The first change we see is in `get_device_memory_objcects.dispatch`

```python
import cudf
import dask_cuda.get_device_memory_objects
import dask.sizeof


def main():
    s = cudf.Series([1, 2, 3])
    
    items = dask_cuda.get_device_memory_objects.dispatch(s)
    print(items, dask.sizeof.sizeof(s), dask.sizeof.sizeof(items[0]))


if __name__ == "__main__":
    main()
```

Previously, that found the RMM DeviceBuffer. The sizeof the Series was 48 and the sizeof the DeviceBuffer was 24:

```
./old/bin/python ./debug.py 
[<rmm.pylibrmm.device_buffer.DeviceBuffer object at 0x7fb53dfd0940>] 48 24
```

Now, we get a `pylibcudf.column.OwnerWithCAI`. The sizeof the Series is still 48, but now the sizeof this OwnerWithCAI thing is 56, I think because Dask is using some default sizeof for unknown objects?

```
./new/bin/python ./debug.py 
[<pylibcudf.column.OwnerWithCAI object at 0x7fec2b6d5a00>] 48 56
```

This causes the test to fail because the program sees we've crossed some memory limit (thanks to 56 > 24) and spills the cupy array, which breaks the identity assert.

### DIFF
--- a/dask_cuda/get_device_memory_objects.py
+++ b/dask_cuda/get_device_memory_objects.py
@@ -1,3 +1,5 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.
+
 from typing import Set
 
 from dask.sizeof import sizeof
@@ -140,3 +142,15 @@ def register_cupy():  # NB: this overwrites dask.sizeof.register_cupy()
     @sizeof.register(cupy.ndarray)
     def sizeof_cupy_ndarray(x):
         return int(x.nbytes)
+
+
+@sizeof.register_lazy("pylibcudf")
+def register_pylibcudf():
+    import cupy
+    import pylibcudf
+
+    @sizeof.register(pylibcudf.column.OwnerWithCAI)
+    def sizeof_owner_with_cai(x):
+        # OwnerWithCAI implements __cuda_array_interface__
+        # so this should always be zero-copy
+        return cupy.array(x, copy=False).nbytes
--- a/dask_cuda/get_device_memory_objects.py
+++ b/dask_cuda/get_device_memory_objects.py
@@ -153,5 +153,5 @@ def register_pylibcudf():
         # OwnerWithCAI implements __cuda_array_interface__ so this should always
         # be zero-copy
         col = pylibcudf.column.Column.from_cuda_array_interface(x)
-        # col.data() returns a gpumemoryview, which knows the size in bytes 
+        # col.data() returns a gpumemoryview, which knows the size in bytes
         return col.data().nbytes
--- a/dask_cuda/get_device_memory_objects.py
+++ b/dask_cuda/get_device_memory_objects.py
@@ -153,4 +153,5 @@ def register_pylibcudf():
         # OwnerWithCAI implements __cuda_array_interface__ so this should always
         # be zero-copy
         col = pylibcudf.column.Column.from_cuda_array_interface(x)
+        # col.data() returns a gpumemoryview, which knows the size in bytes 
         return col.data().nbytes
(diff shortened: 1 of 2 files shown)


dask_cuda/get_device_memory_objects.py
```diff
@@ -140,3 +142,15 @@ def sizeof_cupy_base_memory(x):
...
     def sizeof_cupy_ndarray(x):
         return int(x.nbytes)
+
+
+@sizeof.register_lazy("pylibcudf")
+def register_pylibcudf():
+    import cupy
+    import pylibcudf
+
+    @sizeof.register(pylibcudf.column.OwnerWithCAI)
```
TomAugspurger, line 152: I think we don't need any compatibility code here, since both this change and the cudf change are going in 25.6.0.

dask_cuda/get_device_memory_objects.py
```diff
@@ -140,3 +142,15 @@ def sizeof_cupy_base_memory(x):
     @sizeof.register(cupy.ndarray)
     def sizeof_cupy_ndarray(x):
         return int(x.nbytes)
+
+
+@sizeof.register_lazy("pylibcudf")
+def register_pylibcudf():
+    import cupy
```
TomAugspurger, line 149: Oh, cupy isn't a (required) dependency. I'll find another way to get nbytes out of this.

rjzamora:
>I think because Dask is using some default sizeof for unknown objects?

I haven't caught up on this issue/fix, but I wanted to make sure you're aware of https://github.com/rapidsai/cudf/pull/18133

Peter recently (early March) dealt with a similar problem in cudf.

TomAugspurger:
Thanks. I just switched over to using `gpumemoryview().nbytes` to compute the `nbytes` instead of cupy.

dask_cuda/get_device_memory_objects.py
```diff
@@ -140,3 +142,15 @@ def sizeof_cupy_base_memory(x):
...
+@sizeof.register_lazy("pylibcudf")
+def register_pylibcudf():
+    import pylibcudf
+
+    @sizeof.register(pylibcudf.column.OwnerWithCAI)
+    def sizeof_owner_with_cai(x):
+        # OwnerWithCAI implements __cuda_array_interface__ so this should always
+        # be zero-copy
+        col = pylibcudf.column.Column.from_cuda_array_interface(x)
+        return col.data().nbytes
```
rjzamora, line 156: Nit: Maybe comment that this is leaning on `gpumemoryview.nbytes`?

rjzamora (approved):
Thanks @TomAugspurger ! I think this makes sense.

dask_cuda/get_device_memory_objects.py
```diff
@@ -140,3 +142,16 @@ def sizeof_cupy_base_memory(x):
...
+@sizeof.register_lazy("pylibcudf")
+def register_pylibcudf():
+    import pylibcudf
+
+    @sizeof.register(pylibcudf.column.OwnerWithCAI)
+    def sizeof_owner_with_cai(x):
+        # OwnerWithCAI implements __cuda_array_interface__ so this should always
+        # be zero-copy
+        col = pylibcudf.column.Column.from_cuda_array_interface(x)
+        # col.data() returns a gpumemoryview, which knows the size in bytes
```
TomAugspurger, line 156:
```suggestion
        # col.data() returns a gpumemoryview, which knows the size in bytes
```

pentschev (approved):
LGTM, thanks @TomAugspurger .

PR #1470, opened 2025-04-22
Add UCXX tests as separate nightly workflow

pentschev:
Use a separate nightly workflow to launch UCXX tests. We have previously attempted to fork a job from the regular nightly workflow but that proved more challenging than it's worth the trouble, so now we switch to a cron-based job that is known to work and suffices for the current requirements.

### DIFF
--- a/.github/workflows/nightly-pipeline.yaml
+++ b/.github/workflows/nightly-pipeline.yaml
@@ -1,6 +1,8 @@
-name: test
+name: Trigger Nightly Build and Test Pipelines
 
 on:
+  schedule:
+    - cron: "0 5 * * *"  # 5am UTC / 1am EST
   workflow_dispatch:
     inputs:
       branch:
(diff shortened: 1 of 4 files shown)


.github/workflows/nightly-pipeline.yaml
```diff
@@ -0,0 +1,37 @@
...
+        type: string
+      build_type:
+        type: string
+        default: branch
+
+jobs:
+  build:
+    uses: ./.github/workflows/build.yaml
+    with:
+      build_type: ${{ inputs.build_type || 'branch' }}
```
bdice, line 25: How did you choose between `'branch'` and `'nightly'` here?

Do we need builds and tests to run twice every night? We are already doing nightly builds and tests here: https://github.com/rapidsai/workflows/blob/301d42d33d484de48c1001858470846db7d716e3/.github/workflows/nightly-pipeline.yaml#L467-L484

pentschev: I just replicated what we [were doing in rapidsmpf](https://github.com/rapidsai/rapidsmpf/blob/branch-25.06/.github/workflows/nightly-pipeline.yaml) while the repository wasn't public and that I know works. I know we have the _regular_ dask-cuda pipeline in the workflows repo, however, they [currently disable UCXX tests](https://github.com/rapidsai/dask-cuda/blob/6f929e69632b182c9582b42d3dea67c108730583/ci/run_pytest.sh#L19) and I need now to enable them without them causing failures to regular nightlies. I want to gain confidence by manual inspection of the results that UCXX is now in good state so we can enable them permanently. Initially I've discussed with @ajschmidt8 trying to [spawn a separate process](https://github.com/rapidsai/dask-cuda/pull/1470/files#diff-245392b692a50c38ecab4381b118862db514035c10983f3bd4f4b7f1f4be4692L28-L46) that wouldn't report results but that was getting far too complicated to get right and it's not worth the time since this is anyway going to be temporary.

bdice: That makes more sense now. Thanks for explaining.

.github/workflows/test_ucxx.yaml
```diff
@@ -4,24 +4,26 @@ on:
...
     with:
       build_type: ${{ inputs.build_type }}
       branch: ${{ inputs.branch }}
       date: ${{ inputs.date }}
       sha: ${{ inputs.sha }}
-      script: ci/test_python.sh
-      env:
-        GH_TOKEN: ${{ github.token }}
-        GHA_DASK_UCXX_ONLY: true
+      gh_token: ${{ github.token }}
```
ajschmidt8, line 28:
```suggestion
```

This isn't a valid input on the `conda-python-tests.yaml` workflow here: https://github.com/rapidsai/shared-workflows/blob/1fdd8cfecadd8223a0741c19a1e76804a10b0f07/.github/workflows/conda-python-tests.yaml#L1-L33

Therefore we'll need to remove it.

ajschmidt8 (changes requested):

.github/workflows/test_ucxx.yaml
pentschev: Good catch, removed.

ajschmidt8 (approved):

pentschev:
Thanks @ajschmidt8 and @bdice !

PR #1471, opened 2025-04-24
Make workflows reusable

pentschev:
Some necessary changes were missed in #1470: make `build` and `test-ucxx` reusable workflows so that they can be called from `nighly-pipeline`.


pentschev:
Thanks @ajschmidt8 !

PR #1472, opened 2025-04-24
Add separate `build-ucxx` workflow

pentschev:
In #1471 the `build` workflow was made reusable but a separate `build-ucxx` workflow is necessary for `nightly-pipeline` because uploading conda/wheels build is not supported for a custom workflow. This change reverts making `build` reusable and use a new `build-ucxx` workflow only for the `conda-python-build` job.

### DIFF
--- a/.github/workflows/build-ucxx.yaml
+++ b/.github/workflows/build-ucxx.yaml
@@ -1,0 +1,50 @@
+name: build
+
+on:
+  push:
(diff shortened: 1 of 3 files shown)


ajschmidt8 (approved):

pentschev:
Thanks again @ajschmidt8 !

PR #1475, opened 2025-05-02
feat: add support for Python 3.13

gforsyth:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/120

This PR adds support for Python 3.13.

## Notes for Reviewers

This is part of ongoing work to add Python 3.13 support across RAPIDS.
It temporarily introduces a build/test matrix including Python 3.13, from https://github.com/rapidsai/shared-workflows/pull/268.

A follow-up PR will revert back to pointing at the `branch-25.06` branch of `shared-workflows` once all
RAPIDS repos have added Python 3.13 support.

### This will fail until all dependencies have been updated to Python 3.13

CI here is expected to fail until all of this project's upstream dependencies support Python 3.13.

This can be merged whenever all CI jobs are passing.

### DIFF
--- a/.github/workflows/build-ucxx.yaml
+++ b/.github/workflows/build-ucxx.yaml
@@ -42,7 +42,7 @@ concurrency:
 jobs:
   conda-python-build:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@branch-25.06
+    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@python-3.13
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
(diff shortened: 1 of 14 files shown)


pentschev:
If that gets too complicated I'm fine if we can just disable cudf/dask-cudf dependency testing to get this merged first, they are only hard dependencies _for testing_, so if building a custom matrix for 3.13 is easy we could just do that without cudf/dask-cudf until they get their own packages.

gforsyth:
> If that gets too complicated I'm fine if we can just disable cudf/dask-cudf dependency testing to get this merged first, they are only hard dependencies _for testing_, so if building a custom matrix for 3.13 is easy we could just do that without cudf/dask-cudf until they get their own packages.

Ok, that sounds easier and faster and less headache-inducing.  Thanks @pentschev !

bdice:
Yes, we typically disable cudf tests in this repo when circular dependencies like this come up.

PR #1476, opened 2025-05-02
Add function to get device handle and fix MIG handle

pentschev:
This change reduces duplication on code getting NVML device handles and fixes one instance where getting the MIG handle was not previously done correctly. It has a small breaking change on the `get_gpu_uuid_from_index` function being renamed to `get_gpu_uuid`, but there are apparently no uses outside of Dask-CUDA.


madsbk (approved):
LGTM

pentschev:
Thanks @madsbk !

PR #1479, opened 2025-05-05
chore: bump upper bound of numba for python 3.13 support

gforsyth:
xref #rapidsai/build-planning#120

This upper pin on `numba` is preventing a solve in `cudf` for the Python 3.13 environment


jameslamb (approved):
Approving so you can merge when ready without a re-review, but you need to update the `conda` recipe as well: https://github.com/rapidsai/dask-cuda/blob/84959a136e4b75191a7bf127091267cd3490f2b4/conda/recipes/dask-cuda/recipe.yaml#L39

Recommend a quick check like this in case there are other places:

```shell
git grep -i numba
```

gforsyth:
Thanks for catching that, @jameslamb !

PR #1487, opened 2025-05-16
Fix persist usage in tests

TomAugspurger:
Recent versions of dask removed the ability to call the synchronous
`<collection>.persist()` with an async client.

Use `client.persist()` instead.

part of https://github.com/rapidsai/rapids-dask-dependency/issues/103


jacobtomlinson (approved):

madsbk (approved):
Looks good

pentschev (approved):
LGTM.

PR #1492, opened 2025-05-20
Fixed flaky `test_cudf_cluster_device_spill` test

### LINKED ISSUE #1490: Flaky test `dask_cuda/tests/test_spill.py::test_cudf_cluster_device_spill[True-params2]`
Spotted in https://github.com/rapidsai/dask-cuda/pull/1488. This fails ~5-20% of the time for me locally.

I rewrote some of the asserts to make it easier to see where the failure is:

```
  File "/home/nfs/toaugspurger/gh/rapidsai/dask-cuda/dask_cuda/tests/test_spill.py", line 108, in device_host_file_size_matches
    assert byte_sum <= total_bytes + device_overhead + host_overhead + disk_overhead, (byte_sum, total_bytes, device_overhead, host_overhead, disk_overhead)
      ^^^^^^^^^^^^^^^^^
AssertionError: (3456902, 0, 0, 0, 0)
```

Some print statements shows the issue is in `dhf.disk` and the files on disk:

```
2025-05-19 19:08:11,455 - distributed.worker - INFO - Run out-of-band function 'worker_assert'
dhf.device_buffer.fast.total_weight=0
dhf.host_buffer.fast.total_weight=0
sum(file_size)=3456902
2025-05-19 19:08:11,455 - distributed.worker - WARNING - Run Failed
```

https://github.com/rapidsai/dask-cuda/blob/cd50675cf95abe8650cc5c538fc278ea7e7b73ad/dask_cuda/tests/test_spill.py#L88-L94.

Will look into this some more.

TomAugspurger:
The test fails because we're asserting that the worker's local data directory is supposed to be empty. There seems to be some a race condition between when the test thread checks the size of that directory and when whatever thread does the cleanup actually runs.

I think the intention of the test is to ensure that the `del cdf` and `gc.collect()` causes the worker memory (and files spilled to disk) to be cleaned up on the worker. However `__del__` is synchronous and the cleanup is happening asynchronously on the worker. `__del__` just emits the RPC instructing the worker to free the memory and clean up the files, but doesn't block until that happens.

This structure updates the test to run the check in a loop and only fail after the check fails to pass after some timeout duration (3 seconds). With these changes the test passes 100 times using `pytest-repeat`. Previously it usually failed within 5-ish repeats.

While not strictly necessary, I've restructured this part of the test a bit. Previously, we used `client.run` to run a function that collected some data on the worker and made some assertions about it (in this case, that the worker data was empty). I've restructured the test to just collect the data on the worker and make the assertion on the client. This makes the tracebacks emitted by pytest nicer and aids in debugging since you can see a bit about why the test failed. We can consider doing this in more places in the tests, but IMO it's not a high priority.

Closes #1490

### DIFF
--- a/dask_cuda/tests/test_spill.py
+++ b/dask_cuda/tests/test_spill.py
@@ -3,14 +3,16 @@
 import gc
 import os
 from time import sleep
+from typing import TypedDict
 
 import pytest
 
 import dask
 from dask import array as da
-from distributed import Client, wait
+from distributed import Client, Worker, wait
 from distributed.metrics import time
 from distributed.sizeof import sizeof
+from distributed.utils import Deadline
 from distributed.utils_test import gen_cluster, gen_test, loop  # noqa: F401
 
 import dask_cudf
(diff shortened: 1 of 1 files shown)


pentschev (approved):
LGTM, thanks @TomAugspurger for the great analysis and nice improvements!

PR #1494, opened 2025-05-20
Add UCXX blocking tests in nightly CI runs

pentschev:
Enable blocking progress mode tests for the UCXX test matrix. The default mode in UCXX is `thread`, but we want to test `blocking` too, as this is the equivalent to UCX-Py's default and will likely be used as initial default when switching from UCX-Py to UCXX.

### DIFF
--- a/.github/workflows/build-ucxx.yaml
+++ b/.github/workflows/build-ucxx.yaml
@@ -1,4 +1,4 @@
-name: build
+name: Build Dask-CUDA for UCXX
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


.github/workflows/build-ucxx.yaml
```diff
@@ -1,4 +1,4 @@
-name: build
+name: Build Dask-CUDA for UCXX
```
pentschev, line 1: The current name is a duplicate of the one in the "default" `build.yaml`, which causes a clash in the "Actions" panel and workflow name matching.

Screenshot 2025-05-20 at 22 39 24

jameslamb: ah yeah! This is a good change, thanks for the explanation.

pentschev:
Thanks @jameslamb !

PR #1502, opened 2025-05-30
Add fix for devices that do not have memory resources

viclafargue:
cc @pentschev


pentschev (changes requested):
This unfortunately resolves only the more superficial error but falls short for the rest of Dask-CUDA. Additionally, it's not clear if this is the proper/final solution, so holding off on this while we discuss this offline.

pentschev (approved):
I've updated the docstrings to document what `None` means, but I think this is a reasonable solution for now. I'll work on follow-up PR(s) to address other Dask-CUDA aspects.

PR #1505, opened 2025-06-10
Handle system on a chip (SoC) devices

pentschev:
Certain devices like system on a chip (SoC) have no dedicated memory resources, and thus share the same unique memory region as the CPU. In those devices spilling device<->host doesn't make sense, since they are the same memory region and thus accessible by both, thus we disable spilling where appropriate.

### DIFF
--- a/dask_cuda/tests/test_dask_cuda_worker.py
+++ b/dask_cuda/tests/test_dask_cuda_worker.py
@@ -316,7 +316,7 @@ def test_cudf_spill(loop):  # noqa: F811
                     assert v == 2
 
 
-@pytest.mark.skip_if_no_device_memory(
+@pytest.mark.skip_if_device_memory(
     "Devices with dedicated memory resources cannot test error"
 )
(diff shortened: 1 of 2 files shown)


dask_cuda/tests/test_dask_cuda_worker.py
```diff
@@ -58,12 +57,43 @@ def get_visible_devices():
                 for v in result.values():
                     del expected[v]
 
+                assert len(expected) == 0
+
+
+def test_memory_limit_and_nthreads(loop):  # noqa: F811
```
pentschev, line 63: This test had to be split from `test_cuda_visible_devices` because `data` is now determined _per-device_ and not globally based on device 0. Therefore, testing devices that are unavailable becomes impossible as NVML is unable to query device memory from devices that don't exist, and `DASK_CUDA_TEST_DISABLE_DEVICE_SPECIFIC` was specifically introduced to create a sort of a mock test.

PR #1506, opened 2025-06-11
Better `parse_device_memory_limit` special-case handling

pentschev:
Make `parse_device_memory_limit` return values more consistent. When `device_memory_limit=None`, also returns `None`, allowing to simplify certain conditionals in plugins. Now also returns `0` for values that evaluate to zero, instead of the entire memory device size, which can still be obtained with `"auto"` or using the fractional `1.0`.

Although this is technically a breaking change, there seems to be no usage of that function outside of Dask-CUDA in RAPIDS repositories.

### DIFF
--- a/dask_cuda/utils.py
+++ b/dask_cuda/utils.py
@@ -558,11 +558,24 @@ def parse_device_memory_limit(device_memory_limit, device_index=0, alignment_siz
         Number of bytes of alignment to use, i.e., allocation must be a multiple of
         that size. RMM pool requires 256 bytes alignment.
 
+    Returns
+    -------
+    The parsed memory limit in bytes, or `None` as convenience if `device_memory_limit`
+    is `None`.
+
     Examples
     --------
     >>> # On a 32GB CUDA device
     >>> parse_device_memory_limit(None)
-    34089730048
+    None
+    >>> parse_device_memory_limit(0)
+    0
+    >>> parse_device_memory_limit(0.0)
+    0
+    >>> parse_device_memory_limit("0 MiB")
+    0
+    >>> parse_device_memory_limit(1.0)
+    27271784038
     >>> parse_device_memory_limit(0.8)
     27271784038
     >>> parse_device_memory_limit(1000000000)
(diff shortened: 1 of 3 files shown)


dask_cuda/utils.py
```diff
@@ -558,11 +558,24 @@ def parse_device_memory_limit(device_memory_limit, device_index=0, alignment_siz
         Number of bytes of alignment to use, i.e., allocation must be a multiple of
         that size. RMM pool requires 256 bytes alignment.
 
+    Returns
+    -------
+    The parsed memory limit in bytes, or `None` as convenience if `device_memory_limit`
+    is `None`.
```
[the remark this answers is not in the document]
pentschev: Yeah, this confuses me a lot. Fixed in https://github.com/rapidsai/dask-cuda/pull/1506/commits/b564604816699489b4c0628dbcc53e091b6b7b11 .

TomAugspurger (commented):
Sounds good.

The CI failures at https://github.com/rapidsai/dask-cuda/actions/runs/15586184966/job/43893615212?pr=1506#step:11:4007 looks plausibly related to this PR:

```
  File "/__w/dask-cuda/dask-cuda/dask_cuda/tests/test_dask_cuda_worker.py", line 537, in test_worker_fraction_limits
    assert ret["device-memory-limit"] == int(device_total_memory * 0.1)
AssertionError: assert 0 == 2415289958
 +  where 2415289958 = int((24152899584 * 0.1))
```

pentschev:
> Sounds good.
> 
> The CI failures at https://github.com/rapidsai/dask-cuda/actions/runs/15586184966/job/43893615212?pr=1506#step:11:4007 looks plausibly related to this PR:

They are, took me a while to handle all those damn corner cases correctly, I think that's fixed now in https://github.com/rapidsai/dask-cuda/pull/1506/commits/395d998aacb80f3ed407dd304c1b3acd0ed74323, let's see if this won't break something else.

dask_cuda/utils.py
```diff
@@ -574,16 +586,60 @@ def parse_device_memory_limit(device_memory_limit, device_index=0, alignment_siz
...
+        v: int, float or str
+            The value to check if fractional.
+
+        Returns
+        -------
+        """
+        # Check if `x` matches exactly `int(1)` or `str("1")`, and is not a `float(1)`
+        is_one = lambda x: not isinstance(x, float) and (x == 1 or x == "1")
+
+        if not is_one(v):
```
TomAugspurger, line 609: I'd suggest just inlining the function rather than defining a lambda that's only used once.

```suggestion
        if not (not isinstance(v, float) and (v == 1 or v == "1")):
```

dask_cuda/utils.py
```diff
@@ -574,16 +586,60 @@ def parse_device_memory_limit(device_memory_limit, device_index=0, alignment_siz
...
+            with suppress(ValueError, TypeError):
+                v = float(v)
+                if isinstance(v, float) and 0.0 < v <= 1.0:
```
TomAugspurger, line 612: Can `isinstance(v, float)` ever be false here?

TomAugspurger (approved):
Small non-blocking questions / comments on the float parsing.

dask_cuda/utils.py
pentschev: Hmm, I think not, I'm uncertain why I had that added in the first place years ago, probably just a leftover from a previous change. Removed that in https://github.com/rapidsai/dask-cuda/pull/1506/commits/569dc3bba1dbf2689da9cdb837e647f57e50fd62, let's see if that doesn't cause any errors.

PR #1507, opened 2025-06-11
Fix docstring escaping

pentschev:
Some of the rST docs have incorrect escaping and this change should fix all of them.


TomAugspurger:
> I think with the changes from here this isn't necessary, but for the future, do you think we should add the same entry to Dask-CUDA's sphinx config?

Agreed, not necessary. Just letting you know that sometimes you'll see \`text\` that's deliberate.

PR #1509, opened 2025-06-12
Drop CUDA 11 checks and warnings

pentschev:
Since CUDA 11 is not supported anymore
(see https://github.com/rapidsai/dask-cuda/pull/1503), there's no need anymore for checks and documentation mentioning async memory version requirements.


pentschev:
Thanks @madsbk !

PR #1513, opened 2025-06-30
Use CUDA 12.9 in Conda, Devcontainers, Spark, GHA, etc.

jakirkham:
Addressing issues:
* https://github.com/rapidsai/build-planning/issues/195
* https://github.com/rapidsai/build-planning/issues/173

## Description

Use CUDA 12.9 throughout different build and test environments.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -16,7 +16,7 @@ To install the latest version of Dask-CUDA along with the CUDA Toolkit:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


pentschev (approved):
Like in UCX-Py, no other changes are needed because both CUDA 12.8 and 12.9 are [tested in CI](https://github.com/rapidsai/dask-cuda/blob/6c3223d7f82f4b2e7eb209a8485b1e32cb5357a7/dependencies.yaml#L87-L94).

LGTM. Thanks John!

PR #1516, opened 2025-07-02
Initialize UCXX when distributed-ucxx installed and `protocol=ucx`

pentschev:
UCXX is becoming the default UCX communicator via a proxy backend with https://github.com/rapidsai/rapids-dask-dependency/pull/116. This proxy backend now points `protocol="ucx"` to UCXX provided distributed-ucxx is installed, otherwise fallback to UCX-Py and emit a warning, and while UCX-Py is not archived provide a `protocol="ucx-old"` as fallback. This change ensures the proper (UCXX or UCX-Py) is initialized based on the specified protocol and whether distributed-ucxx is installed or not.

### DIFF
--- a/dask_cuda/initialize.py
+++ b/dask_cuda/initialize.py
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: Copyright (c) 2019-2025, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-License-Identifier: Apache-2.0
+
+import importlib
 import logging
 import os
 
@@ -22,21 +26,14 @@ def _create_cuda_context_handler():
         numba.cuda.current_context()
 
 
-def _create_cuda_context(protocol="ucx"):
-    if protocol not in ["ucx", "ucxx"]:
-        return
+def _initialize_ucx():
     try:
         # Added here to ensure the parent `LocalCUDACluster` process creates the CUDA
         # context directly from the UCX module, thus avoiding a similar warning there.
         try:
-            if protocol == "ucx":
-                import distributed.comm.ucx
+            import distributed.comm.ucx
 
-                distributed.comm.ucx.init_once()
-            elif protocol == "ucxx":
-                import distributed_ucxx.ucxx
-
-                distributed_ucxx.ucxx.init_once()
+            distributed.comm.ucx.init_once()
         except ModuleNotFoundError:
             # UCX initialization has to be delegated to Distributed, it will take care
             # of setting correct environment variables and importing `ucp` after that.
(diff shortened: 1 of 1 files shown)


dask_cuda/initialize.py
```diff
@@ -22,21 +26,51 @@ def _create_cuda_context_handler():
...
+            and not distributed.comm.ucx.cuda_context_created.has_context
+        ):
+            distributed.comm.ucx._warn_existing_cuda_context(ctx, os.getpid())
+
+        _create_cuda_context_handler()
 
-                distributed.comm.ucx.init_once()
-            elif protocol == "ucxx":
-                import distributed_ucxx.ucxx
+        if not distributed.comm.ucx.cuda_context_created.has_context:
```
wence-, line 55: If we got `ModuleNotFoundError` on line 37, then we arrive here and `distributed.comm.ucx` does not exist. How does this work?

dask_cuda/initialize.py
```diff
@@ -47,40 +81,40 @@ def _create_cuda_context(protocol="ucx"):
...
-            if (
-                ctx.has_context
-                and not distributed_ucxx.ucxx.cuda_context_created.has_context
-            ):
-                distributed_ucxx.ucxx._warn_existing_cuda_context(ctx, os.getpid())
+        if (
+            ctx.has_context
+            and not distributed_ucxx.ucxx.cuda_context_created.has_context
+        ):
+            distributed_ucxx.ucxx._warn_existing_cuda_context(ctx, os.getpid())
```
wence-, line 88: Same question here, these variables are all unbound if `ModuleNotFoundError` was caught.

dask_cuda/initialize.py
```diff
@@ -47,40 +81,40 @@ def _create_cuda_context(protocol="ucx"):
...
-                    )
+        if not distributed_ucxx.ucxx.cuda_context_created.has_context:
+            ctx = has_cuda_context()
+            if ctx.has_context and ctx.device_info != cuda_visible_device:
+                distributed_ucxx.ucxx._warn_cuda_context_wrong_device(
+                    cuda_visible_device, ctx.device_info, os.getpid()
+                )
 
     except Exception:
         logger.error("Unable to start CUDA Context", exc_info=True)
```
wence-, line 100: Should we make this exception slightly more informative? Or is `exc_info=True` doing that?

dask_cuda/initialize.py
```diff
@@ -47,40 +81,40 @@ def _create_cuda_context(protocol="ucx"):
...
+
+    has_ucxx = bool(importlib.util.find_spec("distributed_ucxx"))
+
+    if protocol == "ucxx" or (has_ucxx and protocol == "ucx"):
+        # With https://github.com/rapidsai/rapids-dask-dependency/pull/116,
+        # `protocol="ucx"` now points to UCXX (if distributed-ucxx is installed),
+        # thus call the UCXX initializer.
+        _initialize_ucxx()
+    else:
+        _initialize_ucx()
```
wence-, line 115: I was confused about where the warning is, but it's in rapids-dask-dependency, I think

wence- (commented):
I think the logic is not really changing, but I am nonetheless slightly confused

dask_cuda/initialize.py
pentschev: This is a tricky case, admittedly I didn't remember myself, great catch! What happens is `distributed.comm.ucx` is always available so long `distributed` is installed (which it always is for Dask-CUDA to work). However, `ucx-py` may not be installed and thus raise `ModuleNotFoundError` when `import ucp` fails. With that the `distributed.comm.ucx` module can be populated with the context information, _but_ it will indeed never come here because `protocol="ucx"` then becomes invalid. The decision to store context information in `distributed.comm.ucx` for all cases stems from the comment in lines 31-32, in this manner we prevent duplicate warnings and also allow other protocols such as `protocol="tcp"` to raise the warning as it used to be the case [long ago](https://github.com/rapidsai/dask-cuda/blob/a884233cab309c5b66a9e8ca5a102a634260c6d1/dask_cuda/initialize.py), and this has since become a bug for `protocol="tcp"`.

I have now fixed the bug for "other protocols" in https://github.com/rapidsai/dask-cuda/pull/1516/commits/66228b4c19339f44fffd30ae3c89edecb7c93e8a and added a TODO in https://github.com/rapidsai/dask-cuda/pull/1516/commits/c0c44eb559c46f88d0eb95966da79c2596dfb0fe, but that will need to be reworked for non-UCX cases after `distributed.comm.ucx` is removed along with UCX-Py, and filed https://github.com/rapidsai/dask-cuda/issues/1517 to update that later.

pentschev:
Thanks for the nice catch here @wence- , everything should be addressed now if you could take another look.

dask_cuda/initialize.py
wence-: Thanks

dask_cuda/initialize.py
wence-: No, this is ok.

wence- (approved):

pentschev:
Thanks again Lawrence!

PR #1519, opened 2025-07-04
Update UCX implementation checks

pentschev:
With UCXX being now the default there are some checks that need to be update to verify if UCXX or UCX-Py is running according to the environment's conditions. This change centralizes those checks which will allow us to remove most of the UCX-Py code by just removing the new utility functions `_get_active_ucx_implementation_name` and `get_ucx_implementation` along with the choice for the `"ucx-old"` protocol.

### DIFF
--- a/ci/test_python_ucxx.sh
+++ b/ci/test_python_ucxx.sh
@@ -40,6 +40,7 @@ set_exit_code() {
 }
 trap set_exit_code ERR
 set +e
+
 rapids-logger "pytest dask-cuda with ucxx in thread progress mode"
 UCXPY_PROGRESS_MODE=thread \
   ./ci/run_pytest.sh \
--- a/dask_cuda/utils.py
+++ b/dask_cuda/utils.py
@@ -1015,7 +1015,7 @@ def _get_active_ucx_implementation_name(protocol):
     ValueError
         If protocol is not a valid UCX protocol.
     """
-    has_ucxx = bool(importlib.util.find_spec("distributed_ucxx"))
+    has_ucxx = importlib.util.find_spec("distributed_ucxx") is not None
 
     if protocol == "ucxx" or (has_ucxx and protocol == "ucx"):
         # With https://github.com/rapidsai/rapids-dask-dependency/pull/116,
--- a/dask_cuda/worker_spec.py
+++ b/dask_cuda/worker_spec.py
@@ -91,7 +91,7 @@ def worker_spec(
             raise TypeError(
                 "Enabling InfiniBand or NVLink requires protocol='ucx', "
                 "protocol='ucxx' or protocol='ucx-old'"
-            )
+            ) from None
 
     if CUDA_VISIBLE_DEVICES is None:
         CUDA_VISIBLE_DEVICES = os.environ.get(


dask_cuda/utils.py
```diff
@@ -972,3 +988,41 @@ def convert(self, value, param, ctx):
...
+def _get_active_ucx_implementation_name(protocol):
+    """Get the name of active UCX implementation.
+
+    Determine what UCX implementation is being activated based on a series of
+    conditions. UCXX is selected if:
+    - The protocol is `"ucxx"`, or the protocol is `"ucx"` and the `distributed-ucxx`
+      package is installed.
+    UCX-Py is selected if:
+    - The protocol is `"ucx-old"`, or the protocol is `"ucx"` and the `distributed-ucxx`
+      package is not installed, in which case a `FutureWarning` is also raised.
```
TomAugspurger, line 1002: This `FutureWarning` is happening elsewhere? Or has that not been implemented yet?

dask_cuda/worker_spec.py
```diff
@@ -81,10 +84,14 @@ def worker_spec(
...
-    ) and protocol != "ucx":
-        raise TypeError("Enabling InfiniBand or NVLink requires protocol='ucx'")
+    if enable_tcp_over_ucx or enable_infiniband or enable_nvlink:
+        try:
+            _get_active_ucx_implementation_name(protocol)
+        except ValueError:
+            raise TypeError(
+                "Enabling InfiniBand or NVLink requires protocol='ucx', "
+                "protocol='ucxx' or protocol='ucx-old'"
+            )
```
TomAugspurger, line 94:
```suggestion
            raise TypeError(
                "Enabling InfiniBand or NVLink requires protocol='ucx', "
                "protocol='ucxx' or protocol='ucx-old'"
            ) from None
```

You probably want a `from None` here, to ensure that the `ValueError` doesn't end up in the traceback with the "while handling ... another exception was raised."

This is one of the reasons I don't like using exceptions for control flow like this, but that's not a big deal.

TomAugspurger (approved):
A couple minor, non-blocking suggestions / questions.

ci/test_python_ucxx.sh
```diff
@@ -40,31 +40,19 @@ set_exit_code() {
...
-  --cov-report=term \
-  -k "ucxx"
-
-rapids-logger "Run local benchmark with ucxx"
-./ci/run_benchmarks.sh -p ucxx
-
-export UCXPY_PROGRESS_MODE=blocking
-rapids-logger "pytest dask-cuda with ucxx in blocking progress mode"
-./ci/run_pytest.sh \
+rapids-logger "pytest dask-cuda with ucxx in thread progress mode"
```
bdice, line 43: I'd like a blank line here for logical separation.
```suggestion

rapids-logger "pytest dask-cuda with ucxx in thread progress mode"
```

dask_cuda/utils.py
```diff
@@ -972,3 +988,41 @@ def convert(self, value, param, ctx):
...
+    Returns
+    -------
+    The selected implementation type, either "ucxx" or "ucx-old".
+
+    Raises
+    ------
+    ValueError
+        If protocol is not a valid UCX protocol.
+    """
+    has_ucxx = bool(importlib.util.find_spec("distributed_ucxx"))
```
bdice, line 1018: I think this might be a more canonical way to write this check:
```suggestion
    has_ucxx = importlib.util.find_spec("distributed_ucxx") is not None
```

dask_cuda/utils.py
pentschev: [The `FutureWarning` is the one that happens in rapids-dask-dependency](https://github.com/rapidsai/rapids-dask-dependency/blob/88d91b634422c4c87e1db6fc85a3410c7685a351/rapids_dask_dependency/patches/distributed/comm/__rdd_patch_ucx.py#L566-L580).

pentschev:
Thanks @bdice and @TomAugspurger , since you both pre-approved and everything was addressed without changes and tests are passing I'll go ahead and merge it. Please let me know if I missed anything and I'll address in a follow-up PR.

PR #1527, opened 2025-08-01
Update rapids-build-backend to 0.4.1

KyleFromNVIDIA:
And remove MANIFEST.in, since it is no longer needed.

Issue: https://github.com/rapidsai/build-planning/issues/207


KyleFromNVIDIA:
Crap. The new version is successfully adding `GIT_COMMIT` to the wheel:

https://github.com/rapidsai/dask-cuda/actions/runs/16683244480/job/47226686689?pr=1527#step:12:388

but not to the Conda package:

https://github.com/rapidsai/dask-cuda/actions/runs/16683244480/job/47226825884?pr=1527#step:10:768

KyleFromNVIDIA:
Ugh. It was because I forgot to update the Conda recipe. This run should work.

jameslamb (approved):
Looks good to me, thanks!

Let's do the `rapids-build-backend` pinning update for every RAPIDS library using it, and the `MANIFEST.in` changes for every library relying on it to package these files. A tracking issue in `build-planning` similar to https://github.com/rapidsai/build-planning/issues/62, tying those PRs all together, would be helpful.

> _And remove MANIFEST.in, since it is no longer needed._

Linking to https://github.com/rapidsai/rapids-build-backend/pull/68, so anyone finding this from search or git blame in the future will know *why* `MANIFEST.in` was no longer necessary here.

PR #1531, opened 2025-08-08
Add numba-cuda as runtime dependency

pentschev:
The new `numba-cuda` package is replacing the old `numba.cuda` module from the `numba` package, which is now unmaintained. We recently discovered issues with getting the CUDA context on the new r580 driver that are resolved simply by having `numba-cuda` installed, therefore make that as a required dependency.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -150,16 +150,16 @@ dependencies:
           - zict>=2.0.0
       - output_types: [conda]
         packages:
-          - &numba_cuda numba-cuda>=0.14.0,<0.15.0a0
+          - &numba-cuda-dep numba-cuda>=0.14.0,<0.15.0a0
     specific:
       - output_types: [requirements, pyproject]
         matrices:
           - matrix: {cuda: "12.*"}
             packages:
-              - &numba_cuda_cu12 numba-cuda[cu12]>=0.14.0,<0.15.0a0
+              - &numba-cuda-cu12-dep numba-cuda[cu12]>=0.14.0,<0.15.0a0
           - matrix: # Fallback for no matrix
             packages:
-              - *numba_cuda_cu12
+              - *numba-cuda-cu12-dep
   test_python:
     common:
       - output_types: [conda, requirements, pyproject]


dependencies.yaml
```diff
@@ -148,6 +148,18 @@ dependencies:
...
+          - &numba-cuda-dep numba-cuda>=0.14.0,<0.15.0a0
+    specific:
+      - output_types: [requirements, pyproject]
+        matrices:
+          - matrix: {cuda: "12.*"}
+            packages:
+              - &numba-cuda-cu12-dep numba-cuda[cu12]>=0.14.0,<0.15.0a0
+          - matrix: # Fallback for no matrix
+            packages:
+              - *numba-cuda-cu12-dep
```
KyleFromNVIDIA, line 162:
```suggestion
      - output_types: [conda]
        packages:
          - &numba_cuda numba-cuda>=0.14.0,<0.15.0a0
    specific:
      - output_types: [requirements, pyproject]
        matrices:
          - matrix: {cuda: "12.*"}
            packages:
              - &numba_cuda_cu12 numba-cuda[cu12]>=0.14.0,<0.15.0a0
          - matrix: # Fallback for no matrix
            packages:
              - *numba_cuda_cu12
```

KyleFromNVIDIA (changes requested):

dependencies.yaml
pentschev: Can you clarify why that's superior/preferred? For reference, those names were copied [from cuDF](https://github.com/rapidsai/cudf/blob/branch-25.10/dependencies.yaml), so presumably having a consistent naming makes it easier for future maintenance.

dependencies.yaml
```diff
@@ -148,6 +148,18 @@ dependencies:
           - pynvml>=12.0.0,<13.0.0a0
           - rapids-dask-dependency==25.10.*,>=0.0.0a0
           - zict>=2.0.0
+      - output_types: [conda]
+        packages:
+          - &numba-cuda-dep numba-cuda>=0.14.0,<0.15.0a0
```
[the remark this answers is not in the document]
jameslamb: Ok we talked about this in https://github.com/rapidsai/ucxx/pull/481#discussion_r2263156844 and the answer was "no, not right now".

I'm ok with merging this as-is, thanks for the explanation.

dependencies.yaml
KyleFromNVIDIA: We typically use underscores for YAML anchors in `dependencies.yaml`. The one you linked is not consistent with the others, and should be changed as well.

KyleFromNVIDIA (approved):

bdice (approved):

pentschev:
Thanks all for the reviews.

Just for the sake of completeness here about https://github.com/rapidsai/dask-cuda/pull/1531#discussion_r2263306991, Kyle agreed we should go forward with the current naming and fix with rapids-reviser later if needed. Therefore, I'm triggering an auto-merge.

PR #1536, opened 2025-08-19
Build and test with CUDA 13.0.0

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/208

* uses CUDA 13.0.0 to build and test
* moves some dependency pins:
  - `numba-cuda`: `>=0.19.1,<0.20.0a0`
* adds `[cu12]` and `[cu13]` extras for wheels (https://github.com/rapidsai/dask-cuda/pull/1536#discussion_r2292073877)
* makes `cudf` and `dask_cudf` optional dependencies in tests (tests that need them are skipped if they're not present), and temporarily removes them from test environments for CUDA 13 CI (https://github.com/rapidsai/dask-cuda/pull/1536#issuecomment-3212474898)
* temporarily adds a test-time dependency on `dask[dataframe]` for wheels (https://github.com/rapidsai/dask-cuda/pull/1536#discussion_r2298528510)

Contributes to https://github.com/rapidsai/build-planning/issues/68

* updates to CUDA 13 dependencies in fallback entries in `dependencies.yaml` matrices (i.e., the ones that get written to `pyproject.toml` in source control)

## Notes for Reviewers

This switches GitHub Actions workflows to the `cuda13.0` branch from here: https://github.com/rapidsai/shared-workflows/pull/413

A future round of PRs will revert that back to `branch-25.10`, once all of RAPIDS supports CUDA 13.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -218,16 +218,6 @@ dependencies:
               - dask-cudf==25.10.*,>=0.0.0a0
           - matrix:
             packages:
-      - output_types: conda
-        matrices:
-          - matrix:
-              arch: x86_64
-            packages:
-              - numactl-devel-cos7-x86_64
-          - matrix:
-              arch: aarch64
-            packages:
-              - numactl-devel-cos7-aarch64
   depends_on_dask_cuda:
     common:
       - output_types: conda
--- a/conda/environments/all_cuda-129_arch-aarch64.yaml
+++ b/conda/environments/all_cuda-129_arch-aarch64.yaml
@@ -14,7 +14,6 @@ dependencies:
 - dask-cudf==25.10.*,>=0.0.0a0
 - distributed-ucxx==0.46.*,>=0.0.0a0
 - kvikio==25.10.*,>=0.0.0a0
-- numactl-devel-cos7-aarch64
 - numba-cuda>=0.19.1,<0.20.0a0
 - numpy>=1.23,<3.0a0
 - numpydoc>=1.1.0
--- a/conda/environments/all_cuda-129_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-129_arch-x86_64.yaml
@@ -14,7 +14,6 @@ dependencies:
 - dask-cudf==25.10.*,>=0.0.0a0
 - distributed-ucxx==0.46.*,>=0.0.0a0
 - kvikio==25.10.*,>=0.0.0a0
-- numactl-devel-cos7-x86_64
 - numba-cuda>=0.19.1,<0.20.0a0
 - numpy>=1.23,<3.0a0
 - numpydoc>=1.1.0
--- a/conda/environments/all_cuda-130_arch-aarch64.yaml
+++ b/conda/environments/all_cuda-130_arch-aarch64.yaml
@@ -12,7 +12,6 @@ dependencies:
 - cuda-version=13.0
 - distributed-ucxx==0.46.*,>=0.0.0a0
 - kvikio==25.10.*,>=0.0.0a0
-- numactl-devel-cos7-aarch64
 - numba-cuda>=0.19.1,<0.20.0a0
 - numpy>=1.23,<3.0a0
 - numpydoc>=1.1.0
--- a/conda/environments/all_cuda-130_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-130_arch-x86_64.yaml
@@ -12,7 +12,6 @@ dependencies:
 - cuda-version=13.0
 - distributed-ucxx==0.46.*,>=0.0.0a0
 - kvikio==25.10.*,>=0.0.0a0
-- numactl-devel-cos7-x86_64
 - numba-cuda>=0.19.1,<0.20.0a0
 - numpy>=1.23,<3.0a0
 - numpydoc>=1.1.0


bdice (approved):
All seems fine. We may have to skip cuDF tests or request an admin-merge to avoid a circular dependency.

dependencies.yaml
```diff
@@ -156,10 +157,13 @@ dependencies:
...
             packages:
-              - &numba_cuda_cu12 numba-cuda[cu12]>=0.19.0,<0.20.0a0
+              - numba-cuda[cu12]>=0.19.0,<0.20.0a0
+          - matrix: {cuda: "13.*"}
+            packages:
+              - &numba_cuda_cu13 numba-cuda[cu13]>=0.19.0,<0.20.0a0
           - matrix: # Fallback for no matrix
             packages:
-              - *numba_cuda_cu12
+              - *numba_cuda_cu13
```
jameslamb, line 166: Blegh, ok we have to do something here.

CUDA 12.9 wheel tests are failing like this:

```text
ERROR: Cannot install -r /tmp/requirements-test.txt (line 3) and numba-cuda[cu13]==0.19.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    numba-cuda[cu13] 0.19.0 depends on cuda-python==13.*; extra == "cu13"
    cudf-cu12 25.10.0a261 depends on cuda-python<13.0a0 and >=12.9.1
    cudf-cu12 25.10.0a260 depends on cuda-python<13.0a0 and >=12.9.1
```

([build link](https://github.com/rapidsai/dask-cuda/actions/runs/17137026241/job/48615775943?pr=1536#step:11:301))

The fallback matrix is used because of this:

https://github.com/rapidsai/dask-cuda/blob/0c83e513e2dfe6207ce6e57f3cf8c456b3b16ef2/pyproject.toml#L127

We distribute unsuffixed packages here (`dask-cuda`, not `dask-cuda-cu{12,13}`). Which has been fine, because this project's code isn't directly CUDA-major-version dependent and neither was its list of dependencies.... until https://github.com/rapidsai/dask-cuda/pull/1531

That PR introduced a CUDA-major-version-specific dependency like `numba-cuda[cu12]`. When I reviewed that, I forgot `dask-cuda` isn't suffixed 😭 

This is a problem because now we want to depend on `numba-cuda[cu12]` when targeting CUDA 12 and `numba-cuda[cu13]` when targeting CUDA 13.

The options I see are:

* depend on `numba-cuda` without extras
* start distributing suffixed packages like `dask-cuda-cu{12,13}`

It looks like those extras are probably necessary for correctness:

https://github.com/NVIDIA/numba-cuda/blob/d2827b7701289932ba390cdbf90642dcc7b2eeab/pyproject.toml#L26-L44

But I'm not sure. What do you think @bdice @brandon-b-miller @pentschev ?

jameslamb: We talked offline and @bdice mentioned it might be possible to factor the `numba.cuda` usage here out / down in a way that allows us to continue publishing unsuffixed `dask-cuda` wheels.

jakirkham:
> > _We may have to skip cuDF tests or request an admin-merge to avoid a circular dependency._
> 
> I was expecting we'd admin-merge for this one, as I think that's what we usually do for `dask-cuda`.

Typically have disabled the relevant tests and then come back with a follow up PR to re-enabling them

That way any relevant test failures that come up are scoped to a PR that is addressing them (and are not broadly effecting development)

dependencies.yaml
jameslamb: Trying with this change: #1537

jameslamb:
> _Typically have disabled the relevant tests and then come back with a follow up PR to re-enable them_

Ok I have definitely seen `dask-cuda` PRs just admin-merged, but we can try this. I do think what you're suggesting is preferable to every PR needing admin merges until `cudf` is done.

dependencies.yaml
```diff
@@ -148,6 +165,30 @@ dependencies:
...
           - rapids-dask-dependency==25.10.*,>=0.0.0a0
           - zict>=2.0.0
+  extra_cu12:
+    common:
+      - output_types: [pyproject]
+        packages:
+          # TODO: add 'cudf' and 'dask-cudf' back to this dependency list once there are CUDA 13 packages for those
+          # ref: https://github.com/rapidsai/dask-cuda/pull/1536#issuecomment-3212474898
+          # - cudf-cu12==25.10.*,>=0.0.0a0
+          # - dask-cudf-cu12==25.10.*,>=0.0.0a0
```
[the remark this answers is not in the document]
jameslamb: Thank you!

Posting for the record here on GitHub... in an offline conversation, we decided to do the following

* add a temporary test-time-only dependency on `dask[dataframe]` 
* remove that in the future when there are `cudf` CUDA 13 packages
* make `dask-cudf` depend on `dask[dataframe]`

dependencies.yaml
```diff
@@ -148,54 +165,69 @@ dependencies:
...
-              cuda_suffixed: "true"
+              arch: x86_64
             packages:
-              - cudf-cu12==25.10.*,>=0.0.0a0
-              - dask-cudf-cu12==25.10.*,>=0.0.0a0
-              - distributed-ucxx-cu12==0.46.*,>=0.0.0a0
-              - kvikio-cu12==25.10.*,>=0.0.0a0
-              - ucx-py-cu12==0.46.*,>=0.0.0a0
-              - &numba_cuda_cu12 numba-cuda[cu12]>=0.19.0,<0.20.0a0
+              - numactl-devel-cos7-x86_64
```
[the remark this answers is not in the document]
jameslamb: oh great, sure I'll try that here!

pentschev (approved):
Everything passing, great work! Thanks @jameslamb !

gforsyth (approved):
Nice work, @jameslamb !  This looks good to go.

jameslamb:
Thanks everyone!

PR #1541, opened 2025-08-26
Fix `dask_setup` and cleanup

### LINKED ISSUE #1540: Errors in `dask_setup`
The `protocol` is not being passed to `dask_setup` properly, presumably because that is already part of the base `dask worker` and we shouldn't be overriding it. Due to that `protocol=None` is always the case, and causes errors when attempting to determine the UCX implementation:

```python
2025-08-25 13:14:45,053 - distributed.worker - INFO - Starting Worker plugin CUDFSetup-8d8d95fa-8801-4a19-a940-f870f96b1207
Unable to start CUDA Context
  File "/opt/conda/dask-cuda/dask_cuda/initialize.py", line 104, in _create_cuda_context
    ucx_implementation = _get_active_ucx_implementation_name(protocol)
  File "/opt/conda/dask-cuda/dask_cuda/utils.py", line 1028, in _get_active_ucx_implementation_name
    raise ValueError("Protocol is neither UCXX nor UCX-Py")
ValueError: Protocol is neither UCXX nor UCX-Py

During handling of the above exception, another exception occurred:

  File "/opt/conda/dask-cuda/dask_cuda/initialize.py", line 42, in _warn_generic
    and not distributed.comm.ucx.cuda_context_created.has_context
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

pentschev:
Remove unused `dask_setup` arguments, leaving only `--create-cuda-context` using `worker._protocol` to determine the protocol correctly instead of the incorrect "overload" of the argument from `cli.py`. Additionally add tests to check that `dask_setup` is called.

Fixes #1540 .


TomAugspurger:
Yep. I think Peter is OOF today, so I'll merge this and keep an eye on CI.

jameslamb:
Thank you!

pentschev:
Thanks Tom. I'm mostly out today indeed, but I'm pretty sure the failure is not related to Dask, see the (or one of, haven't checked all the log) actual failure [here](https://github.com/rapidsai/cudf/actions/runs/17249655997/job/48950368319?pr=19768#step:10:4091).

jameslamb:
Oh thank you! Did not see that hiding in those logs too... ok maybe we had 2 problems there 😅

PR #1543, opened 2025-08-28
`get_device_memory_objects()`: use _columns instead of _value

madsbk:
Use `_columns` now that `_value` has been removed by https://github.com/rapidsai/cudf/pull/19799


pentschev (approved):
LGTM, thanks Mads!

PR #1547, opened 2025-09-02
WIP: use 'nvidia-ml-py' package for 'pynvml' module

jameslamb:
Contributes to https://github.com/rapidsai/build-infra/issues/293

The `pynvml` *package* (https://github.com/gpuopenanalytics/pynvml) provides 2 things:

* the `pynvml` **Python module**, transitively via a dependency on the `nvidia-ml-py` package
* the `pynvml_utils` Python module

This project's doesn't need the `pynvml_utils` Python module, so this PR proposes dropping the dependency on the `pynvml` package in favor of `nvidia-ml-py`.

### DIFF
--- a/conda/environments/all_cuda-129_arch-aarch64.yaml
+++ b/conda/environments/all_cuda-129_arch-aarch64.yaml
@@ -17,6 +17,7 @@ dependencies:
 - numba-cuda>=0.19.1,<0.20.0a0
 - numpy>=1.23,<3.0a0
 - numpydoc>=1.1.0
+- nvidia-ml-py>=12.560.30
 - pandas>=1.3
 - pre-commit
 - pynvml>=12.0.0,<13.0.0a0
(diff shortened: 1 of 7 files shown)


pentschev:
I think this looks good, but I think we won't be able to "see this in action" until all packages (including non-RAPIDS) remove the `pynvml` dependency, [currently it is still installed](https://github.com/rapidsai/dask-cuda/actions/runs/17392671280/job/49369335504?pr=1547#step:10:301) as I expected would be the case:

```
  + pynvml                                  12.0.0  pyhd8ed1ab_0                  conda-forge              26kB
```

jameslamb:
Thanks for checking that, you're right.

Let's start with `ucxx`:

https://github.com/rapidsai/ucxx/pull/498

I found that nothing is pulling in `pynvml` in test jobs there. Hopefully it's just that + `cudf` / `dask-cudf` pulling it in and we'll be able to eliminate it from this environment.

AyodeAwe (approved):

Issue #1517, opened 2025-07-03
Remove UCX-Py support


pentschev:
# Tasks to complete UCX-Py deprecation/archival

- [x] Ensure CUDA context attributes are moved from `distributed.comm.ucx` to some other place that is reachable also for `protocol="tcp"`: https://github.com/rapidsai/dask-cuda/pull/1548
- [x] Remove `distributed.comm.ucx`, potentially warn users of the removal and suggest them switching to `distributed-ucxx`: https://github.com/dask/distributed/pull/9105
  - [x] Remove schema: https://github.com/dask/distributed/pull/9127
  - [x] Wait for a new Distributed release
- [x] Make `distributed-ucxx` respond for both `protocol="ucx"` and `protocol="ucxx"`: https://github.com/rapidsai/ucxx/pull/504
- [x] Remove `distributed.comm.ucx.ucx-protocol` configuration injected by`get_ucx_config`, currently used for internal tracking of `initialize()` protocol: https://github.com/rapidsai/dask-cuda/pull/1546
- [x] Remove `test_rdd_protocol` along with special cases where it is tested in CI after removal of `distributed-ucxx`: https://github.com/rapidsai/dask-cuda/pull/1546
- [x] Remove all `"ucx-old"` tests: https://github.com/rapidsai/dask-cuda/pull/1546

## Details

Currently all CUDA context checking in https://github.com/rapidsai/dask-cuda/blob/branch-25.08/dask_cuda/initialize.py relies on `distributed.comm.ucx` which will soon be removed along with the UCX-Py deprecation, this will need to be updated for other protocols such as `protocol="tcp"` when that occurs.

`distributed.comm.ucx.ucx-protocol` configuration should be removed from `get_ucx_config`, currently used for internal tracking of `initialize()` protocol.

pentschev:
This is now complete.

PR #1548, opened 2025-09-03
Handle cuda context warnings

pentschev:
Update to support CUDA context warnings for all protocols, even after removal of `distributed.comm.ucx`. Depends on #1546. Partially addresses https://github.com/rapidsai/dask-cuda/issues/1517.


dask_cuda/initialize.py
```diff
@@ -25,98 +64,76 @@ def _create_cuda_context_handler():
...
 
-        distributed.comm.ucx.init_once()
-    except ModuleNotFoundError:
-        # UCX initialization has to be delegated to Distributed, it will take care
-        # of setting correct environment variables and importing `ucp` after that.
-        # Therefore if ``import ucp`` fails we can just continue here.
-        pass
 
+def _create_cuda_context():
+    global pre_existing_cuda_context, cuda_context_created
```
TomAugspurger, line 112: I could be misreading the diff, but is this `global` doing anyting? I dont' see `pre_existing_cuda_context` or `cuda_context_created` reassigned here.

Should the `global` move to `_create_cuda_context_and_warn` instead, where those are assigned?

dask_cuda/tests/test_initialize.py
```diff
@@ -225,3 +227,195 @@ def test_dask_cuda_import():
...
+
+def _test_cuda_context_warning_with_subprocess_warnings(protocol):
+    """Test CUDA context warnings from both parent and worker subprocesses.
+
+    This test creates a standalone script that imports a problematic library
+    and creates LocalCUDACluster with processes=True. This should generate
+    warnings from both the parent process and each worker subprocess, since
+    they all inherit the CUDA context created at import time.
+    """
+    import os
```
TomAugspurger, line 240: These could be moved to the top of the file?

dask_cuda/tests/test_initialize.py
```diff
@@ -225,3 +227,195 @@ def test_dask_cuda_import():
...
+    )
+
+    main_script_path = os.path.join(temp_dir, "test_subprocess_warnings.py")
+    with open(main_script_path, "w") as f:
+        f.write(main_script_code)
+
+    try:
+        # Run the main script in a subprocess
+        result = subprocess.run(
+            ["/raid/pentschev/miniforge3/envs/rn-250904/bin/python", main_script_path],
```
TomAugspurger, line 328:
```suggestion
            [sys.executable, main_script_path],
```

dask_cuda/tests/test_initialize.py
```diff
@@ -225,3 +227,195 @@ def test_dask_cuda_import():
...
+            capture_output=True,
+            text=True,
+            timeout=30,  # Reduced timeout for simpler test
+            cwd="/datasets/pentschev/src/dask-cuda",
```
TomAugspurger, line 332: Maybe `os.getcwd()`? Though that might be already be the default, so maybe you need to adjust that somehow?

```suggestion
            cwd=os.getcwd(),
```

dask_cuda/tests/test_initialize.py
```diff
@@ -225,3 +227,195 @@ def test_dask_cuda_import():
...
+                "This is often the result of a CUDA-enabled library calling a "
+                "CUDA runtime function before Dask-CUDA" in warning
+            ), f"Warning missing explanatory text: {warning}"
+    finally:
+        # Clean up temporary files
+        try:
+            if os.path.exists(temp_dir):
+                shutil.rmtree(temp_dir)
+        except Exception as e:
+            print(f"Cleanup error: {e}")
```
TomAugspurger, line 397: Possibly equivalent

```suggestion
        shutil.rmtree(temp_dir, onexc=print)
```

pentschev:
Had to revert the `shutil.rmtree(..., onexc=...)` change, it's [unfortunately only available starting in Python 3.12](https://docs.python.org/3/library/shutil.html#shutil.rmtree). Also, `onerror` (I believe `onexc` behaves the same) seems ridiculously verbose, for example `shutil.rmtree("/usr", onerror=print)` will print "Permission denied" for every file/subdirectory.

Issue #1566, opened 2025-10-07
Unintended distribution of packages


albertvillanova:
It looks like your package distribution currently includes modules that should not be part of the public release, such as: `examples.ucx`.

For example, in a Colab notebook you can directly import and run `examples.ucx`:
- Open a Colab notebook
- Execute:
```python
In [1]: from examples.ucx import client_initialize

In [2]: client_initialize.Client.__doc__
Out[2]: 'Connect to and submit computation to a Dask cluster\n\n    The Client connects users to a Dask cluster.  It provides an asynchronous...'

In [3]: !ls /usr/local/lib/python3.12/dist-packages/examples/ucx
Out[3]:
client_initialize.py  local_cuda_cluster.py
```
This indicates that this directory (`examples/ucx`) is being packaged and distributed, which is not intended behavior.

**Why this is problematic**
- Package bloat: Unnecessary files increase the size of the distribution.
- Namespace conflicts: For example, using a top-level `tests` package can shadow or conflict with users’ own `tests` modules.
- Exposure of internal code: Test modules are meant for development and CI, not for end users.
- Maintenance risks: Users might rely on internal test utilities that are not part of the public API and may change or be removed without notice.

**Suggested fix**

Update the packaging configuration to explicitly exclude development-only files and directories (like `examples/ucx/`) from the distribution.

PR #1567, opened 2025-10-07
Exclude 'examples' directory from distribution

TomAugspurger:
Closes https://github.com/rapidsai/dask-cuda/issues/1566.

After this change `unzip -l dist/dask_cuda-25.12.0-py3-none-any.whl | grep -i examples` shows nothing. Previously, the `examples/ucx` files were present.


gforsyth (approved):
:shipit:

PR #1598, opened 2025-12-16
Register sizeof for `pylibcudf.gpumemoryview`

TomAugspurger:
`test_dataframes_share_dev_mem` is currently failing on `main` (https://github.com/rapidsai/dask-cuda/actions/runs/20257350468/job/58162113282#step:11:5622) after https://github.com/rapidsai/cudf/pull/20869.

Something changed in `dask_cuda.get_device_memory_objects.dispatch`, I think related to `Series._column` returning a `gpumemoryview`. We lacked a `dask.sizeof` implementation for `gpumemoryview`, so fell back to the default, which gave inaccurate memory usage results, which broke these tests that are finely tuned to the amount of memory usage that dask / dask-cuda see.


vyasr:
Thanks for the fix Tom!

PR #1615, opened 2026-01-31
Update cuda-core pin

pentschev:
Due to broken pins in numba-cuda conda-forge packages for versions 0.25 and 0.26 we need to update Dask-CUDA to ensure it can pick up a newer cuda-core version.


bdice (approved):

aamijar:
Hi @pentschev, I wonder if this PR is causing some issues in raft?
https://github.com/rapidsai/raft/actions/runs/21557712256/job/62117364035?pr=2936
https://github.com/rapidsai/raft/actions/runs/21528320914/job/62108982260?pr=2933

PR #1644, opened 2026-04-12
Replace rmm.mr.get_current_device_resource_type with type(rmm.mr.get_current_device_resource())

bdice:
## Summary

- Replace all uses of `rmm.mr.get_current_device_resource_type` with `type(rmm.mr.get_current_device_resource())` in tests, enabling removal of the `get_current_device_resource_type` function from RMM.

## Details

`rmm.mr.get_current_device_resource_type` is being removed from RMM. The only usage in dask-cuda is in tests. This PR replaces all 12 call sites across `test_dask_cuda_worker.py` and `test_local_cuda_cluster.py` with the equivalent `type(rmm.mr.get_current_device_resource())`.


pentschev:
LGTM, thanks @bdice !

pentschev (approved):

PR #1645, opened 2026-04-15
Migrate from pynvml to cuda.core.system

mdboom:
This migrates from pynvml to the new Cython/cybind-based bindings for NVML in `cuda.core.system`.


TomAugspurger:
/ok to test https://github.com/rapidsai/dask-cuda/pull/1645/commits/00453df8d649a3cdadc553edfad33b908a79644e

dask_cuda/utils.py
```diff
@@ -100,30 +55,26 @@ def get_gpu_handle(device_id=0):
...
+            if device.mig.is_mig_device:
+                device = device.mig.parent
         else:
-            handle = pynvml.nvmlDeviceGetHandleByIndex(device_id)
-        return handle
-    except pynvml.NVMLError:
+            device = system.Device(index=int(device_id))
+        return device
+    except system.NvmlError:
         raise ValueError(f"Invalid device index or UUID: {device_id}")
```
TomAugspurger: A bit out of scope, but while we're here:

```suggestion
    except system.NvmlError as e:
        raise ValueError(f"Invalid device index or UUID: {device_id}") from e
```

TomAugspurger (approved):
Looks nice overall.

IMO, the changes to `dask_cuda.utils` (removal of `unpack_bitmask`, change in return types) aren't worth trying to deprecate.

dask_cuda/utils.py
pentschev: +1 to this.

pentschev (changes requested):
> IMO, the changes to dask_cuda.utils (removal of unpack_bitmask, change in return types) aren't worth trying to deprecate.

Yes, absolutely not worth it, just removing them (as is being proposed) is enough.

@mdboom it seems you're still missing some `cuda-core` pinning changes:

- https://github.com/rapidsai/dask-cuda/blob/c0f9a3f1a87a7a90666141b07e4627709a8ae6de/dependencies.yaml#L257-L266
- https://github.com/rapidsai/dask-cuda/blob/c0f9a3f1a87a7a90666141b07e4627709a8ae6de/conda/recipes/dask-cuda/recipe.yaml#L40

-And since we're upgrading the version of `cuda-core` it would be nice to properly handle (i.e., remove) previous compatibility changes:
- https://github.com/rapidsai/dask-cuda/blob/c0f9a3f1a87a7a90666141b07e4627709a8ae6de/dask_cuda/initialize.py#L12-L15
- https://github.com/rapidsai/dask-cuda/blob/c0f9a3f1a87a7a90666141b07e4627709a8ae6de/dask_cuda/_compat.py#L22-L29

As well as in the following test files:
- https://github.com/rapidsai/dask-cuda/blob/main/dask_cuda/tests/test_initialize.py
- https://github.com/rapidsai/dask-cuda/blob/main/dask_cuda/tests/test_utils.py

Could you do that?

pentschev:
/ok to test https://github.com/rapidsai/dask-cuda/commit/97bfb9406a378d43fcb591d8c45417d6676fecca

PR #1652, opened 2026-05-26
Improve UCX documentation

pentschev:
The docs now present `protocol="ucx"` with automatic transport selection as the normal path, while manual UCX config examples are explicitly benchmark-oriented. Benchmark CLIs and result metadata now distinguish automatic vs manual UCX transport selection, with tests covering the default automatic behavior.


TomAugspurger (approved):

pentschev:
Thanks Tom!

PR #1661, opened 2026-06-24
Bump cuda-core minimum to 0.5.1

bdice:
Bumps the minimum supported cuda-core version from 0.3.2 to 0.5.1 in dependency metadata and generated conda environment files.

This aligns with numba-cuda-mlir's lower bound, which helps us align the compatible bounds across RAPIDS.


pentschev (approved):

jameslamb (approved):
Works for me. If this causes any issues we have plenty of time left in 26.08 development to find out (esp. in https://github.com/rapidsai/build-planning/issues/285).

PR #1664, opened 2026-07-06
remove workaround for cuda-core<0.5.1, test against oldest cuda-core in UCXX tests

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/285

This project is already not depending on `cuda-python` and carrying the same `cuda-core` floor I'm proposing across RAPIDS (`>=0.5.1`), and it's already testing against oldest `cuda-core` in most places:

https://github.com/rapidsai/dask-cuda/blob/467fad08c868b0e99dea2441dc2f17f10923c526/dependencies.yaml#L294-L296

This proposes a few other small changes related to that dependency:

* removes a workaround for `cuda-core<0.5` (unnecessary since the project requires `cuda-core>=0.5.1`)
* threads `RAPIDS_DEPENDENCIES` through the UCXX tests too

### DIFF
--- a/ci/test_python_ucxx.sh
+++ b/ci/test_python_ucxx.sh
@@ -16,7 +16,7 @@ rapids-logger "Generate Python testing dependencies"
 rapids-dependency-file-generator \
   --output conda \
   --file-key test_python \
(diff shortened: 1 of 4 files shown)


bdice (approved):
Confirmed, it seems like no changes are needed in the dependency pinnings. LGTM.

TomAugspurger (approved):

PR #1682, opened 2026-09-02
Turn on autosectionlabel in sphinx docs

wence-:
This will allow linking to arbitrary sections through intersphinx inventories. Without these one can only link to a given page or some kind of object name.


bdice (approved):

josephine-wolf-oberholtzer (approved):
LGTM

